PR #135, opened 2015-05-19
adds omniauth-cas

### LINKED ISSUE #109: Authentication via Omniauth CAS
Think about CAS omniauth as the best option for PUL. @eliotjordan had some good suggestions on how to do this https://github.com/pulibrary/pul-store/issues/326. This will be available to any PU user with an official net id.

eliotjordan:
closes #109


kevinreiss:
@eliotjordan I think this looks good as a first pass.

PR #139, opened 2015-05-21
BL 5.13.1, index and browse rake tasks

### LINKED ISSUE #128: index:seed task for specs
Store MARCXML in a fixture that gets indexed into the test Solr instance. This will depend on Orangeindex (our instance of traject), and so we'll need to start tagging versions Orangeindex to manage updates/versions.

This data should only be used to help w/ integration tests (feature specs). Relevancy tests should be in the Orangeindex repo.

### LINKED ISSUE #129: Stream the browse list record processing
To avoid the need to load millions of records into memory.
Call numbers can be in CSV format
Subjects and Authors need to be json, but can be streamed with this gem: https://github.com/brianmario/yajl-ruby

### LINKED ISSUE #136: remove migration dependency from browse list seed
Omniauth adds more migrations. See #109

tampakis:
added index and browse list rake tasks

put getdir in string_functions library

Closes #128, closes #129, closes #136.


kevinreiss:
Looks good. :ok_hand:

PR #149, opened 2015-06-16
pulls librarian view from bibdata service

tampakis:
Pulls marc record for librarian view from bibdata service. This will allow us remove the raw marcxml from our Solr index.


jpstroop:
You're going to hate me ( :stuck_out_tongue_closed_eyes: ), but I think you should probably add a test to make sure this works; seems like something that could get broken in a upgrade pretty easily.

tampakis:
done

PR #156, opened 2015-07-09
Add fonts

tampakis:
Match library footer

Adjust spacing and font sizes for facets

Add main catalog to header

Make search text for search button visible to screen readers only

Move search constraints and did you mean to top of results

Consistent cross-browser select styles for search box


jpstroop:
Is it just me or is this a lot to be sending over the wire? Is this typical these days, @kevinreiss ?

kevinreiss:
@jpstroop We were having a lot of cross browser issues with web fonts so we
switched to this approach where all the font files are included.

jpstroop:
OK. Seems like sending fonts is expensive no matter what:

http://www.webpagetest.org/result/150709_GX_18KS/1/details/

kevinreiss:
Yes, it isn't good, especially on 3G. I think we can possibly do more with compression. On library.princeton.edu they are served from varnish when on port 80 and get an indefinite cache header which does help some.

kevinreiss:
Or we could get serious and use something like https://typekit.com/.

PR #172, opened 2015-07-28
Add basic integration with umlaut.

kevinreiss:
@tampakis don't merge yet. Let's discuss how or if we want to do integration tests for the umlaut responses. This will embed full-text links where appropriate and add a holdings display for every record. We should think of the placeholder we want while results load in the holdings area. Perhaps some representation of the known holding locations for the record.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -47,7 +47,7 @@ GEM
       activerecord (>= 3.0)
     addressable (2.3.8)
     arel (6.0.2)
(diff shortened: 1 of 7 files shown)


jpstroop:
:+1:

kevinreiss:
@tampakis I think this is good to go for the demo now.

tampakis:
:+1: We'll come up with appropriate feature tests after seeing Umlaut a bit in action.

PR #176, opened 2015-08-04
Tooltips and other UI enhancements

axamei:
Add new fixtures and facets view

Make default cover larger

Add tooltip for subject search and browse

Add tooltips to related names and change request_spec to reflect new tooltip attributes

Update request spec

Add tooltip to holding info call number browse


tampakis:
:+1:

PR #199, opened 2015-08-14
spellchecks suggests search phrase instead of individual words

tampakis:
- Allows records to be retrieved when bibdata (which pulls in the live Marc data) goes down.
 - Fixes bug in Blacklight spelling suggestions feature, which returns a funky hash when using Solr 5. Submitted PR to Blacklight (https://github.com/projectblacklight/blacklight/pull/1235) so we won't have to override the Gem.
 - Using the collate feature which suggests a new search phrase rather than spelling suggestions for individual words.


kevinreiss:
:+1:

Issue #160, opened 2015-07-14
Improve facet/filter query performance


tampakis:
It takes about 7 seconds for the main catalog page to render. 6 of those 7 seconds are spent populating the facets. They should be caching on the filter cache, but enabling that didn't really improve performance.
[hints for better utilizing cache](http://solr.pl/en/2011/02/07/optimization-filter-cache/)
[new filter query feature in Solr 5.2](http://yonik.com/solr/query-syntax/#FilterQuery)

tampakis:
Information about performance with different facet caches on slide 34: http://www.slideshare.net/lucenerevolution/seeley-solr-facetseurocon2011

PR #203, opened 2015-08-17
only request access and format facet in solr on homepage

tampakis:
Only the format and access_facets will render and make a request to solr on the homepage to improve Solr performance. Advances #160.


kevinreiss:
:+1: Good move on this one.

tampakis:
Coverage is up now.

PR #214, opened 2015-08-20
Vertically align remove button in facets to be in middle. Closes #210

### LINKED ISSUE #206: Use "Classic Catalog" instead of "Main Catalog" on record view page
screenshot from 2015-08-19 16 13 03

### LINKED ISSUE #208: Get "orangelight" out of title bar
title

Replace w/ what, I'm not sure, but I doubt LISSC etc. will go for OrangeLight.

### LINKED ISSUE #210: Line up facet value "x", and count, add label for "x"
facet

The "x" is high, and I also wonder if the count is higher. I can't tell.

Also, not to harp on tooltips, but I wonder if there are accessibility issues around the fact that what this does isn't labeled?

The HTML markup is this:

```html
<a class="remove" 
    href="/?q=john+cage+variations&amp;search_field=all_fields">
    <span class="glyphicon glyphicon-remove"></span>
    <span class="sr-only">[remove]</span>
</a>
```

I wonder if add `@data-original-title` and/or `@title` with the value 'Remove Format facet value "Book"' (or something like that) to the anchor tag that would be a good idea?

### LINKED ISSUE #213: Pivot Facets don't have 'x' button
The 'x' button has disappeared from the pivot facet panel.The only way to remove a pivot facet limit is in the render constraints bar at the top.
pivot_facet

axapu:
Get orangelight out of title bar and fix Guided search title. Closes #208

%~~Change main catalog to classic catalog. Closes #206~~

Closes #213.

### DIFF
--- a/config/locales/blacklight.en.yml
+++ b/config/locales/blacklight.en.yml
@@ -1,6 +1,6 @@
 en:
   blacklight:
-    application_name: 'Orangelight'
+    application_name: 'Catalog | Princeton University Library'
     welcome: 'Welcome!'
     voyager: 'Main Catalog'
 
--- a/app/assets/javascripts/orangelight.js
+++ b/app/assets/javascripts/orangelight.js
@@ -1,6 +1,7 @@
 //= require 'jquery'
 
 $( document ).ready(function() {
+    //link highlighting of subject heirarchy
     $(".search-subject").hover(
         function() {
             $(this).prevAll().addClass("subject-heirarchy");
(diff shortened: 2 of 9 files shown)


jpstroop:
I forget which it is now, but we should keep the label for WebVoyage that Marvin wanted, whatever it was. Is this PR doing that?

axapu:
Is WebVoyage the name to replace Orangelight? I changed it to "Catalog | Princeton University Library"

jpstroop:
No, sorry, WebVoyage is the current catalog.princeton.edu. Nikitas talked with Marvin directly and I think we need to stick with "Main Catalog" (as opposed to "Classic ...") for now. Is that correct, @tampakis?

jpstroop:
ps: I'm cool with "Catalog | Princeton University Library"

config/locales/blacklight.en.yml
```diff
@@ -1,8 +1,8 @@
 en:
   blacklight:
-    application_name: 'Orangelight'
+    application_name: 'Catalog | Princeton University Library'
     welcome: 'Welcome!'
-    voyager: 'Main Catalog'
+    voyager: 'Classic Catalog'
```
tampakis: This should stay Main Catalog

axapu:
Okay. I'll revert that change. Close #206?

jpstroop:
+1, just did

tampakis:
:+1:

PR #237, opened 2015-09-02
Remove view profile link, disable account edit, link to CAS directly …

### LINKED ISSUE #171: CAS Authenticated Users and Passwords
We should set up the user profile page to not allow users signed in with SSO to set or change the password fields. We should just reserve that for guest users. @tampakis @axapu Any objections to this proposed change?

axapu:
…via my account closes #171 

Fix white space issue with Firefox trailing letters


tampakis:
:+1:

PR #242, opened 2015-09-10
provide links to other versions of bib record

### LINKED ISSUE #217: Add links to other versions based on Voyager config
See here: https://github.com/pulibrary/orangeindex/blob/development/voyagersamplecfg/newdisplay12.cfg#L139-L151

and here:

https://github.com/pulibrary/orangeindex/blob/development/voyagersamplecfg/newdisplay12.cfg#L288-L289

The ones that contain 'BBID' should be pretty straight forward. E.g., given `3500 BBID787W` use the bib id in `787|w` to build a link.

The others, like `020Z020A` are trickier, because there's no guarantee that we have a that title, and hitting a service to check at index time would _really_ slow things down. We should probably try to find a way to rely on Umlaut for those. I assume that in `020Z020A`, z before a is an order of preference?

We'll also need to find a way to get the ISBN from, e.g. "9780691134970 (hardcover : alk. paper)" if we haven't already. I'm sure it won't be as easy as tokenizing on the first whitespace in the string, but that might be a place to start.

tampakis:
Closes #217 
@marvinmarvinb - Other versions of the bib record are now linked across OCLC, ISBN, and ISSN numbers. See attached image.
other_versions


jpstroop:
:+1: Will merge when Travis is :green_heart:

PR #243, opened 2015-09-10
redirect to bib record from lccn, isbn, issn, oclc num

### LINKED ISSUE #224: Alternative routes for standard numbers
e.g. 

`/catalog/oclc/12345` should `301` to `/catalog/whatever` similar for ISBN, LCCN.

tampakis:
Closes #224. Visiting `/catalog/oclc/12345` will redirect to a bib matching the oclc number if a match is found. Otherwise it will redirect to a keyword search. The standard numbers are normalized for the search.

Note: A full reindex is required for this feature and the "other versions" feature to work.


jpstroop:
Maybe a full reindex over the weekend? Or should you do it sooner so that we know everything is working for LISSC?

PR #255, opened 2015-09-15
Umlaut Fixes

kevinreiss:
This is a first pass.

### DIFF
--- a/app/views/catalog/_show_main_content.html.erb
+++ b/app/views/catalog/_show_main_content.html.erb
@@ -19,14 +19,14 @@
   <script type="text/javascript">
     jQuery(document).ready(function(){
       var ctx = $("span.Z3988").attr("title");
-      var umlaut_base = "https://getit-dev.princeton.edu";
+      var umlaut_base = "https://getit.princeton.edu";
       var updater = new Umlaut.HtmlUpdater(umlaut_base, ctx );
       updater.add_section_target({
         umlaut_section_id: "fulltext",
         selector:"#full_text",
         before_update: function(html, count) {
           //hide the section heading on the snippet, we don't want it
-          //$(html).find(".section_heading h3").hide();
+          $(html).find(".section_heading h3").hide();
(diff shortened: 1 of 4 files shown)


app/views/catalog/_show_main_content.html.erb
```diff
@@ -19,14 +19,14 @@
   <script type="text/javascript">
     jQuery(document).ready(function(){
       var ctx = $("span.Z3988").attr("title");
-      var umlaut_base = "https://getit-dev.princeton.edu";
+      var umlaut_base = "https://getit.princeton.edu";
```
kevinreiss: This should be configured somewhere if we are going to continue to use it.

config/initializers/blacklight_initializer.rb
```diff
@@ -24,39 +24,57 @@ def export_as_openurl_ctx_kev(format = nil)
     id = to_marc.find{|field| field.tag == '001'}
```
kevinreiss: The idea is that we will move the entire export_as_openurl_ctx_kev function logic (and any attendant helper code) to orangeindex and just inject an OpenURL into the solr document. Correct @tampakis.

tampakis: Correct.

kevinreiss: Should I make an issue in orangeindex?

tampakis: I don't know if this is ideal, but we've been opening the orangeindex issues in orangelight.

kevinreiss: Okay. I'll open one.

app/views/catalog/_show_main_content.html.erb
tampakis: Maybe in the [blacklight.yml](https://github.com/pulibrary/orangelight/blob/development/config/blacklight.yml)?

PR #258, opened 2015-09-17
authorized ips/hosts

tampakis:
Only allow hostnames and ips on the whitelist to access the application.


jpstroop:
http://stackoverflow.com/a/7809304/714478 ?

tampakis:
Since the whitelist isn't in version control, I needed to have travis copy over the template files to read for the tests

jpstroop:
OK. I was thinking you could grab the list from the config, but then also get the local IP and append it to that list.

PR #262, opened 2015-09-18
disable hostname whitelist, blank ip list

tampakis:
The hostname whitelist feature has been unreliable, so it is disabled in this PR.
Changed logic in application controller to allow all users through if the ip whitelist is empty.


kevinreiss:
:+1:

PR #288, opened 2015-10-14
Live availability

### LINKED ISSUE #107: Real Time Availability on Index View
For non-serials think about doing this directly against Voyager item record. Serials display only holdings level information w/stackmap or offsite link. Records with digital surrogates should provides these links on the index view. What non-Voyager records do here?

terrellt:
Closes #107

### DIFF
--- a/app/assets/javascripts/availability.js.coffee
+++ b/app/assets/javascripts/availability.js.coffee
@@ -13,17 +13,23 @@ class AvailabilityUpdater
     for record_id, availability_info of records
       this.apply_record(record_id, availability_info)
   apply_record: (record_id, availability_info) ->
-    return if availability_info.more_holdings == true
-    for location_code, status of availability_info
-      element = $("*[data-record-id='#{record_id}'][data-loc-code='#{location_code}']")
-      continue unless element.length > 0
-      availability_element = element.children(".availability-icon")
-      availability_element.addClass("glyphicon")
-      if status == "Not Charged"
-        availability_element.addClass("glyphicon-ok text-success")
-      else
-        availability_element.addClass("glyphicon-remove text-danger")
-      availability_element.prop('title', status)
+    if availability_info.more_holdings == true
+      this.record_needs_more_info(record_id)
+    else
+      for location_code, status of availability_info
+        this.apply_record_icon(record_id, location_code, status)
     true
+  record_needs_more_info: (record_id) ->
+    element = $("*[data-record-id='#{record_id}'] .availability-icon")
+    element.addClass("glyphicon glyphicon-info-sign")
+    element.prop('title', "Click on the record for full availability info")
+  apply_record_icon: (record_id, location_code, status) ->
+     availability_element = $("*[data-record-id='#{record_id}'][data-loc-code='#{location_code}'] .availability-icon")
+     availability_element.addClass("glyphicon")
+     if status == "Not Charged"
+       availability_element.addClass("glyphicon-ok text-success")
+     else
+       availability_element.addClass("glyphicon-remove text-danger")
+     availability_element.prop('title', status)
   record_ids: ->
     $("*[data-availability-record][data-record-id]").map((_, x) -> $(x).attr("data-record-id"))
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/availability.js.coffee
```diff
@@ -0,0 +1,29 @@
...
+    for record_id, availability_info of records
+      this.apply_record(record_id, availability_info)
+  apply_record: (record_id, availability_info) ->
+    return if availability_info.more_holdings == true
+    for location_code, status of availability_info
+      element = $("*[data-record-id='#{record_id}'][data-loc-code='#{location_code}']")
+      continue unless element.length > 0
+      availability_element = element.children(".availability-icon")
+      availability_element.addClass("glyphicon")
+      if status == "Not Charged"
```
tampakis: It turns out that available/unavailable is more involved than "Not charged"/else. I'd be ok with merging as is if we open up a new ticket about how to classify based on some of the other item statuses mentioned in the [issue](https://github.com/pulibrary/orangelight/issues/107#issuecomment-147882725). I'm thinking that logic may ultimately end up in bibdata.

terrellt: :+1: to a new ticket and a new PR for that logic, when it becomes clear.

app/assets/javascripts/availability.js.coffee
```diff
@@ -0,0 +1,29 @@
...
+  request_availability: ->
+    ids = this.record_ids().toArray()
+    params = $.param({ids: ids})
+    url = "#{@availability_url}?#{params}"
+    $.getJSON(url, this.process_records)
+  process_records: (records) =>
+    for record_id, availability_info of records
+      this.apply_record(record_id, availability_info)
+  apply_record: (record_id, availability_info) ->
+    return if availability_info.more_holdings == true
```
tampakis: Since the "more_holdings" field is supposed to mean that the user should click on the record to get full availability info, I'm thinking we could use the info icon (glyphicon glyphicon-info-sign) with a tooltip "Click on the record for full availability info" for this case.

jpstroop:
> I'm thinking that logic may ultimately end up in bibdata.

I'm not sure I agree. Other uses of the bibdata services, like @kevinreiss's fulfullment app, might need these more granular values. I know it's a pain hard-coding these strings, but I think it's probably a better approach for now.

app/assets/javascripts/availability.js.coffee
jpstroop: > I'm thinking that logic may ultimately end up in bibdata.

I'm not sure I agree. Other uses of the bibdata services, like @kevinreiss's fulfullment app, might need these more granular values. I know it's a pain hard-coding these strings, but I think it's probably a better approach for now.

tampakis: I think we should still pass in these more granular values, but we might also want to pass in an additional value to correspond to "available/unavailable/click for more info" along with the status message.

tampakis: (via the availability service in bibdata)

app/assets/javascripts/availability.js.coffee
terrellt: So there's two kinds of more holdings: There's the view determined it has more holdings, so it doesn't display anything, and then there's the view displays holdings, but the API shows "more_holdings: true". Do both cases need that message?

tampakis: Yes.

tampakis: Eventually we might want to provide more of an explanation for why the user needs to click on the record (ie. more copies/items in a given holding location vs more holding locations), but I think "Click on record for full availability" is sufficient for this PR.

terrellt:
@tampakis Give that a shot

tampakis:
Looks good! Want to rebase?

terrellt:
@tampakis Removed the extra comment commit, the rest should be work based.

tampakis:
:+1:

PR #290, opened 2015-10-15
Automatically fetch book covers from google books.

### LINKED ISSUE #131: Covers
Somewhat related to #106 (for search results).

Options: 
 * Umlaut (would require an OpenURL embedded on the page cf. #85)
 * Hit Amazon, Google, Hathi directly

terrellt:
Closes #131

### DIFF
--- a/app/assets/javascripts/book_covers.js.coffee
+++ b/app/assets/javascripts/book_covers.js.coffee
@@ -1,0 +1,48 @@
+jQuery ->
+  window.book_cover_manager = new BookCoverManager
+class BookCoverManager
+  google_url: "https://www.googleapis.com/books/v1/volumes?q="
+  identifiers: {
+    isbn: "isbn_display",
+    oclc: "oclc_s",
+    lccn: "lccn_display"
+  }
+  constructor: ->
+    this.find_book_covers()
+  find_book_covers: ->
+    number = this.get_number()
+  requests: []
+  get_number: ->
+    identifier = null
+    for identifier_type, field of @identifiers
+      identifiers = this.find_identifier(identifier_type, field)
+      if identifiers?
+        for identifier in identifiers
+          this.fetch_identifier(identifier)
+    true
+  fetch_identifier: (identifier) ->
+    url = "#{@google_url}#{identifier}"
+    @requests.push($.getJSON(url, this.process_record))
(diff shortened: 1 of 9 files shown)


app/assets/javascripts/book_covers.js.coffee
```diff
@@ -0,0 +1,42 @@
...
+class BookCoverManager
+  constructor: ->
+    this.find_book_covers()
+  find_book_covers: ->
+    number = this.isbn() || this.oclc()
+    return unless number?
+    url = "#{@google_url}#{number}"
+    $.getJSON(url, this.process_record)
+  isbn: ->
+    isbn = $("dd.blacklight-isbn_display li")
```
tampakis: There is a normalized version of the ISBN in the [isbn_s](https://github.com/pulibrary/orangeindex/blob/development/lib/traject_config.rb#L727) field.

tampakis:
The Google Book api also accepts LCCN as an identifier. Could pass that over in addition to ISBN/OCLC? The field is [lccn_s](https://github.com/pulibrary/orangeindex/blob/development/lib/traject_config.rb#L715).

app/assets/javascripts/book_covers.js.coffee
terrellt: Yeah, but I don't think this is in the view, right?

tampakis: Good point. Would it make sense to include it without displaying it?

terrellt: I could do that - is there a concern that the ISBN in the display isn't queryable?

terrellt:
> The Google Book api also accepts LCCN as an identifier. Could pass that over in addition to ISBN/OCLC? The field is lccn_s.

Done.

jpstroop:
FWIW, lccn is likely to get us the most of the three, so if there's a preferred order, I'd go w/ that first.

terrellt:
@jpstroop It'll try them all and stop when it finds a cover. In the test data it was only ever isbn that returned one.

jpstroop:
> if there's a preferred order

_and_ it makes a difference.

jpstroop:
:+1:

tampakis:
:+1:

PR #294, opened 2015-10-15
remove assets cache as a linked directory

tampakis:
Compile assets on each deploy


jpstroop:
@terrellt, could you review?

app/assets/javascripts/application.js
```diff
@@ -34,4 +34,6 @@
 //= require modernizr
 
 //= require orangelight
+//= require availability
```
terrellt: Is the explicit requires just a matter of preference? The require tree statement is enough for the asset pipeline.

tampakis: It didn't work without these explicit requires. Maybe there's a setting that in the config that we don't have enabled?

terrellt: That's strange...if it works in dev it should compile in prod.

PR #298, opened 2015-10-16
Alternative proposal for availability indication

jpstroop:
Perhaps another way to skin #107?

It's not fully baked--some records still show the more info glyphicon for some reason, and I think we'd want to work on getting the labels to be a consistent width.

Just an idea...

### DIFF
--- a/app/assets/javascripts/availability.js.coffee
+++ b/app/assets/javascripts/availability.js.coffee
@@ -21,15 +21,21 @@ class AvailabilityUpdater
     true
   record_needs_more_info: (record_id) ->
     element = $("*[data-record-id='#{record_id}'] .availability-icon")
-    element.addClass("glyphicon glyphicon-info-sign")
+    element.addClass("label label-default")
+    element.text("View Record for Availability")
     element.prop('title', "Click on the record for full availability info")
   apply_record_icon: (record_id, location_code, status) ->
-     availability_element = $("*[data-record-id='#{record_id}'][data-loc-code='#{location_code}'] .availability-icon")
-     availability_element.addClass("glyphicon")
-     if status == "Not Charged"
-       availability_element.addClass("glyphicon-ok text-success")
-     else
-       availability_element.addClass("glyphicon-remove text-danger")
-     availability_element.prop('title', status)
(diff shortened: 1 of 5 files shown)


jpstroop:
screenshot from 2015-10-16 10 57 34

jpstroop:
ping @terrellt @kevinreiss @tampakis

kevinreiss:
I like the presentation of this with the colored background and text instead of an icon, but I still think staff are probably going to want to see the status, but we can see if it comes up.

jpstroop:
The text for the label is the status. I just added a hook to change "Not Charged" to "Available" because English.

jpstroop:
https://github.com/pulibrary/orangelight/blob/availability_labels/app/assets/javascripts/availability.js.coffee#L30-L36

screenshot from 2015-10-16 11 16 46

terrellt:
:+1: Works for me. The reason the info label is showing up sometimes is because sometimes ruby handles the display. https://github.com/pulibrary/orangelight/blob/development/app/helpers/application_helper.rb#L56

terrellt:
We were talking about this in slack, but to document: I think it'd be good to have bibdata return what a status means (available or not), so that we don't inevitably duplicate that logic.

jpstroop:
I'll change that on this PR

kevinreiss:
+1 to that @terrellt.

jpstroop:
I just squashed these ... I (humbly) think this is avenue we should pursue.

tampakis:
I agree that this is heading in the right direction. The part that will be a challenge is (as @jpstroop said), getting the labels to be a consistent width and determining the correct amount of left padding to fit the different status codes to be pulled in next to the holding info (so the alignment is consistent).

jpstroop:
> getting the labels to be a consistent width

I'm sure we can solve that and think we should forge ahead

terrellt:
:+1: We can merge this and open a ticket to refactor with an updated bibdata API (pointing to bibdata ticket). Agreed?

terrellt:
The only thing that may become trouble is the increased width in regard to mobile displays.

PR #306, opened 2015-10-16
Use bulk google API and add covers to search.

### LINKED ISSUE #301: Covers in search results
Let's give it a try.

tpendragon:
Closes #301

### DIFF
--- a/app/assets/javascripts/book_covers.js.coffee
+++ b/app/assets/javascripts/book_covers.js.coffee
@@ -1,7 +1,7 @@
 jQuery ->
   window.book_cover_manager = new BookCoverManager
 class BookCoverManager
-  google_url: "https://www.googleapis.com/books/v1/volumes?q="
+  google_url: "https://books.google.com/books?callback=?&jscmd=viewapi&bibkeys="
   identifiers: {
     isbn: "isbn",
     oclc: "http://purl.org/library/oclcnum",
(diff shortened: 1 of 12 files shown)


tpendragon:
If this gets merged and deployed, the solr config on the staging server will have to get updated.

jpstroop:
:+1: The only thing I notice is that w/ e.g. 4131923 the image is getting upsampled:

http://localhost:3000/catalog/4131923 

https://books.google.com/books/content?id=7zZnAAAAMAAJ&printsec=frontcover&img=1&zoom=5

it doesn't seem like that `zoom` param actually does anything, so maybe we should just let the browser figure out the size?

tpendragon:
@jpstroop I don't process any of the parameters right now, I just accept the thumbnail parameter from google books. When I ask for the URL without the zoom level for some images I don't get anything back.

tpendragon:
Oh, I see, you're saying don't restrict the size of the thumbnail element so that it can be smaller/larger if necessary?

jpstroop:
> Oh, I see, you're saying don't restrict the size of the thumbnail element so that it can be smaller/larger if necessary?

Yes, that.

jpstroop:
Though: https://books.google.com/books/content?id=7zZnAAAAMAAJ&printsec=frontcover&img=1&zoom=1

tpendragon:
Changed the CSS. I can manipulate the zoom level to 1, but I don't think google is providing any promises with that since it's not returned in the API.

tpendragon:
Changing zoom to 1 seems to work.

jpstroop:
Every image in the test set responds to zoom=1. I think you should try it. Some of them also include a param `edge=curl` which I think we should strip.

tpendragon:
> Every image in the test set responds to zoom=1. I think you should try it. Some of them also include a param edge=curl which I think we should strip.

Done.

jpstroop:
OK. I can merge assuming there's :green_heart: from Travis, but do you want to muck w/ Solr?

PR #316, opened 2015-11-02
Add print css and fix bookmark button in record view. Closes #315 and…

### LINKED ISSUE #296: Printing records
From Marvin:

> I see there is no obvious way for a user to Print a record. Could use a Print button
>
> Lots of other options, but not print

### LINKED ISSUE #315: Bookmark button looks strange
We should probably make it a similar size or relocate
screenshot from 2015-11-02 08 33 48

axapu:
… closes #296

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -353,3 +353,6 @@ DEPENDENCIES
   turbolinks
   uglifier (>= 1.3.0)
   yajl-ruby
+
+BUNDLED WITH
+   1.10.4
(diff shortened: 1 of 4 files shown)


tampakis:
Thanks @axapu! The print css is a major improvement over previous behavior. There are still a couple of issues I see with the format on the record show so I'm going to keep that issue open and leave a few comments there.

axapu:
Sounds good @tampakis

PR #321, opened 2015-11-09
Fix print styling for item record view

### LINKED ISSUE #296: Printing records

axapu:
closes #296


kevinreiss:
:+1:

Issue #218, opened 2015-08-24
Use Solr document fields in the open url


kevinreiss:
http://pulsearch-dev.princeton.edu/catalog/9026021 has two ISBNs attached to it but the OpenURL created for the "COINS" template has a blank rft.isbn parameter in it. Need to audit the code that creates this to make sure it reliably includes ISBN's when they are present.

Current Output:

```
<span class="Z3988" title="ctx_ver=Z39.88-2004&amp;rft_val_fmt=info%3Aofi%2Ffmt%3Akev%3Amtx%3Abook&amp;rfr_id=info%3Asid%2Fblacklight.rubyforge.org%3Agenerator&amp;rft.genre=book&amp;rft.btitle=The+next+justice+repairing+the+Supreme+Court+appointments+process+%2F&amp;rft.title=The+next+justice+repairing+the+Supreme+Court+appointments+process+%2F&amp;rft.au=Eisgruber%2C+Christopher+L.&amp;rft.date=c2007.&amp;rft.place=Princeton+%3A&amp;rft.pub=Princeton+University+Press%2C&amp;rft.edition=&amp;rft.isbn=&amp;rft_id=pul_9026021"></span>
```

kevinreiss:
The alternative format displayed in the document's head 

```
<link rel="alternate" title="openurl_ctx_kev" type="application/x-openurl-ctx-kev" href="http://pulsearch-dev.princeton.edu/catalog/9026021.openurl_ctx_kev">
```

also produces a slightly different OpenURL. We should probably get them in sync.

tampakis:
There are two issues here, which we should discuss at the meeting today. 
 - For this record both of the isbns are in $z, which means that the ISBN is [obsolete](https://www.loc.gov/marc/bibliographic/bd020.html). The parameter is blank because it grabs the 020 $a from the Marc.
 - I think we should pass the parameter the normalized isbn value from the isbn field rather than extract it from the marc.

kevinreiss:
@tampakis @jpstroop we received a feedback comment asking for zotero support on the search results page that is related to this open issue. Zotero will scrape all ctx span tags (shown in the initial comment on this thread) embedded on a page as individual citations. We could extend the index view to include a ctx span with an OpenURL for each hit as a means to support this request. 

[...]

PR #322, opened 2015-11-10
DO NOT Merge - WIP Add ctx span to index view, add lccn/oclcnum to ctx openurl.

kevinreiss:


jpstroop:
I think this could use a couple of specs

kevinreiss:
I added a few basis feature specs since it is still unclear how much of this functionality will move to the Orangeindex/Traject process (for OpenURL generation) and a standalone/embedded engine for library requests (Copies in Library).

tampakis:
I've been going back and forth on this. The important bit is that the URL contains the params/values that Traject extracts. I'm just not sure if the URL should be compiled in Orangeindex or Orangelight.

kevinreiss:
Okay. The current method is pretty heavy to execute to generate the <span> tag Zotero can scrape off the results. I will re-write the template to not parse the marc and just generate all values from the normalized data in the solr doc. How does that sound? 

I can also address #218 with a further revision of this.

config/initializers/blacklight_initializer.rb
```diff
@@ -65,7 +66,14 @@ def export_as_openurl_ctx_kev(format = nil)
         export_text << "&amp;rft.isbn=#{(isbn.nil? or isbn['a'].nil?) ? "" : isbn['a']}"
       end
     end
-    export_text << "&amp;rft_id=" + (id.nil? ? "" : CGI::escape("pul_"+id.value))
+    
+    export_text << "&amp;rft_id=" + (id.nil? ? "" : CGI::escape("http://bibdata.princeton.edu/bibliographic/"+id.value))
```
tampakis: How does the rft_id value get used?
Stylewise I think it's more readable to interpolate the string: `"http://bibdata.princeton.edu/bibliographic/{id.value}"`

kevinreiss: rft_id is catchall parameter in OpenURL that can be repeated that is supposed to contain either a reference to an identifier like a DOI, LCCN, or OCLCnum or just a basic encoded URL for the resource in question. It is up to the OpenURL resolver to do something with them.

kevinreiss:
Still working on this per earlier comment from @tampakis about how to construct the OpenURL. Don't merge.

kevinreiss:
Hey, I tried working through generating a good citation from the solr index and in order to do that effectively we are going to have make some granular field definitions for things related to dates and publications. The MARC data has a richer set of data than we have normalized into the index. We should probably do that in the context of a different branch than this group of changes.

config/requests.yml.tmpl
```diff
@@ -0,0 +1,14 @@
+# requests.yml
+
+defaults: &defaults
+  umlaut_base: https://getit.princeton.edu
+  bibdata_base: http://bibdata.princeton.edu/bibliographic
```
jpstroop: Should any code that uses `http://bibdata.princeton.edu/locations` be refactored to use config/constants from this file? Or is that all Javascript? js.erb? or is that bad practice?

kevinreiss: That is what I was thinking so we have a clean way to point at test versions of external services like bibdata down the road. I think we can  expose config values to the javascript layer directly.

kevinreiss: If we wind up adding a lot of config that needs to be exposed to JS we could create some sort of config object that gets registered in the Blacklight JS namespace. For a few values we could just use data attributes that our custom JS knows to pull when it needs something like this.

Issue #235, opened 2015-09-02
Hide dirty data on the production server


tampakis:
Examples:
 - [blank call numbers](http://pulsearch-dev.princeton.edu/browse/call_numbers)
 - [names without characters](http://pulsearch-dev.princeton.edu/browse/names)
 - [subjects](http://pulsearch-dev.princeton.edu/browse/subjects) seem pretty good. Prevent blank headings from creeping in
 - Exclude [unknown format](http://pulsearch-dev.princeton.edu/?f%5Bformat%5D%5B%5D=Unknown)
 - Exclude dates [later than 2017](http://pulsearch-dev.princeton.edu/catalog?utf8=%E2%9C%93&search_field=all_fields&q=&range%5Bpub_date_start_sort%5D%5Bbegin%5D=2018&range%5Bpub_date_start_sort%5D%5Bend%5D=9999&commit=Limit)
 - Make sure only libraries are included in the [library facet](http://pulsearch-dev.princeton.edu/catalog/facet/location?search_field=all_fields&utf8=%E2%9C%93)

joycebcat:
@tampakis Nikitas, be careful about excluding dates later than 2017.  Anytime you put a block on based on a moving wall there is the danger that you start losing legitimate things.  The exclusion would be best based around the idea of blocking things with dates more than 3 years ahead of the current year.

tampakis:
@joycebcat yes of course. Current year +3

PR #336, opened 2015-11-25
add library facet to homepage, remove access facet

tampakis:
Replace access facet with library facet on homepage. Library facet is collapsed by default and now the first facet to appear on the left.

Added the ability to filter out 'dirty' locations by setting a facet minimum count requirement. Related to #235.


kevinreiss:
:+1:

Issue #346, opened 2015-12-02
Online records without 856s


tampakis:
@joycebcat and @mzelesky Check out https://pulsearch-dev.princeton.edu/catalog/8938641
It doesn't have an 856 associated with it. It's causing an error when the record comes up in the [search results](https://pulsearch-dev.princeton.edu/catalog?utf8=%E2%9C%93&search_field=all_fields&q=8938641) because my code assumes an 856 present in online holding locations. I need to make sure that the code still works when that isn't the case, but I think we also need to make sure that there actually is an electronic access link for our online access records.

joycebcat:
@tampakis @mzelesky Nikitas, what is going on here is that we have placed an order for an electronic resource but the transaction isn't complete and we don't have a URL yet.  This is mostly going to happen when we order a permanent access title, which is the case here.

tampakis:
Is this an appropriate place holder for those cases: https://pulsearch-dev.princeton.edu/catalog?utf8=%E2%9C%93&search_field=all_fields&q=8938641

joycebcat:
@tampakis Do we have a display chosen for material which is on order?  It makes the most sense to include electronic material marked on order with anything else marked on order.  Whether we will ever have other material where there is an elf mfhd with no link is a question.  If we will have a default display which says "link missing", the text next to it should not include the words "about this error".  Maybe at a meeting we could get additional input on how/to who these should be reported.

tampakis:
We can talk more about this on Tuesday. On order records currently display a green label that says "requestable"

tampakis:
Online records without 856s no longer throw errors and the text/label displayed (as of #353) is currently:
```
Link Missing: Online access is not currently available.
```

PR #348, opened 2015-12-02
allow search results to render when 856 is missing

tampakis:
Advances #346


kevinreiss:
:+1:

app/helpers/application_helper.rb
```diff
@@ -119,13 +119,19 @@ def holding_block_search args
...
-      render_arrow = !holding['library'].blank? and !holding['call_number'].blank?
+      render_arrow = (!holding['library'].blank? and !holding['call_number'].blank?)
       arrow = render_arrow ? ' &raquo; ' : ''
       info = ''
       if holding['library'] == 'Online'
-        info << content_tag(:span, 'ONLINE', class: 'availability-icon label label-primary',
-                            title: 'Electronic Access')
-        info << links.shift.html_safe
+        if links.empty?
+          info << content_tag(:span, 'LINK MISSING', class: 'label label-danger',
```
kevinreiss: @tampakis If we are rendering something with the value "Link Missing" should we be logging this somewhere so it can be corrected by cataloging. I'm assuming that "Link Missing is an error.

PR #358, opened 2016-01-06
linked records, 246 labels, travis update

### LINKED ISSUE #44: 246 should have labels based on indicators and $i
See  http://www.loc.gov/marc/bibliographic/bd246.html

Should probably follow the logic of the 1st indicator as explained, use 2nd indicator for label, or `$i` if it is present (prefer `$i`)

### LINKED ISSUE #274: Linking to other records via 773/774
I'm looking the Voyager config had 2 additional fields that would link to other records (similar to the other verisions field):
```
Contained in:
   3500 BBID773W

Related record(s):
   3500 BBID774W
```

These fields are currently not included in Orangelight (as links to other records). Should we add it?

### LINKED ISSUE #331: Title subfield h and indexing
@tampakis @jpstroop The title starts with search includes |h when searching.  This has long been a problem in our discovery systems.  |h appears in non-book format material before RDA in between the title and the subtitle.  It is a pain for users because they don't expect it and the print material is retrieved in searches but other formats are not.  Ideally the title starts with search would skip over the |h entirely.  You should be able to get this record:
Behind the beautiful forevers [electronic resource] : life, death, and hope in a Mumbai undercity 
with this title starts with search:
behind the beautiful forevers life

tampakis:
- the travis config update should allow tests to run faster
 - include links to bibs provided in 773/774 provided that they actual correspond to a live record ([example that doesn't](http://catalog.princeton.edu/cgi-bin/Pwebrecon.cgi?BBID=3388644)). Closes #274.
 - Update solr config to include 245 titles without $h for the left-anchored search. Closes #331.
 - Uses label described by 246 2nd indicator or $i when present. Closes #44. Not #246.


tampakis:
The travis update seems like a good improvement. Not sure how much of it has to do with off-peak commits, but the travis started the build right after I made the PR and ran 30 seconds faster than the earlier PR today.

Issue #373, opened 2016-01-21
Update Devise


kevinreiss:
Per http://rubysec.com/advisories/CVE-2015-8314/.

PR #375, opened 2016-01-21
Rails 4.2.5, devise 3.5.4. Closes #374 and #373.

### LINKED ISSUE #374: Rails 4.2.5
Upgrade to Rails 4.2.5

kevinreiss:


kevinreiss:
The reason for the failures is the Faraday gem in the project moved up a version and the user agent string in the Stub requests is no longer correct. I'll see if that can be mocked in a way that doesn't pin those to something like that.

PR #383, opened 2016-01-29
Prep app for new relic monitor.

kevinreiss:
Add new relic gem and config file template. Set up as a shared file for deployment purposes.


tpendragon:
Yay!

tampakis:
:+1:

PR #386, opened 2016-02-01
linked 776/787 bib ids

### LINKED ISSUE #376: bib linking profiles not present 787 and 776
@tampakis Hi Nikitas, the equivalent of bib linking which you have done in BL is OK except that two profiles we have in Voyager are missing.  

787W to bib ID
and
776W to bib ID

Example record for the first profile is: 4153328 and for the second profile is: 4178285

tampakis:
Closes #376. Will take effect on full reindex.


tampakis:
Updated icon name and css class to reflect format label change from thesis to dissertation.

kevinreiss:
:+1:

PR #400, opened 2016-02-11
on shelf availability, disable facet enum cache, specify deploy branch

tampakis:
- "Uknown" records default to "On Shelf" status and display as available in the interface.
 - Made advanced and default search handler consistent, disable facet enum cache (which was causing performance issues - never checked in to version control)
 - Allow user to specify branch for cap deployment. Instructions in updated readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,15 +9,15 @@
 Versions:
 
 * Ruby: 2.2.3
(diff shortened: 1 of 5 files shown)


kevinreiss:
:+1:

PR #402, opened 2016-02-12
provide full item availability

tampakis:
It doesn't look the prettiest, but if a mfhd has multiple items, each item's availability gets pulled in on the record show. @axamei we should probably work on improving the side availability panel appearance next week

### DIFF
--- a/app/assets/javascripts/availability.js.coffee
+++ b/app/assets/javascripts/availability.js.coffee
@@ -13,7 +13,8 @@ class AvailabilityUpdater
     'Remote Storage Request', 'Hold Request', 'Recall Request']
   missing_statuses = ['Missing', 'Lost--Library Applied',
     'Lost--System Applied', 'Claims Returned', 'Withdrawn']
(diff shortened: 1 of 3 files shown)


kevinreiss:
Hi, I believe that @tpendragon has been shifted the js work over to es6 in the plum project. Is that something that we want to consider for OL too instead of using coffeescript. I think that over time that is going to be the preferred syntax to use over coffeescript.

app/assets/javascripts/availability.js.coffee
```diff
@@ -13,7 +13,8 @@ class AvailabilityUpdater
     'Remote Storage Request', 'Hold Request', 'Recall Request']
   missing_statuses = ['Missing', 'Lost--Library Applied',
```
kevinreiss: Is it worth pulling these statuses from a config file rather than hardcoding them?

tampakis:
I would be fine working with es6 over coffeescript. The problem is that I really don't know any flavor of js, and was just expanding on the code that was already in Orangelight. I also don't know how to load config files.

kevinreiss:
I think we can tag those both as future issues to look at.

PR #408, opened 2016-02-23
add current issues in physical holdings when journal

tampakis:
Replicates Voyager's "Location has (current)" functionality for serials


kevinreiss:
:+1: Good addition. This should please library staff.

PR #410, opened 2016-02-25
multivolume on-site access records should have limited status

tampakis:
This [record](https://pulsearch.princeton.edu/catalog/337937) should display "On-Site Access" for the holding and not show the individual items since they should always be on-site (individual copies/volumes only display if they are unavailable).


kevinreiss:
Not really sure why a zero percent change causes a failure @tampakis. Merging :+1:

PR #425, opened 2016-03-17
include lc authorized pu department in name browse

tampakis:


kevinreiss:
I can't really see why coveralls is complaining.

PR #439, opened 2016-04-14
Add rubocop

kevinreiss:
@tampakis I took a stab at getting this started on the recent account stuff committed today.  It's going to be a long-term project. I used the rubocop.yml from plum.


eliotjordan:
This beast is ready for review. Not all of the rubocop issues were resolved, as you can see by the sizable settings file, but most of the low hanging fruit is addressed. Thankfully, many of the more difficult issues are concentrated in a few areas:

```	
app/controllers
app/helper
spec/
```

Addressing those is beyond the scope of this PR. In fact, I can see the need for a dedicated refactoring sprint at some point. Also, I didn't hook rubocop into the rspec tests because the Rakefile is need of some TLC. I made an [issue for that.](https://github.com/pulibrary/orangelight/issues/510) 

@tampakis @kevinreiss

kevinreiss:
It seems like there are some obvious syntax/whitespace errors left. Should we fix those in this PR and leave more complex problems for a future refactoring sprint?

eliotjordan:
@kevinreiss Is there an example you can point to? PR passes all white space cops as far I can tell.

eliotjordan:
I'd suggest we leave more complex issues for another pull request. This thing is big enough. We also don't want this to hang around much longer. Rebases against this branch won't be fun.

kevinreiss:
@eliotjordan I ran rubocop on an old copy sorry about that. This branch looks good to me. The commits seem pretty atomic to me. I'll defer to @tampakis if he'd prefer it rebased down to one commit. I'm thinking probably yes.

Issue #447, opened 2016-04-19
Select options next to search box not rendering on accounts page


axamei:
![selectseach](https://cloud.githubusercontent.com/assets/8161144/14648171/e8e7aa36-062e-11e6-8006-45c16e46cec0.png)

Search options should be consistent.

kevinreiss:
@axamei I think we can close this one.

axamei:
Fixed by #448

PR #448, opened 2016-04-19
Addresses #447.

kevinreiss:
Mixing in Blacklight::Configurable flipped the site to use the blacklight.html.erb layout file instead of application.html.erb.


kevinreiss:
Okay, other side effects ensued by removing that include.

kevinreiss:
Okay, it turns out you don't need to reference either "Blacklight::Configurable" or call the layout directly.

PR #452, opened 2016-04-19
Add success/danger attributes to renewed item row 

### LINKED ISSUE #449: Row Display for Renew Items Should reflect failure/success
The "message" cell for each item selected for renewal comes back with a "Renewed" or "Not Renewed" value. We can use these to assign a specific CSS selector to the row in question so it can be color coding for failure or success. We'll need a helper to scrape the string of the the "status" element in Voyager's XML response. See https://github.com/pulibrary/orangelight/blob/development/spec/fixtures/failed_renew_request_recall.xml#L48 for an example.

kevinreiss:
Should close #449. I believe "danger" is the universal bootstrap class attribute for "red background".

### DIFF
--- a/app/helpers/account_helper.rb
+++ b/app/helpers/account_helper.rb
@@ -42,5 +42,14 @@ module AccountHelper
       content_tag(:b, "#{item[:renew_status]["status"]}")
     end
   end
-  
+
+  def renew_state item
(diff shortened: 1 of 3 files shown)


tampakis:
Cool!

PR #462, opened 2016-04-27
Toggles the 'request' button on

kevinreiss:
Shows request button when item is in a closed location or when an item is in an open location but has an unavailable availability label applied to the holding. Also uses the old request link base for the button destination for now.


tampakis:
Looks good

PR #475, opened 2016-04-29
Add google analytics. Closes #474

### LINKED ISSUE #474: Add Google Analytics
Add basic tracking code. Once it is there create a separate  issue to discuss custom events to track.

### LINKED ISSUE #486: Google Analytics Tracking
Put in place code that will support google analytics custom even tracking. As a first pass adding for use of the stackmap links.

axamei:
Also closes #486

### DIFF
--- a/app/assets/javascripts/google_analytics_events.js.coffee
+++ b/app/assets/javascripts/google_analytics_events.js.coffee
@@ -1,0 +1,8 @@
+$ ->
+  # Track if user is clicking on stack map pin icon in results page
+  $('#documents .find-it').on 'click', ->
+    ga('send', 'event', 'Availability', 'Search Results', 'StackMap Pin')
+
+  # Track if user is clicking on stack map pin icon in show page
+  $('#availability .find-it').on 'click', ->
+    ga('send', 'event', 'Availability', 'Item Record', 'StackMap Pin')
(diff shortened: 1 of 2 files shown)


kevinreiss:
Do we want to use the Universal Analytics ```ga.send()``` syntax instead of ```_gaq.push()``` syntax. I think the latter is deprecated and will stop working at some point, though UA has been out for well over a year and they still haven't cut off the old style syntax.

axamei:
@kevinreiss we should stay away from deprecated `_gaq.push()` syntax. I updated it to use the Universal Analytics syntax.

kevinreiss:
It might be good to avoid hardcoding the analytics string into the project code. What do you think @tampakis? We could print out a configured value into a data attribute in an erb template and have the coffescript load that as the tracking code value.

tampakis:
Yeah I think we should load the analytics string from environment variable. I never knew you could make coffee files into erb: http://stackoverflow.com/questions/16883711/how-to-access-environment-variables-in-coffeescript

tampakis:
I just configured it on the pulsearch server. We should add it in the coffeescript as:
`<%= ENV['GOOGLE_ANALYTICS'] %>`

kevinreiss:
Coffescript can read these values? Nice.

app/assets/javascripts/google_analytics_events.js.coffee
```diff
@@ -0,0 +1,4 @@
+$ ->
+  # Track if user is clicking on stack map pin icon
+  $('.find-it').on 'click', ->
+    ga('send', 'event', 'Stack Map Pin','Clicked')
```
kevinreiss: I think the we can add three unique values here. We are using Action and Category here, but but we can also add a third Label value. It would be good to add something contextual there that denotes if the action happened on a record page or on a search results page.

Issue #468, opened 2016-04-28
Create Feedback Form


kevinreiss:
Should reflect the context of the OL page form is invoked from.

PR #479, opened 2016-05-02
Add feedback form. 

kevinreiss:
Addresses #468. Need to set CC and TO values for form emails in app_configs.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -68,6 +68,8 @@ gem 'coveralls', require: false
 
 gem 'newrelic_rpm'
 
+gem 'mail_form'
+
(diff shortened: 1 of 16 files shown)


config/requests.yml
```diff
@@ -2,14 +2,15 @@
 
 defaults: &defaults
   umlaut_base: https://getit.princeton.edu
-  bibdata_base: https://bibdata.princeton.edu
```
tampakis: Don't update this file in the PR

kevinreiss: Doh!

PR #504, opened 2016-05-05
Auto dismiss alert-info and alert-success elements in flash_messages

### LINKED ISSUE #503: Do not dismiss/hide elements with alert-info / alert-success classes in browse
Contain auto-dismiss action to only elements inside flash_messages

axamei:
Select only alert-info and alert-success elements to auto dismiss in flash_messages container. Closes #503


tampakis:
Thanks for the fix.

PR #509, opened 2016-05-09
Fix print stylesheet for IE

axamei:


kevinreiss:
👍 The water is safe for printing now.

PR #512, opened 2016-05-11
Buttonize pager on top of search results and in item record

axamei:


tampakis:
The vertical alignment looks a bit off to me. I think the prev/next button is on a different plane from the sort by/per page buttons and from the bookmark all button on the search results (the prev/next vertical alignment being off might be my imagination, but the bookmark all is definitely noticeable):
screen shot 2016-05-11 at 4 56 00 pm

On the record show the prev next is higher up than the back to search button:
screen shot 2016-05-11 at 4 55 12 pm

Also the bookmark button appears higher than the cite/send to button:
screen shot 2016-05-11 at 4 55 33 pm

PR #513, opened 2016-05-11
Update Rakefile and travis config

### LINKED ISSUE #510: Cleanup Rakefile
The Rakefile could use a spruce-up. As part of that, maybe we should use rake ci as the default hook for running the test suite. Right now it doesn't work work as it should.

eliotjordan:
- cleans up Rakefile
- fixes `rake ci` and makes it the default command for travis
- hooks rubocop into tests

closes #510

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,18 +2,13 @@ language: ruby
 sudo: false
 rvm:
   - 2.3.0 # Ruby most recent
-# uncomment this line if your project needs to run something other than `rake`:
+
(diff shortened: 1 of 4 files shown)


eliotjordan:
This is ready for review. I took a stab at updating to solr_wrapper, but quickly realized that involves upgrading to solr 5. I'll post that work as a WIP PR later.

tampakis:
Sweet sauce.

PR #514, opened 2016-05-12
Replace jetty with solr_wrapper

eliotjordan:
Replaces jettywrapper with solr_wrapper. To do this, the solr schema was upgraded to version 5. Everything seems to work fine, but I would take a very close look at  `solrconfig.xml` and `schema.xml`. 

Also, solr_wrapper uses solr version 6 by default, which seems to work with a version 5 schema and config.


README.md
```diff
@@ -51,16 +49,33 @@ cap staging deploy # deploys development branch to staging
...
+
+```bash
+rake spec
+```
+
+To run a specific test:
 
 ```bash
-rspec spec
+rake spec SPEC=spec/models/solr_document_spec.rb:27
```
tampakis: Why would we want to run only this specific spec?

eliotjordan: It was meant to be an example. I thought it was better than:

`rake spec SPEC=path/to/your/spec.rb:linenumber`

But maybe not. Actually, that doesn't seem that bad.

tampakis: oops didn't read the context closely enough

eliotjordan: What do you think? Leave it alone, or insert the more generic version?

eliotjordan:
Made the README a little easier to read. Test coverage is back up.

PR #540, opened 2016-05-18
request button links directly to live request system

### LINKED ISSUE #539: Disable request mockup page
As demonstrated throughout the user testing, it isn't helpful at all for the user to see mockup forms for the request page while the new system is being developed. Just redirect to the existing request system.

tampakis:
Closes #539


kevinreiss:
Should we also delete the controller, view, and route I added for the demo page? I imagine we will just mount the request engine when it is ready.

tampakis:
Oh I thought that the additional controllers/views/routes were part of the request engine. Just removed them.

kevinreiss:
👍

PR #551, opened 2016-05-24
Add Where to Find It text next to pin in show page availability panel

### LINKED ISSUE #550: Add "Where to Find it" Link on Record Show page
Users were looking for a link with the text "Where to Find it" during the usability testing last week.

axamei:
Add Where to Find It text next to pin in show page availability panel. Closes #550 
Remove link from  availability label

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -40,7 +40,7 @@ module ApplicationHelper
     if DONT_FIND_IT.include?(library)
       ''
     else
-      ' ' + link_to("[#{t('blacklight.holdings.stackmap')}]".html_safe, "#{ENV['stackmap_base']}?loc=#{location}&id=#{bib}", :target => '_blank', class: 'find-it', 'data-map-location' => location.to_s, 'data-toggle' => 'tooltip')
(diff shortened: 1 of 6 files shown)


app/helpers/application_helper.rb
```diff
@@ -40,7 +40,7 @@ def locate_link(location, bib, library = nil)
     if DONT_FIND_IT.include?(library)
       ''
     else
-      ' ' + link_to("[#{t('blacklight.holdings.stackmap')}]".html_safe, "#{ENV['stackmap_base']}?loc=#{location}&id=#{bib}", :target => '_blank', class: 'find-it', 'data-map-location' => location.to_s, 'data-toggle' => 'tooltip')
+      ' ' + link_to('<span class="glyphicon glyphicon-map-marker"></span> <span class="map-text">Where to Find It</span>'.html_safe, "#{ENV['stackmap_base']}?loc=#{location}&id=#{bib}", :target => '_blank', class: 'find-it', 'data-map-location' => location.to_s, 'data-toggle' => 'tooltip')
```
kevinreiss: @axamei I think we should probably make it so that the "Where to Find It" phrase is read from a key the config/locales/blacklight.en.yml file.

kevinreiss: It is actually already there https://github.com/pulibrary/orangelight/blob/development/config/locales/blacklight.en.yml#L32.

tampakis:
Looks good to me!

PR #569, opened 2016-06-01
Replacing getdir methods with extracted string_rtl gem

### LINKED ISSUE #454: Use string_rtl gem for determining script direction
Gem is also used in https://github.com/pulibrary/pomegranate/pull/76/files

escowles:
Closes #454

### DIFF
--- a/app/views/catalog/_index_default.html.erb
+++ b/app/views/catalog/_index_default.html.erb
@@ -5,13 +5,13 @@
   <% index_fields(document).each do |solr_fname, field| -%>
     <% if should_render_index_field? document, field %>
       <% if document[solr_fname].length == 1 || !document[solr_fname].is_a?(Array) %>
-        <% valdir = getdir(document[solr_fname])%>
+        <% valdir = Array(document[solr_fname]).first.dir %>
         <% value = render_index_field_value document, :field => solr_fname %>
         <li class="blacklight-<%= solr_fname.parameterize %>" dir="<%= valdir %>"><%= value %></li>
       <% else %>
         <% valdirs = [] %>
         <% document[solr_fname].each do |value| %>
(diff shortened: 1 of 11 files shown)


app/views/catalog/_index_default.html.erb
```diff
@@ -5,13 +5,13 @@
   <% index_fields(document).each do |solr_fname, field| -%>
     <% if should_render_index_field? document, field %>
       <% if document[solr_fname].length == 1 || !document[solr_fname].is_a?(Array) %>
-        <% valdir = getdir(document[solr_fname])%>
+        <% valdir = document[solr_fname].first.dir %>
```
tampakis: This approach won't work for Solr single-valued fields (`!document[solr_fname].is_a?(Array)`) because they are of type String. Can we add an elsif for strings where we call `document[solr_fname].dir` instead?

app/views/catalog/_show_default.html.erb
```diff
@@ -16,13 +16,13 @@
           <% end %>
 
           <% if document[solr_fname].length == 1 || !document[solr_fname].is_a?(Array) %>
-            <% valdir = getdir(document[solr_fname])%>
+            <% valdir = document[solr_fname].first.dir %>
```
tampakis: Same as the index view.

app/views/catalog/_index_default.html.erb
escowles: How about just wrapping it in an Array to make sure first will always work?  Seems like less duplicate code that way:

`valdir = Array(document[solr_fname]).first.dir`

tampakis: Yeah that works too and is cleaner.

PR #588, opened 2016-06-08
advanced location facet with libraries and recap locations

### LINKED ISSUE #494: See all items in a Library's collection even if at ReCAP
From Julie M.:

> https://pulsearch.princeton.edu/catalog/9230299
> The RBSC recap material is coded as 'recap' not rarebooks and so, you need to do two searches to see if we have a title in rbsc. The link above is one example.

This further blurs the distinction between a physical location and a logical collection, but we're already on that road by having "Rare Books ..." as a library, so I don't think that's an issue. I do think there could be some merit to mapping the recap location codes to holding libraries and allowing users to flag whether or not offsite storage locations should be included in results.

tampakis:
Enhances advanced holding location facet to group holding locations and recap locations by library. Closes #494.
Fixes advanced search bug where submitting only the value in the third query field would error.


kevinreiss:
@tampakis Excellent new feature.

PR #616, opened 2016-06-23
location label based on code lookup against location service data

### LINKED ISSUE #617: Item availability label should have fallback when enum is null
For item records with no [enum or chron](https://bibdata.princeton.edu/availability?mfhd=8532183), [Orangelight](https://pulsearch.princeton.edu/catalog/8803295) displays the item label as "Undefined." Set a fallback label (like "Item").

tampakis:
Pull the holding location labels directly from the location service data (rather than the solr doc). Allows records like https://pulsearch.princeton.edu/catalog/9287769 to render updated holding location labels without reindexing.

Second commit closes #617. Fallback label for item enum when nil.


kevinreiss:
Thanks @tampakis.

PR #619, opened 2016-06-27
update labels and field ordering

### LINKED ISSUE #278: Rare Books and Copy Number Examples
Here are two bibliographic IDs where copy notation matters:

616086 (Copy Two)
2478499 (Copy Five)

We might want to add these to the test data set for future development. What can we do to make the request of these with proper item level information in context. If you look up the bib IDs in current systems we don't do too well.

### LINKED ISSUE #325: Reconfigure note fields
From Don:
>
I went through the Princeton tag table and MARC formats to supplement them.
From our tag tables, 551 and 552 are missing.  I bet we don't even have any of them!
542: Never even heard of that.
570: Obsolete in 1990.

>
Suggested changes
502       abcdgo
518       3adop
534       (add subfield 3)
538       3ai
583       Never display.  Regardless of Ind1 value, it's always nonpublic information as far as I know.
584       (add subfield 3)
586       (add subfield 3)
505       I suggest merging them all as "Contents."  I don't think indicator usage supports the distinctions, and I don't know that users care.  (If they do care then they are kind of out of luck.)
561       796 and 797 should not be included, if indeed we have any of these non-standard ancient fields.  Generally speaking, I don't think we should mix entities and strings--even if the entities are so weird.
533       Don't include with 530.  It's already in "reproductions note."
555       The two types of related documents are supposed to be distinguished by Ind1 values, as in BIBs 359280 and 823463.  Maybe display can be configured on that basis, so there would be 2 separate note labels for 555 as specified in MARC.
590       How about adding to 500?  It's the same kind of content.  What was 590 is now 500 with subfield 5.
546       3ab

### LINKED ISSUE #589: Change cumulative index finding aid label
Change 555 label to "Cumulative index/Finding aid"

### LINKED ISSUE #592: Change the display order of various fields
Group 525, 770, and 772 fields together, in that order, and slightly lower in the list of fields. Feedback:
>supplement notes split up in display
>example: [Bulletin of labour statistics](https://pulsearch.princeton.edu/catalog/492688)

>Fields 525, 770, and 772 all concern supplements -- 525 is a general supplement note, 770 & 772 link supplement and parent when the supplement is cataloged as a separate serial. Some records, as the above example, will have both types of note. The Beta catalog displays the 770 rather high in the record, the 525 after several other fields. It would seem preferable to have these grouped together if possible, with 525 preceding 770.

Move 130 field higher up. Feedback:
> 130 not appearing prominently
> example: [New York times (Swollen Magpie Press)](https://pulsearch.princeton.edu/catalog/8293256)

> In the Voyager OPAC, uniform titles were displayed at the very top of the record. We feel that this prominence is in some cases very helpful, as the parenthetical qualifier serves as a clue that there may be other publications with the same title; see example.

### LINKED ISSUE #593: 515 note displayed twice
Display 515 note directly under description field and remove the 515s from the description field.
>515 notes repeated
>example: [Guitar player](https://pulsearch.princeton.edu/catalog/426704)

>This particular record has three 515 notes (numbering peculiarities). We notice that the Beta catalog shows these notes twice, once as part of "Description" and again under their own label. We think the separate heading is helpful for this category of notes -- it's a bit crowded and confusing to have them immediately under the 362 -- but the position NEAR the 362 is a good idea; could they be kept under "Numbering peculiarities" but moved immediately under "Description"?

tampakis:
- Updated note fields and labels. Closes #325.
 - Reordered fields as per feedback. Closes #592. Closes #593.
 - Made sure all show labels are sentence-cased. Closes #589.
 - Includes copy number with holdings. Closes #278.
 - `&#8203;` character allows for better word splitting on smaller screens.


spec/features/availability_spec.rb
```diff
@@ -36,17 +36,4 @@
       expect(page.all('.location--online .umlaut .fulltext .response_item').length).to be >= 4
     end
   end
-
```
kevinreiss: @tampakis Are temp location displays tested somewhere else?

tampakis: The record that was tested here no longer has a temp location (not on reserve anymore). I don't know how to stub the Ajax call.

spec/features/availability_spec.rb
```diff
@@ -38,12 +38,12 @@
   end
 
   describe 'Physical Holdings in temp locations', js: true do
-    it 'displays temp location on search results' do
+    xit 'displays temp location on search results' do
```
tampakis: Skipping, rather than removing the failing specs, until we can get a better way to test the availability javascript code (which in general really isn't tested well).

PR #628, opened 2016-07-11
Update requests gem ref. Closes #499. Redirect back to request form when appropriate.

### LINKED ISSUE #499: Use MARCIT OpenURL to link to Query SFX API
For online titles that come from the Marcit service use the OpenURL stored in the 856 field rather than the one in the blacklight ctx span tag to query the SFX API to fetch full-text links for display in the 'Online' section. See https://pulsearch.princeton.edu/catalog/8614008 for an example. 7 full-text options should appear instead of two.

kevinreiss:
Closes #499.


kevinreiss:
@tampakis I think this one can be merged. Ran into some weirdness with different rspec behavior in user logins when run within the context of ```rake ci``` vs. ```rspec spec/features/login_account_spec.rb```. Hence all the commented out code in ``` spec/features/login_account_spec.rb```.

PR #638, opened 2016-07-29
Add Course Reserves Functionality

tpendragon:
The workflow looks like this:

* On course reserves page, query bibdata for reserves and display a list
  of dropdowns which set pseudo-facets.
* When the pseudo-facets come through, strip them from the query and
  hot-index reserve records which match the query parameters into a
  secondary core.
* Use the hot-indexed records to do a join query and display only the
  bibs that match those reserve parameters.

### DIFF
--- a/app/helpers/blacklight_helper.rb
+++ b/app/helpers/blacklight_helper.rb
@@ -31,6 +31,27 @@ module BlacklightHelper
     solr_parameters['facet.pivot'] = []
   end
 
+  def course_reserve_filters(solr_parameters, user_parameters)
+    return unless user_parameters[:f]
+    instructor = Array(user_parameters[:f][:instructor_name]).first
+    course = Array(user_parameters[:f][:course]).first
+    department = Array(user_parameters[:f][:department]).first
+    filter = Array(user_parameters[:f][:filter]).first
+    return if instructor.blank? && course.blank? && department.blank? && filter.blank?
+    courses = CourseReserveRepository.all.query(instructor: instructor, course_with_id: course, department_with_identifier: department)
+    index_course_reserves(courses, user_parameters)
+    solr_parameters[:fq] ||= []
+    solr_parameters[:fq].reject! do |x|
+      %w(course instructor_name department filter).include?(x.split('=')[1].split('}')[0])
+    end
+    solr_parameters[:fq] << "{!join from=bib_ids_s to=id fromIndex=#{ReserveIndexer.core}}#{courses.solr_query}"
+  end
+
+  def index_course_reserves(courses, _user_parameters)
+    ReserveIndexer.connection.delete_by_query(courses.solr_query)
+    ReserveIndexer.index!(courses)
+  end
+
   # Returns suitable argument to options_for_select method, to create
   # an html select based on #search_field_list with labels for search
   # bar only. Skips search_fields marked :include_in_simple_select => false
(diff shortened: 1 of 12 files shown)


mzelesky:
I expected the 'Edit Search' button within a Reserve search to take me back to the Course Reserve selection page, with the previous reserve info pre-populated, instead of to the standard Blacklight Advanced Search. Is this possible?

app/helpers/blacklight_helper.rb
```diff
@@ -31,6 +31,33 @@ def only_advanced_facets(solr_parameters, user_parameters)
...
+
+  def index_course_reserves(courses, user_parameters)
+    return unless first_reserve_request(user_parameters)
+    ReserveIndexer.connection.delete_by_query(courses.solr_query)
+    ReserveIndexer.index!(courses)
+  end
+
+  def first_reserve_request(user_parameters)
+    return if user_parameters[:page]
+    return if user_parameters[:f] && !user_parameters[:f].except(:instructor_name, :department, :course).keys.empty?
```
tampakis: I think the course reserves reindex if a user enters a query [:q]

tpendragon:
> I expected the 'Edit Search' button within a Reserve search to take me back to the Course Reserve selection page, with the previous reserve info pre-populated, instead of to the standard Blacklight Advanced Search. Is this possible?

I think there's a general question here - should buttons in the context of the course reserves perform differently? It seems to make sense in this case, but it means we should probably have separate routes for course reserve searches (meaning a second catalog controller). What do you think @tampakis?

app/helpers/blacklight_helper.rb
tpendragon: Yeah. I'm not even sure if I want to keep this here though or just let it reindex every time. The performance difference is pretty minimal and I'm somewhat concerned about live updates only showing up when you go through the course reserves first page - bookmarks and such would be troublesome.

tampakis:
Separate routes definitely make sense. Also a better indication in the search results that the search is scoped within course reserves. I think this and other UI concerns can be addressed in a separate PR and maybe @axamei  can help us out - for instance there isn't an actual clickable way to get to the course reserves.

app/helpers/blacklight_helper.rb
tampakis: Performance is good enough that we probably can get away with indexing on every query.

jpstroop:
I struggled with that too. When you're not thinking as someone who is testing or reviewing reserves functionality, I think you expect the buttons to behave consistently. I might be wrong. Maybe we're overloading "Start Over" and we need yet another link back to the Reserves page?

tpendragon:
> for instance there isn't an actual clickable way to get to the course reserves.

Yeah, this was left out on purpose - I figured you and the orangelight group would know best where to put it.

tpendragon:
> a better indication in the search results that the search is scoped within course reserves.

Are the facets not sufficient?

app/repositories/course_reserve_repository.rb
```diff
@@ -0,0 +1,131 @@
...
+  Course = Struct.new(:reserve_list_id, :department_name, :department_code, :course_name,
+                      :course_number, :section_id, :instructor_first_name,
+                      :instructor_last_name) do
+    attr_accessor :bib_ids
+    def instructor
+      "#{instructor_last_name}, #{instructor_first_name}"
+    end
+
+    def department_with_identifier
+      "#{department_code}:#{department_name}"
```
tampakis: Can you add a space after the colon? I don't know why Voyager doesn't add the space, but adding would be consistent with the course list dropdown.

PR #652, opened 2016-08-10
Add solr wrapper rake tasks

eliotjordan:
Adds various solr wrapper rake tasks. Sometimes things get janky and solr wrapper won't start because 'there is already a blacklight-core'. Simply run `rake solr:clean` et voilà.


kevinreiss:
👍 Helpful command......

PR #653, opened 2016-08-10
Strip question mark from solr query parameters

### LINKED ISSUE #606: Solr treats question mark as a wildcard character
From Beta feedback:
> I had a small problem with colons in my search. 
Doing a keyword search for "China and Angola: a marriage of convenience" gives me zero results. On the other hand, searching for "China and Angola : a marriage of convenience?" gives me the results I was looking for. Ideally both I should be able to return hits for this book whether or not there is a space between the colon.

The issue with colons is addressed in pul_solr#12, but a second issue also causes zero results to be returned. It's the question mark at the end of the query. With the edismax parser, `?` is interpreted as a single wildcard character. So while the word "convenienced" would match, the original word "convenience" doesn't match (punctuation marks get stripped at index time).

eliotjordan:
The question mark is a single-character wildcard in the edismax parser and should be stripped from the solr query parameters. An [example link](https://pulsearch.princeton.edu/catalog?utf8=%E2%9C%93&search_field=left_anchor&q=China+and+Angola%3A+a+marriage+of+convenience%3F).

Closes #606


app/controllers/catalog_controller.rb
```diff
@@ -49,9 +49,6 @@ def lccn
     # new way of calling local document presenter
     config.document_presenter_class = PrincetonPresenter
 
-    config.navbar.partials.delete(:search_history)
-    config.add_nav_action(:reserves, partial: 'course_reserves/nav')
```
tampakis: Can you keep lines 52 and 53 in?

eliotjordan: Not sure what happened there. I'll fix.

PR #654, opened 2016-08-11
Change position of browsable exact match

### LINKED ISSUE #627: Show 2 browse entries before exact match instead of 3
Update the [browsables_controller](https://github.com/pulibrary/orangelight/blob/development/app/controllers/orangelight/browsables_controller.rb#L42) to start the browse view from 2 entries before the matching entry. For some longer entries in the call number browse table, the query match appears far down on the page.

eliotjordan:
When browsing, the exact match record is now three from the top. Closes #627.

### DIFF
--- a/spec/requests/orangelight/browse_spec.rb
+++ b/spec/requests/orangelight/browse_spec.rb
@@ -1,10 +1,6 @@
 require 'rails_helper'
 
 RSpec.describe 'Orangelight Browsables', type: :request do
-  before(:all) do
-    system 'rake db:seed'
-  end
-
   describe 'Redirect Tests for Browse' do
     it 'browse_name redirects to browse search' do
       get '/catalog?search_field=browse_name&q=velez'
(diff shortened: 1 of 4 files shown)


spec/requests/orangelight/browse_spec.rb
```diff
@@ -54,7 +54,7 @@
       get '/browse/call_numbers.json?q=microfilm'
       r = JSON.parse(response.body)
       q_normalized = StringFunctions.cn_normalize('microfilm')
-      expect(r[3]['sort']..r[4]['sort']).to cover q_normalized
+      expect(r[2]['sort']..r[3]['sort']).to cover q_normalized
```
tampakis: I think this should be r[1] and r[2] since the position has been moved up by 2.

eliotjordan:
Having some hugely frustrating issues with tests passing in my local instance and failing on Travis. Working on a fix.  :confounded:

eliotjordan:
Ok. That did it. Not sure why Travis tests passed previously, but I added `rake db:seed` to the config.

tampakis:
Oh I figured out the issue - it's this: https://github.com/pulibrary/orangelight/blob/development/spec/requests/orangelight/browse_spec.rb#L5
So in the dev instance, the DB was seeded twice. Could you remove that line in this PR as well?

tampakis:
I think by removing the second seed, you'll have to change to r[1] and r[2]

eliotjordan:
@tampakis No change to the position is needed. It makes sense. The first microfilm document is fourth in the array r[3]. The addition of the ` Your query "microfilm" would be here.` row skews the positioning compared with browsing using an actual call number. It all looks fine to me though.

tampakis:
Opened issue #655 about the weird ordering of the call number browse search in dev/test.

Issue #585, opened 2016-06-07
Set CAS timeout to be shorter for public terminals


jpstroop:
The committee thinks 30 minutes is too long (even though people stay logged into email, google, facebook, etc. at the same machines). See if there's a glob pattern for machine names, or else a registry of IP addresses :disappointed:

tampakis:
```ruby
# config/initializers/session_store.rb
Rails.application.config.session_store :cookie_store, key: '_orangelight_session',
                                                      expire_after: 30.minutes
```
30 minutes is too long for public terminals and too short for personal machines. Removing timeout until a better solution is found.

PR #659, opened 2016-08-12
Barcode login

tampakis:
- Adds omniauth "strategy" for barcode login. Allows users without a valid net id (guest patrons) to login to check their account.
 - Relaxed the requirement for emails to be unique (in theory we can now let users set a preferred email address).
 - Users logging out from CAS get redirected to Princeton CAS logout page, but Barcode users get redirected to catalog home page
 - Users with a valid net id that try logging in via barcode get redirected to cas instead


config/initializers/session_store.rb
```diff
@@ -1,4 +1,4 @@
 # Be sure to restart your server when you modify this file.
 
-Rails.application.config.session_store :cookie_store, key: '_orangelight_session',
-                                                      expire_after: 30.minutes
+# Rails.application.config.session_store :cookie_store, key: '_orangelight_session',
+#                                                       expire_after: 30.minutes
```
eliotjordan: Is there a reason to keep this file?

tampakis: It's there as a reminder for https://github.com/pulibrary/orangelight/issues/585

db/schema.rb
```diff
@@ -66,6 +66,15 @@
...
 
+  create_table "orangelight_name_titles", force: :cascade do |t|
+    t.text    "label"
+    t.integer "count"
+    t.text    "sort"
+    t.string  "dir"
+  end
+
+  add_index "orangelight_name_titles", ["sort"], name: "index_orangelight_name_titles_on_sort", using: :btree
+
```
eliotjordan: Is this table part of the barcode / cas login? If not, perhaps it should move to a different PR.

tampakis: Oh good catch!

eliotjordan:
Nice work!  I really like the concise way you write some of the specs. Especially the allow statements. I imagine this will go into pulmap/lime/turnip sometime soon. A few things:
  1. How about increasing the [bibdata.rb](https://coveralls.io/builds/7430293/source?filename=lib%2Forangelight%2Fbibdata.rb#L3) test coverage to 100 by adding a `500` status test and a `ConnectionFailed` test to the [account controller spec](https://github.com/pulibrary/orangelight/blob/development/spec/controllers/account_controller_spec.rb#L45)? It seems like as a good time as any since you pulled out logic to create this new module.
  1. It would be nice if the barcode tooltips had more specific messages. E.g. `Please fill in your last name` and `Please enter a valid barcode`. Not a deal breaker, but it would be good to at least make an issue for it.

config/initializers/session_store.rb
eliotjordan: Put the snippet in the issue as a reminder instead?

tampakis:
@eliotjordan - addressed the coverage for the get_patron method and opened #663.

eliotjordan:
👍

PR #666, opened 2016-08-16
Upgrade to Blacklight 6

eliotjordan:
Ongoing work to upgrade Orangelight to Blacklight 6.


eliotjordan:
@tampakis Ready for you. I think fixing the failing advanced search test will push coverage back to normal.

PR #677, opened 2016-08-29
update home and about pages for production

jpstroop:
@kevinreiss I think you could use some variation on the text in `app/views/pages/about.html.erb` for the news item.


kevinreiss:
Okay @jpstroop. I'll take this and draft something to show to Marvin. Not sure why this build failed. Restarting it.

tampakis:
Solr_wrapper needs to be updated. It is in #666. Just rebasing now.

Issue #805, opened 2016-12-28
Disable action_view logging in production


tampakis:
The log info about rendering html partials is clogging up the logs.
Set config.action_view.logger to nil in production for:
 - [ ] bibdata
 - [ ] orangelight

PR #806, opened 2016-12-28
Production logs

tampakis:
Advances #805


tampakis:
yeah i should have run rake db:migrate on #801, but i forgot.

PR #814, opened 2017-01-03
fixes action_view logger nil error

tampakis:
Specifies logger in debug statements to prevent errors due to https://github.com/pulibrary/orangelight/commit/1e937344e629592d7122dc2933cb70d4bace1d94


tpendragon:
This seems odd. Why turn off action view logging? Too many logs? If so, why not just set the log level higher?

PR #833, opened 2017-01-12
Adds param 'query' to borrow direct route

kevinreiss:
For interoperability with https://github.com/team-umlaut/umlaut_borrow_direct#configure-the-service-in-your-configumlaut_servicesyml.


kevinreiss:
@tampakis the failed spec is an item whose underlying item status has changed. Okay, if we just exempt those tests from CI? Seems like they are in order for a re-factor after the work to offload status labeling moves from availability.js to marc_liberation?

PR #845, opened 2017-01-31
Restrict external full-text calls to books and journals.

kevinreiss:
When other formats pass openURLs the SFX knowledge base produces a lot of noise due to how some vendor products respond to OpenURLs. Restricting calls to only standard number items doesn't help in these cases. This PR  will only make a request for full-text when a standard number is present and the record format is either "Journal or Book." Some of the vendor target databases in SFX only support title level matching, no matter if a standard number is present or not.

Examples:
https://pulsearch.princeton.edu/catalog/9041463
https://pulsearch.princeton.edu/catalog/8044154

The same ids running this branch in dev:
https://pulsearch-dev.princeton.edu/catalog/9041463
https://pulsearch-dev.princeton.edu/catalog/8044154

This isn't a perfect solution. The urlify helper in particular is kind of out of control. If we can update the solr document to try and denote whether or not a given link drawn from a 856 value is a "full-text" option I think we could definitely simplify all the checking for what "linking" services we are trying to federate from umlaut we do in urlify and in in the show_availablility_sidebar partial.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -4,7 +4,7 @@ module ApplicationHelper
   # First argument of link_to is optional display text. If null, the second argument
   # (URL) is the display text for the link.
   # Proxy Base is added to force remote access when appropriate
-  def urlify(electronic_access)
+  def urlify(electronic_access, document)
     urls = ''
     links = JSON.parse(electronic_access)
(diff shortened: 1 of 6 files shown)


jpstroop:
Seems like a good idea.

PR #879, opened 2017-05-18
Update all references to catalog.princeton.edu to https.

kevinreiss:
To be deployed when the voyager opac and web services are changed over to https by default.


kevinreiss:
@tampakis could you merge this one when you have a chance.

PR #917, opened 2017-08-09
redirect voyager urls to blacklight catalog page

tampakis:
When Orangelight gets moved to catalog.princeton.edu, Voyager urls like `https://catalog.princeton.edu/cgi-bin/Pwebrecon.cgi?BBID=9642463`
will redirect to `https://catalog.princeton.edu/catalog/9642463`


kevinreiss:
Kind of funny to see ```Pwebrecon.cgi```. 👍

Issue #922, opened 2017-08-11
Support SCSB Availability


kevinreiss:
Use the ```scsb_id``` param passed to the availability controller in marc liberation for the "show" display and the ```barcodes``` param for the "search" view. Also make sure availability works in browse view.

kevinreiss:
Still need to do this in "browse" view.

kevinreiss:
This turns out to be more painful than I thought. The SCSB system provides to two availability lookups. One via a single BIB ID value that gives you all barcodes/item status for the record and one via barcode that lets you supply up to 50 barcodes at a time. The current browse model only gives you access to the BIB ID and a holding ID. Doing lookups for every individual SCSB bib ID in a given browse display seems prohibitive as does loading the solr document to fetch the first barcode value in the items key in the holdings hash. I put in a placeholder block that tells the user to see the full record for availability.

kevinreiss:
Do we want to leave this as is @tampakis in the browse view?

tampakis:
Will revisit as part of ILS migration

kevinreiss:
It appears we decided not have any availability on the browse screens post alma. Closing. Re-open if I am mistaken.

PR #923, opened 2017-08-11
Scsb holdings and availability.

kevinreiss:
@tampakis does this look heading in a direction that is workable. Need to finish specs and handling the ajax operations.


app/assets/javascripts/availability.js.coffee.erb
```diff
@@ -31,8 +32,18 @@ class AvailabilityUpdater
       $.getJSON(url, this.process_records)
     else if $("*[data-availability-record='true']").length > 0
       id = window.location.pathname.split('/')[2]
-      url = "#{@availability_url}?id=#{id}"
+      console.log(id)
```
tampakis: make sure to remove that when you finish up

kevinreiss:
@tampakis This is ready to go (finally). This is all new code to deal with SCSB records. All other availability and holding displays should be unchanged.

kevinreiss:
Will close #922.

app/helpers/application_helper.rb
```diff
@@ -152,13 +156,42 @@ def process_physical_holding(holding, bib_id, holding_id, is_journal, pub_date)
...
+
+  def recap_supervised_items?(holding)
+    if holding.key? 'items'
+      restricted_items = holding['items'].map do |item|
+        return true if item['use_statement'] == 'Supervised Use'
+      end
+      restricted_items.compact!
+      return false if restricted_items.empty?
+    end
+  end
```
tampakis: This method is sloppy. If there aren't any items in the holding it returns nil. I think it would be cleaner if it looked like this:
```ruby
  def recap_supervised_items?(holding)
    if holding.key? 'items'
      restricted_items = holding['items'].select { |item| item['use_statement'] == 'Supervised Use' }
      !restricted_items.empty?
    else
      false
    end
  end
```
The select method will populate an array for any items that match the boolean condition. If the array is empty there are no supervised items and the method returns false. Ideally this method should be tested more directly

jpstroop: Longer term, checking if all these different strings match (e.g. `item['use_statement'] == 'Supervised Use'`) is going to be difficult to maintain. Seems like a file of constants might be a good idea.

kevinreiss: I agree generally agree that hardcoding strings like this isn't a good practice. But at the present time this is literally the only constraint.

PR #929, opened 2017-08-28
Remove lewis banner

kevinreiss:
Also takes out hacks to display inline hints about inaccessible locations from the holdings display code in app helper.


kevinreiss:
Companion PR at https://github.com/pulibrary/marc_liberation/pull/257.

PR #934, opened 2017-09-06
display certificate program for theses

tampakis:


kevinreiss:
@tampakis, you're back!

PR #943, opened 2017-09-23
SCSB release fixes. Also Pin to solr 6.6.1 for now in dev/test."

### LINKED ISSUE #937: Use "Request in Reading Room" label only when all items in a mixed holding are restricted.
For mixed holdings in which some items are available in the Reading Room, but others are not, OL should just use the "Request" label rather than "Request in Reading Room" label. 

Example:
https://pulsearch-dev.princeton.edu/catalog/SCSB-6041751

kevinreiss:
Pin to solr 6.6.1 for now in dev/test

Fixes #937 for SCSB release.

Puts in placeholder view record statement. Advances #937.

Set default value for OL db username.

Basic local dev instructions.


app/helpers/application_helper.rb
```diff
@@ -185,7 +185,12 @@ def recap_item_list(holding)
   def recap_supervised_items?(holding)
     if holding.key? 'items'
       restricted_items = holding['items'].select { |item| item['use_statement'] == 'Supervised Use' }
-      !restricted_items.empty?
+      # !restricted_items.empty?
+      if restricted_items.count == holding['items'].count
+        true
+      else
+        false
+      end
```
tampakis: Lines 188-192 can be expressed with a single line:
```ruby
      restricted_items.count == holding['items'].count
```

kevinreiss: Thanks. Correct that.

PR #975, opened 2017-10-17
Update requests gem with scsb recall pickup fix.

kevinreiss:
Updates requests gem to include a bugfix for SCSB recall pickup lists.


kevinreiss:
@tampakis the modal-6-x branch we are pointing at has been merged into https://github.com/projectblacklight/blacklight/tree/release-6.x by @jcoyne. I will update OL to point at that branch for now since the commit reference we have in Gemfile.lock for your fix is now stale.

PR #981, opened 2017-10-24
Shorten account error message per circulation staff. 

kevinreiss:
Also advance requests gem to latest master for bugfix related to SCSB pickup location and on shelf no item holdings.


kevinreiss:
@tampakis This can be merged and deployed when you have a chance. Fixes a number of bugs related to pickup locations for SCSB "charged" materials and updates an account error message requested by circulation.

PR #982, opened 2017-10-26
Advanced search zindex

### LINKED ISSUE #971: Increase contrast when hovering over bookmarks and course reserves buttons
Currently when hovering over the bookmarks/course reserves buttons in the search bar, the link text turns to white, which is difficult to see over the light gray background.

tampakis:
Advanced search link now works on smaller screens.

Bookmarks/Course Reserves look like and behave more links on hover. Closes #971.


tampakis:
@axamei, I deployed to production as a quick fix so that users on smaller screens can click the advanced search link. Not sure if there was a cleaner solution.

PR #985, opened 2017-11-09
Don't inject proxy prefix for open access titles. Fixes #967

### LINKED ISSUE #967: Exempt Open Access Links from Proxy Prefix
When an online record has the label "Open Access" derived from the subfield x of the 856 field OL should not provide the proxy prefix before the URL in online links. Deals with the longstanding issue with Open Access handles being sent to the proxy for authentication discussed at the last LISSC+ meeting.

kevinreiss:


kevinreiss:
@tampakis when green this should be good.

app/helpers/application_helper.rb
```diff
@@ -28,6 +32,18 @@ def urlify(electronic_access, document)
     end
   end
 
+  def umlaut_fulltext_eligible?(document)
```
tampakis: Rather than copy the same method twice, could you include `Blacklight::Solr::Document::Marc` in the helper? Then you could change line 21/25 back to what it was before.

Issue #1007, opened 2017-12-11
[Orangelight/production] ActionView::Template::Error: undefined method `[]' for nil:NilClass


eliotjordan:
## Backtrace

    line 158 of [PROJECT_ROOT]/app/helpers/blacklight_helper.rb: linked_records
    line 1 of [PROJECT_ROOT]/app/views/catalog/_other_versions_show.html.erb: _app_views_catalog__other_versions_show_html_erb__1357375038105022882_70178584774940
    line 10 of [PROJECT_ROOT]/app/views/catalog/_show_default.html.erb: block in _app_views_catalog__show_default_html_erb___4405944016186085182_70178570722160

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/54399/faults/35754237)

jrgriffiniii:
I believe that this is resolved in #1038:

https://github.com/pulibrary/orangelight/pull/1038/files#diff-f9d7529488b85e40393a73ab6880971aR19

PR #1038, opened 2017-12-13
Implements the LinkedDocumentResolver in order to provide retrieve linked record documents and renders metadata fields

### LINKED ISSUE #990: Include more information for other version titles
Perhaps more info about the other matched versions can be displayed to reduce confusion when there are multiple other versions matches: https://pulsearch.princeton.edu/catalog/3287143

Also displaying the cataloging note in 776 subfield $i would help. See #595.

jrgriffiniii:
Resolves #990 , #1007 , and #1124

### DIFF
--- a/app/helpers/blacklight_helper.rb
+++ b/app/helpers/blacklight_helper.rb
@@ -166,35 +166,17 @@ module BlacklightHelper
     link_to(child.html_safe, path, id: 'citeLink', data: { ajax_modal: 'trigger' }, class: 'btn btn-default')
   end
 
-  # solr fq field is field parameter provided unless id_nums value starts with BIB
-  def linked_records(id_nums, bib_id, field)
-    fq = ''
-    id_nums.each do |n|
-      bib_match = /(?:^BIB)(.*)/.match(n)
-      solr_field = bib_match ? 'id' : field
-      n = bib_match[1] if bib_match
-      fq += "#{solr_field}:#{n} OR "
-    end
-    fq.chomp!(' OR ')
-    resp = get_fq_solr_response(fq)
-    req = JSON.parse(resp.body)
-    other_versions = []
-    if req.dig('response', 'docs')
-      req['response']['docs'].each do |record|
-        unless record['id'] == bib_id
-          title = record['title_display']
-          other_versions << link_to(title, solr_document_url(record['id']))
-        end
-      end
-    end
-    other_versions.empty? ? [] : [other_versions]
+  # Retrieve an instance of the FacetedQueryService
+  # @return [FacetedQueryService] an instance of the service object
+  def faceted_query_service
+    @faceted_query_service ||= FacetedQueryService.new(Blacklight)
   end
 
   def oclc_resolve(oclc)
     oclc_norm = StringFunctions.oclc_normalize(oclc)
     unless oclc_norm.nil?
       fq = "oclc_s:#{oclc_norm}"
-      resp = get_fq_solr_response(fq)
+      resp = faceted_query_service.get_fq_solr_response(fq)
       req = JSON.parse(resp.body)
     end
     if oclc_norm == 0 || req['response']['docs'].empty?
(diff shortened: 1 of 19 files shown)


jrgriffiniii:
Please see the following screenshot for reference:
issues_990_screenshot_0

app/helpers/blacklight_helper.rb
```diff
@@ -144,25 +144,18 @@ def render_icon(var)
...
-    req = JSON.parse(resp.body)
-    other_versions = []
-    req['response']['docs'].each do |record|
-      unless record['id'] == bib_id
-        title = record['title_display']
-        other_versions << link_to(title, solr_document_url(record['id']))
-      end
-    end
-    other_versions.empty? ? [] : [other_versions]
+    neighborhood = ::NeighborhoodService::DocumentNeighborhood.new(neighbors: id_nums,
```
tpendragon: I'm not sure if `neighborhood` and `neighbors` are clear names here. Can we pick a name which is similar to where it's being used? Maybe `version_ids` and `VersionFetcher` or something?

tpendragon: Or maybe even something relating to the fact that it's a `linked_record`

app/views/catalog/_show_availability_sidebar.html.erb
```diff
@@ -19,15 +19,7 @@
...
-      <div class="panel-body">
-        <% linked_records.each do |records| %>
-          <ul>
-            <% records.each do |link| %>
-              <li><%= link %></li>
-            <% end %>
-          </ul>
-        <% end %>
-      </div>
+      <%= linked_records %>
```
tpendragon: Is a helper appropriate for this? It seems rather than a global function, having something like what was previously here but doing `<% document.linked_records.each` would be clearer. Just a personal opinion though.

tpendragon: It would be even clearer if you could just do:

```ruby
render partial: 'show_linked_records', locals: { linked_records: document.linked_records }
```

app/helpers/blacklight_helper.rb
jrgriffiniii: This has been restructured as a `LinkedDocumentResolver`

app/views/catalog/_show_availability_sidebar.html.erb
jrgriffiniii: This has been restructured as https://github.com/pulibrary/orangelight/pull/1038/files#diff-cce3cb5f9d0085ad25e65b70ff5dcbc9R22

jrgriffiniii:
This may still be considered as "in progress", as it must be tested against the production Solr instance in order to identify any outlying usability issues.

app/services/linked_document_resolver.rb
```diff
@@ -0,0 +1,98 @@
...
+        queries = @siblings.map do |sibling_number|
+          "#{solr_field(sibling_number)}:#{solr_value(sibling_number)}"
+        end
+        queries.join(' OR ')
+      end
+
+      # Retrieve a response from the Solr endpoint for a faceted query
+      # @param fq [String] the Solr facet query
+      # @return [Faraday::Response] the HTTP response to the query
+      def get_fq_solr_response(fq)
```
jrgriffiniii: This should be deduplicated with https://github.com/pulibrary/orangelight/pull/1038/files#diff-65c43d8c8cb6c83aaf3fe8d18991633aR221

jrgriffiniii:
@kevinreiss regarding the updates for restrictions and availability, please see the following screenshots:

orangelight_990_screenshot_1
orangelight_990_screenshot_2

There are still some improvements I can make here, but I want to verify that this was what you requested.

kevinreiss:
The screenshots look pretty promising @jrgriffiniii. If we went with the pattern of a separate well for use restrictions we should probably also find a way to do this with Princeton's RBSC items which also would technically fall under the "supervised use" restriction though we don't current mark them as such. Rendering the availability status as part of the "other versions" would be effective.

PR #1050, opened 2017-12-15
Integrates the `sitemap_generator` Gem and implements the `RobotsGeneratorService` to generate a robots.txt

### LINKED ISSUE #1000: Create sitemap
- [ ] Create sitemap listing all item show pages
 - [ ] Block everything but the sitemap and item show pages in robots.txt

jrgriffiniii:
Resolves #1000


config/deploy/production.rb
```diff
@@ -27,6 +27,8 @@
 server 'lib-orange-prod6', user: 'deploy', roles: %i[web app db]
 
 set :deploy_to, '/opt/rails_app'
+after 'deploy:published', 'sitemap:create'
+after 'deploy:published', 'pulsearch:robots_txt'
```
eliotjordan: These tasks are not yet defined in `deploy.rb`. Discovered this after deploying to staging and noticing that the sitemap wasn't generated.

lib/tasks/pulsearch.rake
```diff
@@ -31,6 +32,17 @@ namespace :pulsearch do
...
+  desc 'Generate a robots.txt file'
+  task :robots_txt do |_t, args|
+    file_path = args[:file_path] || Rails.root.join('public', 'robots.txt')
+    robots = RobotsGeneratorService.new(path: file_path, disallowed_paths: Rails.configuration.robots.disallowed_paths)
+    robots.insert_group(user_agent: '*')
+    robots.insert_crawl_delay(10)
+    robots.insert_sitemap(Rails.configuration.robots.sitemap_url)
+    robots.generate
+    robots.write
+  end
```
escowles: I'm curious about how you see this working going forward.  Do you think we should periodically regenerate our `robots.txt` and compare to the checked-in static version? Or is this code that was useful to generate the original static version, that we don't need to maintain?

config/sitemap.rb
```diff
@@ -0,0 +1,39 @@
+require 'rubygems'
+require 'sitemap_generator'
+
+SitemapGenerator::Sitemap.default_host = 'https://pulsearch.princeton.edu'
```
escowles: It would be better to get this from the Rails config instead of hard-coding `pulsearch.princeton.edu` — since that's not the staging URL, and won't be the production URL for very much longer.

lib/tasks/pulsearch.rake
jrgriffiniii: > is this code that was useful to generate the original static version, that we don't need to maintain?
I felt that this was the preferred approach, with the hope that any adjustment to the Routes for the App would require the config. being adjusted (please see https://github.com/pulibrary/orangelight/pull/1050/files#diff-b29c2e1a89929db60557b9de48f7b633).  At the time that this PR was issued, I could not determine a straightforward approach to accessing all Rails routes and finding some more automated way to blacklist certain patterns, so unfortunately, the robots.txt rules will require some curation.

config/sitemap.rb
jrgriffiniii: I have addressed this, but this PR is now blocked until princeton_ansible/issues/114 is resolved.

jrgriffiniii:
Currently https://github.com/pulibrary/princeton_ansible/pull/116 ensures that the Sitemap is generated nightly by a cron task, as well as to ensure that the host name is passed using the Rails env. variables

PR #1055, opened 2017-12-15
Exclude certain export formats for non-voyager documents

### LINKED ISSUE #1011: [Orangelight/production] NoMethodError: undefined method `[]' for nil:NilClass
## Backtrace

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/54399/faults/35755613)

`export_as_endnote`

### LINKED ISSUE #1012: [Orangelight/production] NoMethodError: undefined method `find_all' for nil:NilClass
`export_as_refworks_marc_txt`

## Backtrace

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/54399/faults/35754707)

### LINKED ISSUE #1015: [Orangelight/production] NoMethodError: undefined method `to_marc' for nil:NilClass
## Backtrace

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/54399/faults/35760792)

### LINKED ISSUE #1053: [Orangelight/production] NoMethodError: undefined method `to_xml' for nil:NilClass
## Backtrace

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/54399/faults/35768499)

eliotjordan:
- Allows rails to return the proper response when a user (or bot, as is the case most of the time) attempts to get a non-voyager document in an unsupported format. 
- This PR also removes the need to pass a list of excluded formats when generating alternate links.

Closes #1011 (endnote)
Closes #1012 (refworks_marc_txt)
Closes #1015 (marc)
Closes #1053 (marcxml)


app/models/concerns/orangelight/document/export.rb
```diff
@@ -0,0 +1,38 @@
...
+        def holding_id
+          @holding_id ||= begin
+            holdings = JSON.parse(fetch(:holdings_1display, '{}')).first
+            holdings.blank? ? nil : holdings[0]
+          end
+        end
+
+        def non_voyager?
+          return true if %w[thesis visuals].include? holding_id
+          return true if fetch(:id, '').start_with?('SCSB')
```
escowles: One small thing: can you do the SCSB check first (because there are tons more of those), and reverse the logic of this method?  I always find negative boolean methods confusing, so reversing the logic would let line 20 above be `return [] if voyager?` which seems clearer to me.

kevinreiss: Do we want to support marcxml exports for SCSB? Some of the citation scrapers like Zotero work better with Marc than what we are doing now. It would be easy enough to do if we were willing to create a field in the solr document that contained a blob of MARCXML since we do get MarcXML as input for those records.

escowles: I think it's worth considering marcxml support for the SCSB records, though I don't want to hold up this ticket for that decision (which probably involves getting input from LISSC, etc.).  So I've created a ticket for that issue: https://github.com/pulibrary/orangelight/issues/1065

eliotjordan: Yeah, negative boolean method names are confusing. I waffled back and forth on that a bit. I'll change to the affirmative method name.

PR #1066, opened 2017-12-18
Prevents suspected bots from viewing availability information for catalog records

### LINKED ISSUE #1001: Don't display realtime availability info for robots
There should be a useragent-sniffing gem to make this easier.

jrgriffiniii:
Resolves #1001

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -81,6 +81,8 @@ gem 'openurl', '~> 1.0'
 
 gem 'honeybadger', '~> 3.1'
 
+gem 'voight_kampff', '~> 1.1'
+
 group :development do
   gem 'capistrano-rails', '~> 1.1.6'
 end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -441,6 +441,8 @@ GEM
       unf_ext
     unf_ext (0.0.7.4)
     unicode-display_width (1.3.0)
+    voight_kampff (1.1.1)
+      rack (>= 1.4, < 3.0)
     warden (1.2.7)
       rack (>= 1.0)
     webmock (3.1.1)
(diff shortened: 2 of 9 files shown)


escowles:
The [Voight-Kampff README](https://github.com/biola/Voight-Kampff#configuration) mentions a Rake task that will download an updated list of bob user agent strings and save it in `config`.  Do we want to do that and checkin that file? Or run the rake task at deploy time?

tampakis:
This looks great @jrgriffiniii! My only concern is with the scope of which content to exclude from bots. The show availability sidebar partial includes the item location information, which might be useful to keep in for bots. Also it would be great to hide the availability sections for the index and browse views. I'll come up with the places in the code where I think make the most since to exclude content.

tampakis:
For bots, don't display:
 - Results page availability - this method in the application is nasty, but I think there are only two spots that need to be updated:
   - [set check_availability to false](https://github.com/pulibrary/orangelight/blob/master/app/helpers/application_helper.rb#L329)
   - [set check_availability to false again when link is missing](https://github.com/pulibrary/orangelight/blob/master/app/helpers/application_helper.rb#L334)
 - [Call number browse availability](https://github.com/pulibrary/orangelight/blob/master/app/views/orangelight/browsables/index.html.erb#L84-L92)

Actually, I think it's fine excluding the entire show_availability sidebar.

tampakis:
@eliotjordan brought up a good point that it would be wise to also block bots from certain routes in OL that make connections to bibdata, at the controller level. This would include the staff view and the export formats.

escowles:
I think blocking bots is a good idea — but do we want to do that in the load-balancer instead of in the app?

tampakis:
When a bot requests a .endnote page would the load balancer know when orangelight makes the call to bibdata?

escowles:
@tampakis No, I mean the load balancer would block the request if a bot requested a .endnote page, so it would never get to OL.

Issue #914, opened 2017-08-09
Upgrade Ruby to Version 2.4.2


tampakis:
We're still on version 2.2.3. Technically Blacklight isn't supported on that version.

tampakis:
Ansiblize!

jrgriffiniii:
Within Ubuntu 16.04 LTS environments, only Ruby release 2.3.1 is available as a dkpg:
```
[...]
Welcome to Ubuntu 16.04.3 LTS [...]
[...]
user@some-host:~$ ruby --version
ruby 2.3.1p112 [...]
```

Hence, the scope of this update has been downgraded.

jrgriffiniii:
Proposed solutions related to infrastructure and modifications to Ansible Roles have been outlined in https://github.com/pulibrary/princeton_ansible/issues/134

PR #1074, opened 2017-12-19
Add .ruby-version for version managers such as chruby, fixes #1027

### LINKED ISSUE #1027: Add ruby version to Gemfile, .ruby-version
Questions:
* I've added a `.ruby-version` to be used by chruby. Not sure whether this would be compatible with whatever ruby version managers y'all are using. Do you prefer me to check in this file or add it to `.gitignore`?

* How do y'all feel about adding the ruby version to the Gemfile? This will make it impossible to run under a different version.

hackmastera:


escowles:
We're using 2.3.1 for Travis (https://github.com/pulibrary/orangelight/blob/master/.travis.yml#L4) — shouldn't we be using the same here?

hackmastera:
hm, I was basing it on https://github.com/pulibrary/orangelight/issues/914 which I think reflects the actual production boxes, except for one which is running a later version (don't remember which exactly).

If our prod boxes are 2.2.3 maybe we should change travis.

escowles:
Yes, switching Travis to 2.2.3 seems reasonable to me, until we can upgrade the production boxes to 2.4.x.

hackmastera:
okay, pushed that up so I guess we'll see if travis likes it!

PR #1078, opened 2017-12-19
Use bibdata-dev in development

hackmastera:


config/requests.yml
```diff
@@ -13,6 +13,8 @@ defaults: &defaults
 development:
   <<: *defaults
   voyager_ub_id: 1@DB
+  bibdata_base: https://bibdata-dev.princeton.edu
+  stackmap_base: https://bibdata-dev.princeton.edu/locations/map
```
kevinreiss: @hackmastera I don't think we need the stackmap_base key set any more. That one was purged when we took that code out of bibdata last week.

hackmastera: oh, great. thanks will take care of it.

hackmastera:
@kevinreiss made that change; this is ready

Issue #1082, opened 2017-12-19
Stub server-side calls to external APIs with webmock


hackmastera:
breaking this out from #656

PR #1083, opened 2017-12-19
Stub bibdata holding_locations with webmock

hackmastera:
refs #1082


hackmastera:
So I didn't do this because I thought we'd be using puffing_billy, but since we're not another option on this PR would be to put a global `before` to stub that bibdata `holding_locations` in every test. A little over 1/4 of tests use it. I'm not sure how much overhead this would add.

PR #1098, opened 2017-12-27
add null alt attribute to book covers

### LINKED ISSUE #1061: Book covers must have alt tags
Level A violation: Images must have alternate text.

Book covers in this context should be treated as decorative elements so a null alt tag is sufficient. Add `alt=""` to [book covers](https://github.com/pulibrary/orangelight/blob/d8bdd5d626fd7523f75580dc0248e6352625cd64/app/assets/javascripts/book_covers.js.coffee#L36) `img` tag.

tampakis:
Adds a null alt attribute to book covers to indicate decorative function. Closes #1061.


axamei:
👍

PR #1100, opened 2017-12-27
Refactors the generation of markup for online and physical item holdings

### LINKED ISSUE #1048: Streamline links to digital content on the same page
Currently, we display links in the online content box that go through a couple of lookups and eventually redirect to the same page (when the viewer is at the bottom of the page).

We should be able to figure that out and just have a relative link to "#view" and make the link text something like "digital content below"

jrgriffiniii:
Resolves #1048 (partially)

### DIFF
--- a/app/services/holding_requests_adapter.rb
+++ b/app/services/holding_requests_adapter.rb
@@ -1,0 +1,172 @@
+# Adapter for SolrDocument instances and the Bibdata Class
+class HoldingRequestsAdapter
+  attr_reader :document
+
+  # Construct the interface for the Solr Document and Bib. Data API
+  # @param document [SolrDocument]
+  # @param bib_data_service [Bibdata]
+  def initialize(document, bib_data_service)
+    @document = document
+    @bib_data_service = bib_data_service
+  end
+
+  # Retrieve the ID for the Solr Document
+  # @return [String] the ID value
+  def doc_id
+    @document.fetch('id')
+  end
+
+  # Access the holding locations from Bib. Data
+  # @return [Hash] location hash structure
+  delegate :holding_locations, to: :@bib_data_service
+
+  # Retrieve the holdings information from the Solr Document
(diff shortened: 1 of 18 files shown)


app/services/holding_requests_adapter.rb
```diff
@@ -0,0 +1,172 @@
...
+    holding['call_number_browse'] || holding['call_number']
+  end
+
+  # Determine whether or not the holding is for a repository item
+  # @return [TrueClass, FalseClass]
+  def repository_holding?(holding)
+    holding['dspace']
+  end
+
+  # Determine whether or not the holding is for a ReCAP item
```
kevinreiss: Technically this would only test for "scsb" partner recap holdings. PUL has a lot of "recap" items we own at location code that don't start with "scsb".  We might want to use a different name for this method.

jrgriffiniii: @kevinreiss Thank you correcting this, I have updated the method names for `HoldingRequestsAdapter#recap_holding?` and `PhysicalHoldingsMarkupBuilder.recap_location?` in order to have this resolved.

However there appear to be cases where this is in place outside of this PR (please see https://github.com/pulibrary/orangelight/blob/master/app/helpers/orangelight/browsables_helper.rb#L3).  I can address these within the scope of this PR as well, or open another issue for this.

kevinreiss: @jrgriffiniii Looking at this a little more I can see that all the scsb related code that was added to the helpers uses a prefix of ```recap_``` for the method names when I really should have used something more specific like```scsb_```.  These all seem to only provide behaviors specific to the SCSB partner records. https://github.com/pulibrary/orangelight/pull/1100/files?diff=unified#diff-a213c754c47e851ef2a146f5397d902eR192 is another example of this. For the sake of clarity do you think it is worth going through all of them and making this change?

PR #1111, opened 2018-01-08
Updates Ruby to 2.4.2

### LINKED ISSUE #914: Upgrade Ruby to Version 2.4.2

jrgriffiniii:
Resolves #914 by updating Devise, other Gem dependencies, and addressing Rubocop violations for interpretation using Ruby release 2.4.3.  Unfortunately, attempts to update Ruby to 2.5.0 are currently blocked by [a Travis CI bug](https://github.com/pulibrary/orangelight/issues/1112)


.ruby-version
```diff
@@ -1 +1 @@
-ruby-2.2.3
+ruby-2.5.0
```
escowles: The README and `.ruby-version` say 2.5.0, but the PR title and `.travis.yml` say 2.4.3.  Which version are we targeting?

app/helpers/application_helper.rb
```diff
@@ -293,7 +295,7 @@ def browse_name(args)
 
   def name_title(args)
     args[:document][args[:field]].each_with_index do |name_t, i|
-      next unless args[:document]['name_title_browse_s'] && args[:document]['name_title_browse_s'].include?(name_t)
+      next unless args[:document]['name_title_browse_s']&.include?(name_t)
```
escowles: I hadn't seen that syntax before, but it seems like a nice way to make those checks less verbose.

PR #1122, opened 2018-01-09
Default to https on staging

hackmastera:
Do we want this?


escowles:
We default to https for https://pulsearch-dev.princeton.edu/ and other apps, so that makes sense for catalog-staging.

Issue #1119, opened 2018-01-09
SCSB "Use Restrictions" Label Appears when there are no use restrictions on items


kevinreiss:
See 

https://pulsearch.princeton.edu/catalog/SCSB-4025167

PR #1125, opened 2018-01-10
Suppress Use Restrictions Label when there aren't any.

kevinreiss:
For #1119.


kevinreiss:
Sorry. I updated the existing feature to spec to check for this condition.

PR #1204, opened 2018-03-06
Removes Open Graph thumbnail URLs

### LINKED ISSUE #1196: Use thumbnail for og:image when we have one
Gives a nicer preview than the PUL logo. Would at least be nice to have for stuff in figgy, and that's most likely to be tweeted about, shared on FB, etc.

screen shot 2018-02-21 at 2 07 40 pm

jrgriffiniii:
Resolves #1196 by updating the `<meta>` elements within the `<head>` in order to expose thumbnails as Open Graph image metadata (please see http://ogp.me/#metadata)

### DIFF
--- a/app/assets/javascripts/book_covers.js.coffee
+++ b/app/assets/javascripts/book_covers.js.coffee
@@ -23,7 +23,23 @@ class BookCoverManager
     true
   fetch_identifiers: (identifiers) ->
     url = "#{@google_url}#{identifiers.join(",")}"
-    $.getJSON(url).done(this.process_results)
+    $.getJSON(url).done(this.process_results, this.update_open_graph_metadata)
+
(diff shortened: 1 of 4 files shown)


tampakis:
I didn't know how to best test this, so I deployed it to production (!!) and although the og:image property gets added to the page when I share catalog links in Slack the orange PUL logo still shows up in the preview.

jrgriffiniii:
@tampakis likely the `<meta property="og:image">` tag is modified *after* the page is scraped by Facebook or Twitter.  I'll need to retrieve the Google Books thumbnail on the server (rather than using JavaScript).

jrgriffiniii:
Following discussions and proposal by @escowles and @tampakis , this now removes the default Open Graph thumbnail and CoffeeScript functionality for extracting these from the Google Books API and IIIF Manifest.

PR #1212, opened 2018-03-12
fixes course reserve filter

### LINKED ISSUE #1186: Course reserves can't be combined with recently added facet
Not sure why the error happens but the course reserve filters aren't compatible with the recently added facet.

Examples (error happens after applying recently added facet):
https://catalog.princeton.edu/catalog?f%5Bcourse%5D%5B%5D=ANT+454%3A+Transcultural+Cinema&f%5Binstructor%5D%5B%5D=Himpele%2C+J.&search_field=all_fields
https://catalog.princeton.edu/?f%5Bcourse%5D%5B%5D=CHM+504%3A+Molecular+Spectroscopy&f%5Binstructor%5D%5B%5D=Wang+et+al%2C+

Also see https://app.honeybadger.io/projects/54399/faults/36610056

christinach:
fixes #1186

### DIFF
--- a/spec/requests/request_spec.rb
+++ b/spec/requests/request_spec.rb
@@ -88,8 +88,6 @@ describe 'blacklight tests' do
 
   describe 'pul_holdings check' do
     it 'excludes scsb items when pul location filter is applied' do
-      get '/catalog.json?per_page=100'
-      expect(response.body).to include('"SCSB-7235709"', '"9741216"')
       get '/catalog.json?per_page=100&f_inclusive%5Badvanced_location_s%5D%5B%5D=pul'
       expect(response.body).to include('"9741216"')
       expect(response.body).not_to include('"SCSB-7235709"')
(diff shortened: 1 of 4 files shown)


spec/requests/request_spec.rb
```diff
@@ -88,8 +88,8 @@
 
   describe 'pul_holdings check' do
     it 'excludes scsb items when pul location filter is applied' do
-      get '/catalog.json?per_page=100'
-      expect(response.body).to include('"SCSB-7235709"', '"9741216"')
+      # get '/catalog.json?per_page=150'
+      # expect(response.body).to include('"SCSB-7235709"', '"9741216"')
```
escowles: Is this test change related?  It would be good to know why it should be removed, or keep it if there's no particular reason to remove it.

tampakis: I suggested to have it removed because now there are over 100 fixtures in the development environment. 100 is the max one can fetch at once in Blacklight. The test was failing because id SCSB-7235709 was on the 2nd page of the results.

escowles: 👍 that makes sense.  Can we delete it instead of commenting it out?

Issue #1070, opened 2017-12-19
Add labels to form elements in Advanced Search


axamei:
Violation: Form elements must have labels.

Add `<label>` or aria-label or aria-labelledby

tpendragon:
Closed by #1232

PR #1232, opened 2018-03-30
adds a label tag with sr-only class in the advanced-search form

christinach:


app/views/advanced/_guided_search_fields.html.erb
```diff
@@ -1,6 +1,7 @@
 
 <div class="advanced-search-field">
   <div class="search-field">
+    <label class="sr-only">Options for advanced search</label>
```
axamei: @christinach When the label element is detached from the input element, the `for` attribute should be used (the value should be the `id` of the form element that is being labeled. In this case, `<label for ="f1" class="sr-only">Options for advanced search</label>` since the `select` element's `id="f1"`

app/views/advanced/_guided_search_fields.html.erb
```diff
@@ -1,6 +1,7 @@
...
     <%= select_tag('f1', options_for_select(advanced_key_value, guided_field(:f1, 'all_fields')), :class=>"search_field", 'aria-label': 'Options for advanced search') %>
```
axamei: We can remove the `aria-label` attribute and value since we are replacing it with the label element.

app/views/advanced/_guided_search_fields.html.erb
christinach: Thank you @axamei; will update the file

app/views/advanced/_guided_search_fields.html.erb
christinach: will do, thanks!

app/views/advanced/_guided_search_fields.html.erb
```diff
@@ -1,7 +1,8 @@
 
 <div class="advanced-search-field">
   <div class="search-field">
-    <%= select_tag('f1', options_for_select(advanced_key_value, guided_field(:f1, 'all_fields')), :class=>"search_field", 'aria-label': 'Options for advanced search') %>
+    <label for="f1" class="sr-only">Options for advanced search</label>
+    <%= select_tag('f1', options_for_select(advanced_key_value, guided_field(:f1, 'all_fields')), :class=>"search_field") %>
     <div name="row1">
       <%= text_field_tag "q1", label_tag_default_for(:q1), :class => 'form-control', autocorrect: "off", autocapitalize: "off", spellcheck: "false", 'aria-label': 'Advanced search terms' %>
```
axamei: @christinach Can you also add a label element for these input fields, i.e., `q1`, `q2`, `q3`? Thank you.

christinach: I will @axamei. Thank you!

christinach:
closes #1070

PR #1241, opened 2018-04-17
If author_display is not present use author_s

### LINKED ISSUE #1228: Use first 7xx field for contributor field info in browse lists when there is no 1xx main entry
For records with no 1xx field and at least one 7xx field, use the first 7xx field in the record as a basis for a contributor for display in the browse lists. Also, change the label of the 'creator' field in the browse list to 'contributor.' [per LISSC 3/26/18]

christinach:
closes #1228


lib/orangelight/browse_lists.rb
```diff
@@ -91,7 +91,11 @@ def browse_cn(_sql_command, facet_request, conn, facet_field, _table_name)
                 date = record['pub_created_display'][0]
               end
               label = cn
-              author = record['author_display'][0..1].last if record['author_display']
+              if record['author_display']
+                author = record['author_display'][0]
```
tampakis: Can you keep the line as `record['author_display'][0..1].last`? This will display the vernacular script over the Latin script when it is present in the record.

christinach: yes I will. thanks.

Issue #960, opened 2017-10-04
Support UV display for records with multiple IIIF manifests


tampakis:
This record has two image manifests associated with it: https://pulsearch.princeton.edu/catalog/3943643

The UV only pulls in one of the manifests.

tampakis:
http://arks.princeton.edu/ark:/88435/kw52jc00n
https://figgy.princeton.edu/catalog/b975163a-db0f-4c61-a2bc-9c0fff0a0cce#?c=0&m=0&s=0&cv=0&xywh=-9072%2C-401%2C23208%2C8000


http://arks.princeton.edu/ark:/88435/np193c809
https://figgy.princeton.edu/catalog/654b1426-9c07-461d-8e43-7145499efc3a#?c=0&m=0&s=0&cv=0&xywh=-9156%2C-401%2C23208%2C8000

PR #1243, opened 2018-04-21
Embeds multiple figgy links in show page

christinach:


christinach:
closes #960

christinach:
closes #960

escowles:
@christinach When I look at 4609321 using this branch, I see only one of the two objects associated with that item: https://catalog-staging.princeton.edu/catalog/4609321  — are there indexing changes needed for this to work?

tampakis:
@escowles, that record is a special case, where I think even though there are 2 arks, they resolve to the same object in figgy.

This one's a better example: https://catalog-staging.princeton.edu/catalog/3943643

PR #1262, opened 2018-06-11
Update overdue label to remove 'recalled'. Fixes #1258.

### LINKED ISSUE #1258: Change Item Status Label
The current "Overdue/Recalled" label at https://github.com/pulibrary/orangelight/blob/39928d057cbc6b3ba60c0d4a4c4144b520cd1805/config/voyager_item_status.yml#L16 is misleading. It should just read "Overdue". Recalled items have a different item status ID entirely. @tampakis I can't remember how the current label was arrived at. It may have just been an error on my part transcribing things from our Voyager item status label compilation.

kevinreiss:


.solr_wrapper
```diff
@@ -1 +1 @@
-version: 7.1.0
```
tampakis: Can you revert this change?

PR #1264, opened 2018-06-12
upgrades OL to use UV3

sdellis:
There are some css classes in `universal-viewer.scss`. I imagine they are safe to remove, but wasn't sure.


tampakis:
Looks good, thanks!

Issue #1263, opened 2018-06-12
Switch to Matomo


escowles:
Coordinate with @kayiwa to use anonymized Matomo instead of Google Analytics.

PR #1268, opened 2018-06-14
Adds matomo

christinach:


app/views/layouts/application.html.erb
```diff
@@ -35,8 +35,9 @@
     <!--[if lt IE 9]>
       <script src="//html5shim.googlecode.com/svn/trunk/html5.js"></script>
     <![endif]-->
-
+
```
tampakis: Get rid of this line

christinach:
I removed the package-lock.json. It was created when I tested the npm matomo package. 
I'm in favor of adding Webpack to control the js and css files. It would be much easier to noParse files that we don't need in specific environments.

christinach:
Right now for testing purposes will track the traffic both in catalog-staging and catalog.

christinach:
colses #1263

PR #1271, opened 2018-06-15
Define chosen-sprite before loading the chosen-rails gem

christinach:


christinach:
With the a commit to update the chosen-rails gem #1266 the user can close the selected value but the chosen-sprite (x) image still do not display. This is an update in #1266. It is a temporary issue until the gem gets fixed.

PR #1297, opened 2018-07-24
Update viewer to be compatible with IE11

christinach:


christinach:
I updated the viewer and tested it using a **windows 7 VM with IE11**:

1. I deployed the PR to **catalog-staging** and the viewer did not display because of a cross origin error.

2. I deployed the PR to **catalog production** and the viewer displayed properly.

PR #1317, opened 2018-08-10
update requests dev and test environment to  bibdata-staging

christinach:
I'm not sure yet why these tests are failing when I run rake ci locally
```
rspec ./spec/features/request_options_spec.rb:37 # Request Options In process status non-requestable location does display a request button
rspec ./spec/features/request_options_spec.rb:49 # Request Options Available status requestable location does display a request button
rspec ./spec/features/request_options_spec.rb:61 # Request Options Multi-item holding with some requestable items does display a request button
```

### DIFF
--- a/config/requests.yml
+++ b/config/requests.yml
@@ -13,11 +13,11 @@ defaults: &defaults
 development:
   <<: *defaults
   voyager_ub_id: 1@DB
-  bibdata_base: https://bibdata-dev.princeton.edu
+  bibdata_base: https://bibdata-staging.princeton.edu
 test:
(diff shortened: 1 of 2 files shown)


kevinreiss:
@christinach these specs are pretty creaky. They are executing live calls against the bibdata availability by holding ID lookup in the bibdata environment the app is pointing at. I'm not sure why the first two fail (they might just be timing out), the third one (bib ID value: 6045464) seems to be sending be sending a bad id value over to bibdata.

christinach:
Thanks @kevinreiss

Issue #1315, opened 2018-08-10
Headers/footers should match main Library website


escowles:

kevinreiss:
Header changed to match the main website for now.

PR #1321, opened 2018-08-13
Updates header to have black background

christinach:


christinach:
closes #1315

christinach:
@axamei the black color from the library website is #040404 ; and this is the one I added in the file. Are we ok with that or do we want black #000000 ?

PR #1348, opened 2018-08-30
only request facets on html pages

### LINKED ISSUE #1346: Limit unnecessary facet requests
Facet queries are calculated in the jvm and are expensive when there are lots of unique facet values. Limit the number of times these facets get requested.

tampakis:
Closes #1346

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -262,4 +262,4 @@ Rails/UnknownEnv:
     - staging
     - production
     - development
-    - test    
+    - test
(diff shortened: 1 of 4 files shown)


.rubocop.yml
```diff
@@ -168,6 +168,10 @@ Style/BlockDelimiters:
   Exclude:
     - 'spec/helpers/scsb_items_helper_spec.rb'
 
+Style/GuardClause:
+  Exclude:
+    - 'app/models/search_builder.rb'
+
```
tampakis: I added this because I wanted an explicit check for whether the format param was nil, which indicates that an html page is requested

PR #1354, opened 2018-09-06
Revert "Updating rubyzip to release 1.2.2"

jrgriffiniii:
Reverts pulibrary/orangelight#1353


escowles:
I'm 👍 on #1353, so we can just leave that merged.

PR #1357, opened 2018-09-07
Revert "Overrides `Blacklight::SearchHelper#search_results` in order to integrate caching support for default searches"

escowles:
Reverts pulibrary/orangelight#1356


escowles:
I think we can just update master before deploying, so closing this revert PR.

PR #1374, opened 2018-09-17
Disables the server-side response cache (avoiding cache corruption errors on the server); Adds failure handlers for bibdata AJAX requests

### LINKED ISSUE #1373: AJAX requests to https://bibdata.princeton.edu/availability? which fail aren't handled
Please see https://github.com/pulibrary/orangelight/blob/master/app/assets/javascripts/availability.js.coffee.erb#L32

jrgriffiniii:
Resolves #1373

### DIFF
--- a/app/helpers/blacklight_helper.rb
+++ b/app/helpers/blacklight_helper.rb
@@ -13,13 +13,7 @@ module BlacklightHelper
     builder.rows = (user_params[:per_page] || user_params[:rows]) if user_params[:per_page] || user_params[:rows]
 
     builder = yield(builder) if block_given?
-    response = if default_search_query?(user_params)
-                 Rails.cache.fetch("#{Time.zone.today.day}/default_search_response", expires_in: 24.hours) do
-                   repository.search(builder)
-                 end
-               else
-                 repository.search(builder)
-               end
+    response = repository.search(builder)
 
     if response.grouped? && grouped_key_for_results
       [response.group(grouped_key_for_results), []]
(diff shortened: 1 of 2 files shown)


app/helpers/blacklight_helper.rb
```diff
@@ -13,13 +13,7 @@ def search_results(user_params)
...
 
     builder = yield(builder) if block_given?
-    response = if default_search_query?(user_params)
-                 Rails.cache.fetch("#{Time.zone.today.day}/default_search_response", expires_in: 24.hours) do
-                   repository.search(builder)
-                 end
-               else
-                 repository.search(builder)
-               end
+    response = repository.search(builder)
```
christinach: @jrgriffiniii the PR looks great! The only part I'm not sure is this one.

jrgriffiniii: I'm sorry that this wasn't clearly documented.  This removes the caching mechanism introduced in https://github.com/pulibrary/orangelight/pull/1356.

Unfortunately, I believe that this was partly (or fully) responsible for leading to the errors encountered in https://pulibrary.slack.com/archives/CA891BNR0/p1537192071000100.  As suggested by another team member, these problems were likely related to a cache having been invalidated.  Given that forcibly replacing these cached values on the server whenever they are corrupted would introduce a significant maintenance burden on all members of the team, simply removing this mechanism entirely seemed more reasonable.

christinach: Thank you for the explanation @jrgriffiniii.

PR #1382, opened 2018-09-19
Don't allow clients to request excessive pages of results

### LINKED ISSUE #1381: Don't allow paging without a query string or facet
Also, don't allow very high paging at all

escowles:
Fixes #1381

### DIFF
--- a/app/helpers/blacklight_helper.rb
+++ b/app/helpers/blacklight_helper.rb
@@ -274,7 +274,7 @@ module BlacklightHelper
     # @param [ActionController::Parameters] params
     # @return [Boolean]
     def search_parameters?(params)
-      params[:q].present? || params[:f].present?
+      !params[:q].nil? || params[:f].present?
     end
 
     # Determines whether or not the user is requesting an excessively high page of results
--- a/app/helpers/blacklight_helper.rb
+++ b/app/helpers/blacklight_helper.rb
@@ -8,6 +8,8 @@ module BlacklightHelper
 
   # (see Blacklight::SearchHelper#search_results)
   def search_results(user_params)
+    raise ActionController::BadRequest if excessive_paging?(user_params)
+
     builder = search_builder.with(user_params)
     builder.page = user_params[:page] if user_params[:page]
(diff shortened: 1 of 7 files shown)


app/helpers/blacklight_helper.rb
```diff
@@ -269,43 +271,19 @@ def truncated_link(doc, field, opts = { counter: nil }, length = 200)
...
-    # Determines whether or not this is a default query
+    # Determines whether or not the user is requesting an excessively high page of results
     # @param [ActionController::Parameters] params
     # @return [Boolean]
-    def default_search_query?(params)
-      empty_search_query?(params) && all_search_fields_query?(params) && first_page_query?(params) && !faceted_query?(params)
+    def excessive_paging?(params)
+      page = params[:page].to_i || 0
+      return false if page == 0
+      return false if search_paramters?(params) && page < 1000
```
jrgriffiniii: `search_paramters?(params)` is missing an `e`

escowles: 👍 I've fixed that typo and added a static 400 Bad Request error page so the display is better.

app/helpers/blacklight_helper.rb
```diff
@@ -269,43 +271,19 @@ def truncated_link(doc, field, opts = { counter: nil }, length = 200)
...
-      !params[:page]
-    end
-
-    # Determines whether or not the request facets upon any fields
     # @param [ActionController::Parameters] params
     # @return [Boolean]
-    def faceted_query?(params)
-      params[:f]
+    def search_parameters?(params)
+      params[:q].present? || params[:f].present?
```
tampakis: Change `params[:q].present?` to `params[:q].nil?`
With the current behavior clicking a blank search and then paging to the second one errors. I think we should allow this behavior.

escowles: 👍 but I'm going to create an issue for whether we should allow paging without a query or facet being selected.  I would really like to know why people are doing this and think about other ways to help them find what they're looking for.

escowles:
See https://github.com/pulibrary/orangelight/issues/1385

PR #1397, opened 2018-09-27
Wrong date range will not raise error, it will display a flash error message

christinach:
flash error message: 'The min year must be less than the max year.'

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -21,6 +21,10 @@ class CatalogController < ApplicationController
     redirect_to solr_document_path(params[:id])
   end
 
+  rescue_from BlacklightRangeLimit::InvalidRange do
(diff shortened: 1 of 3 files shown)


app/controllers/catalog_controller.rb
```diff
@@ -21,6 +21,10 @@ class CatalogController < ApplicationController
     redirect_to solr_document_path(params[:id])
   end
 
+  rescue_from BlacklightRangeLimit::InvalidRange do
+    redirect_to '/', :flash => { :error => 'The min year must be less than the max year.' }
```
escowles: I think we shouldn't use "min year" and "max year".  There are no labels to guide the user, and that wording doesn't seem like what I would expect a user to use.  Can you change the error message to be something like "The start year must be before the end year"?  I think that wording will be more easily understandable.

christinach: ok!

PR #1411, opened 2018-10-03
Validate key in users_params hash

### LINKED ISSUE #1389: [Orangelight/production] NoMethodError: undefined method `[]' for false:FalseClass
## Backtrace

    line 16 of [PROJECT_ROOT]/app/helpers/blacklight_helper.rb: search_results

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/54399/faults/39635992)

christinach:
1. Validate key in user_params hash. If there is one with trailing or leading whitespaces raise a Bad Request and rescue the error; Display a flash error and redirect to '/'.
2. Rescue all the BadRequest errors; Redirect to '/'; Display a flash error message.

closes #1389

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -25,6 +25,10 @@ class CatalogController < ApplicationController
     redirect_to '/', :flash => { :error => 'The start year must be before the end year.' }
   end
 
+  rescue_from ActionController::BadRequest do
+    redirect_to '/', :flash => { :error => 'This is not a valid request.' }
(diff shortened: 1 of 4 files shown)


christinach:
@escowles regarding this issue, I was thinking since it is a BadRequest to log the error and not to display a flash error message. Any thoughts?

escowles:
I think it's good to display a flash message.  @tpendragon said it was real users triggering this error, so it would be good to give them feedback about what went wrong.

PR #1450, opened 2018-10-30
Use new embedded UV URL

tpendragon:
Depends on pulibrary/figgy#2040


christinach:
@tpendragon can you rebase the PR?

PR #1451, opened 2018-10-31
[DO NOT MERGE] test new uv with figgy-staging url

christinach:


christinach:
It was used to test the url from catalog-staging against figgy-staging

PR #1454, opened 2018-11-01
moves excessive paging methods to search builder

tampakis:
For some reason blacklight_params doesn't work correctly when in the blacklight_helper file. Moving the paging methods back to the search builder prevents the error of the blacklight_params being undefined.

### DIFF
--- a/app/helpers/blacklight_helper.rb
+++ b/app/helpers/blacklight_helper.rb
@@ -14,7 +14,6 @@ module BlacklightHelper
   end
 
   def search_results(user_params)
-    raise ActionController::BadRequest if excessive_paging?(user_params)
     user_params_valid(user_params)
 
     builder = search_builder.with(user_params)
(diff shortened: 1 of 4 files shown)


app/helpers/blacklight_helper.rb
```diff
@@ -75,12 +74,6 @@ def series_title_results(solr_parameters)
...
 
-  def only_home_facets(solr_parameters)
-    return if search_parameters?(solr_parameters)
-    solr_parameters['facet.field'] = home_facets
-    solr_parameters['facet.pivot'] = []
-  end
-
   # only fetch facets when an html page is requested
   def html_facets(solr_parameters)
     return if blacklight_params[:format].nil? || blacklight_params[:format] == 'html' ||
```
christinach: what is the blacklight_params[:format] == 'html' ?

tampakis: that code doesn't get changed here, but it's the format of the page requested ie:
https://catalog.princeton.edu/catalog?format=html&search_field=all_fields&q=my+favorite+book
https://catalog.princeton.edu/catalog?format=json&search_field=all_fields&q=my+favorite+book

christinach: thanks

PR #1471, opened 2018-11-21
Fix Borrow Direct Query Parameter Encoding Bug

kevinreiss:
Requests issue tracking problem: https://github.com/pulibrary/requests/issues/392


config/requests.yml
```diff
@@ -13,7 +13,7 @@ defaults: &defaults
 development:
   <<: *defaults
   voyager_ub_id: 1@DB
-  bibdata_base: https://bibdata-staging.princeton.edu
+  bibdata_base: http://bibdata-staging.princeton.edu
```
tampakis: it's https

PR #1475, opened 2018-11-29
Add display none for matomo img tag

christinach:


christinach:
By adding display:none in the style it is not present in the dom and I'm thinking that the `https://lib-analytics.princeton.edu/piwik.php?` might not get executed.

tampakis:
If the link doesn't get executed maybe we should consider not tracking or IE 11/non-javascript users and remove the tag completely

christinach:
closing this issue since the element will be hidden.

PR #1491, opened 2018-12-03
Adding CSV download to bookmarks

### LINKED ISSUE #1448: Emailing a collection of bookmarks or citations only sends the first page
If you have a collection of bookmarked records, only first page of them get sent when emailing them. I think this is a bug in Blacklight.

### LINKED ISSUE #1467: CSV export for bookmarks
From SN@P:
```
I frequently copy titles, characters, call nrs etc from the catalog entry to lists.

I noticed that today, whatever I copy, add an extra line <img src="https://lib-analytics.princeton.edu/piwik.php?idsite=9&amp;rec=1" style="border:0" alt="" />

Can you make sure that that is no longer copied? It makes making lists etc. very cumbersome.


Martin Heijdra, Ph. D. 何義壯
Director, East Asian Library
```

escowles:
Fixes #1448
Fixes #1467

### DIFF
--- a/app/controllers/bookmarks_controller.rb
+++ b/app/controllers/bookmarks_controller.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'csv'
+
 class BookmarksController < CatalogController
   include Blacklight::Bookmarks
   configure_blacklight do |_config|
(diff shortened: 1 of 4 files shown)


tampakis:
This works well, I think it's a great feature! Did Martin mention anything about original script for the other fields? Original script can be included in publisher, description, series, and notes. I like your approach of combining multiple values for a field.

While there are many cases where we can separate the fields by script for publisher (similar to the author_display): ie https://catalog-staging.princeton.edu/catalog/10621132, there are also cases where we can't: https://catalog-staging.princeton.edu/catalog/10783280

By the way where did the idea to include notes come from? I'm not sure it will be that useful because we are only including the general note field (500). There are dozens of other notes fields that wouldn't get included in the CSV export.

escowles:
@tampakis Martin asked for author and title specifically to be separate — maybe the records he was working with didn't have original script variants of other fields?  I did see a few (like publisher).

I included notes because the spreadsheet Martin shared with us had those included.  But there are tons of notes, so I'm not sure which notes should be included.

PR #1495, opened 2018-12-04
Retrieves IIIF Manifests from the GraphQL endpoint on Figgy (rather than using indexed IIIF Manifest URLs)

### LINKED ISSUE #1360: Restructure the retrieval of IIIF Manifests for the Universal Viewer
It would probably be easiest to address this by extending the work on https://github.com/pulibrary/orangelight/compare/issues-1360-jrgriffiniii-graphql-uv?expand=1

jrgriffiniii:
Adjusts the integration of the UniversalViewer from Figgy and deprecates Poltergeist for headless Chrome as the browser environment for functional testing.  Resolves #1360


app/javascript/orangelight/figgy_manifest_manager.js
```diff
@@ -0,0 +1,207 @@
...
+    return elements[this.idx]
+  }
+
+  updateArkLinkElement() {
+    const arkLinkElement = this.getArkLinkElement()
+    if (!arkLinkElement) {
+      return
+    }
+
+    arkLinkElement.href = `#viewer-${this.idx + 1}`
```
escowles: We have many, many ARKs registered with `#view` URLs — can we preserve that id for the first viewer?

PR #1517, opened 2018-12-15
Traps subfield nil code-a value from the authors list

### LINKED ISSUE #1505: "Network error" message when clicking 'cite' button
[Orangelight/production] ActionView::Template::Error: undefined method `value' for nil:NilClass

## Backtrace

    line 72 of [PROJECT_ROOT]/app/models/concerns/blacklight/solr/document/marc.rb: export_as_mla_citation_txt
    line 12 of [PROJECT_ROOT]/lib/orangelight/middleware/invalid_parameter_handler.rb: call

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/54399/faults/42823098)

christinach:
Closes #1505

### DIFF
--- a/app/models/concerns/blacklight/solr/document/marc.rb
+++ b/app/models/concerns/blacklight/solr/document/marc.rb
@@ -262,6 +262,26 @@ module Blacklight
             # TODO: Raise if not present
             self.class.extension_parameters[:marc_format_type]
           end
+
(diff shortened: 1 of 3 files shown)


christinach:
Checks if subfield code a exists before calling the value. 
Removes uniq! which returns nil if there are no duplicates in the array.

PR #1533, opened 2019-01-02
Uses source-map as the default Webpack devtool for debugging in the production environment

### LINKED ISSUE #1530: Configure webpack to use source map
Source map is supposed to allow you to debug js even after it's been bundled by webpack. The package is listed in yarn.lock but it doesn't seem to be working in prod. Chrome just gives a blank file and FF gives `Error while fetching an original source: TypeError: NetworkError when attempting to fetch resource.`

jrgriffiniii:
Resolves #1530


jrgriffiniii:
I'm sorry for that @hackmastera , this has been fixed.

hackmastera:
Sorry, I misspoke. it seems like it would be good to have for all environments. Could you put it in defaults?

PR #1579, opened 2019-02-15
display multi-item availability status

### LINKED ISSUE #1576: Availability status for multi-item holdings don't display
When there are catalog holdings with multiple items/volumes associated, the main availability status is "All items are available." However, if an item within the holding is unavailable the status of "Some items may not be available" triggers. The status for the individual items that are unavailable should display, but they currently don't.

Example, the Recap Lewis Library holding: https://catalog.princeton.edu/catalog/857469

The first 3 checked out items should display on the catalog page: https://catalog.princeton.edu/requests/857469?mfhd=977097&source=pulsearch

tampakis:
Closes #1576.
Also removes the colon from the "Current issues" label to be consistent with the other labels.


tampakis:
Before:
screen shot 2019-02-15 at 5 20 48 pm

After:
screen shot 2019-02-15 at 5 21 12 pm

PR #1594, opened 2019-03-07
Add capistrano tasks to set environment variables on staging server

### LINKED ISSUE #1588: Add Capistrano task for switching OL staging server to point to Figgy staging
At the moment, catalog-staging points to production figgy. It would be helpful to be able to switch to figgy-staging. A capistrano task would make this easy.

eliotjordan:
Closes #1588


tampakis:
Looks fine to me. This could be a pattern that we can establish for switching between Voyager test and production environments within staging as well

PR #1600, opened 2019-03-12
 Add sneakers and processors for rabbitmq events

### LINKED ISSUE #1570: Numismatic sync
The collection is small, so we can probably reindex frequently and/or sync when there are changes.  But we don't want to bog down the OL index with frequent commits.

Child of #1567

eliotjordan:
Closes #1570

### DIFF
--- a/app/services/event_processor/delete_processor.rb
+++ b/app/services/event_processor/delete_processor.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+class EventProcessor
+  class DeleteProcessor < Processor
+    def process
+      index.delete_by_query "id:#{RSolr.solr_escape(id)}"
+      index.commit unless bulk?
+      true
+    end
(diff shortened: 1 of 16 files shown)


app/services/event_processor/delete_processor.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class EventProcessor
+  class DeleteProcessor < Processor
+    def process
+      index.delete_by_query "id:#{RSolr.solr_escape(id)}"
+      index.commit unless bulk?
```
tpendragon: Just to be sure, you tested this on staging and it was fine?

eliotjordan: Yes. It worked. We should setup auto-commits on the orangelight solr cores, though. For bulk updates.

PR #1609, opened 2019-03-19
Method to add link for facet search on a field

### LINKED ISSUE #1586: link_to_search config not working
Class year is supposed to be clickable as a facet: https://catalog.princeton.edu/catalog/dsp01zk51vk08g
https://github.com/pulibrary/orangelight/blob/master/app/controllers/catalog_controller.rb#L175

christinach:
closes #1586

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -172,8 +172,8 @@ class CatalogController < ApplicationController
     config.add_show_field 'contributor_display', label: 'Contributor(s)', helper_method: :browse_name
     config.add_show_field 'department_display', label: 'Department', helper_method: :browse_name
     config.add_show_field 'certificate_display', label: 'Certificate', helper_method: :browse_name
-    config.add_show_field 'class_year_s', label: 'Class year', link_to_search: true
 
+    config.add_show_field 'class_year_s', label: 'Class year', helper_method: :link_to_search_value
     # Linked fields pushed to top of supplemental info
     config.add_show_field 'subject_display', label: 'Subject(s)', helper_method: :subjectify
     config.add_show_field 'related_name_json_1display', hash: true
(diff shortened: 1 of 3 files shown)


app/controllers/catalog_controller.rb
```diff
@@ -294,7 +295,7 @@ class CatalogController < ApplicationController
     config.add_show_field 'recap_notes_display', label: 'RCP', helper_method: :recap_note
 
     # Numismatics fields
-    config.add_show_field 'issue_number_s', label: 'Issue', link_to_search: true, if: false
+    config.add_show_field 'issue_number_s', label: 'Issue', helper_method: :link_to_search_value
```
tampakis: keep the `if: false` so that the issue number can appear at the top of the page

christinach: nice @tampakis , thanks you. i didn't know this blacklight magic.

tampakis: Orangelight magic, but magic nonetheless

app/helpers/application_helper.rb
```diff
@@ -307,6 +307,15 @@ def name_title(args)
     end
   end
 
+  def link_to_search_value(args)
+    if args[:document][args[:field]].present?
```
tampakis: this line isn't needed, blacklight checks to make sure the field is present before running the method

app/helpers/application_helper.rb
```diff
@@ -307,6 +307,15 @@ def name_title(args)
...
+      args[:document][args[:field]].each_with_index do |field_t, i|
```
tampakis: the variable name `field_t` is misleading. `field` is consistent with the other methods

app/helpers/application_helper.rb
```diff
@@ -307,6 +307,15 @@ def name_title(args)
...
+        field_link = link_to(field_t, "/?f[#{args[:field]}][]=#{CGI.escape field_t}", class: 'search-name', 'data-toggle' => 'tooltip', 'data-original-title' => "Search: #{field_t}", title: "Search: #{field_t}")
+        args[:document][args[:field]][i] = field_link.html_safe if field_link
```
tampakis: the `if field_link` check isn't necessary. Neither is the `html_safe` because the `link_to` method returns an html-safe string (and allows the link html to render correctly)

app/helpers/application_helper.rb
christinach: I'll update this one but I don't know where this check is done.

tampakis:
Thanks @christinach !

Issue #1666, opened 2019-07-01
Migrate sitemap cron to whenever


hackmastera:
related to #1444

hackmastera:
The job ran successfully and I deleted the old version from the crontab. I checked ansible and I don't see it there for OL (just pulmap) so I think that's fine.

One remaining issue is that the way whenever writes to the log is always appending, which means the log either needs to be added to logrotate or truncated at the beginning of the job.

PR #1668, opened 2019-07-01
Add new job type so sitemap job will not be silent

hackmastera:
refs #1666


hackmastera:
just remembered you can run whenever locally to see the job it produces. this isn't right.

PR #1671, opened 2019-07-03
Excluding queries from the root.

escowles:


config/robots.yml
```diff
@@ -2,6 +2,8 @@
 defaults: &defaults
   sitemap_url: https://catalog.princeton.edu/sitemap.xml.gz
   disallowed_paths:
+    - '/?q=*'
+    - '/?f*'
     - '/*?q=*'
```
tampakis: Isn't `'/?q=*'` in the subset of this?

escowles: I'm not actually sure — I tried to look up the robots.txt syntax, and didn't find anything definitive about whether the `*` matches 0 characters or 1+ characters.

tampakis: well it doesn't hurt to have this too

PR #1676, opened 2019-07-08
Uses blacklight_unapi branch update-bl-7-bundler

christinach:
Uses local copy of blacklight_unapi gem with BL7 depencency


christinach:
will rename the branch and open a new pull request

PR #1678, opened 2019-07-08
Uses blacklight_unapi branch update-bl-7-bundler

christinach:
Uses blacklight_unapi branch update-bl-7-bundler


christinach:
closes #1681

PR #1707, opened 2019-07-12
Remove chromedriver-helper in favor of wedrivers

christinach:
closes 1657

Remove chromedriver-helper in favor of wedrivers
Update capybara to remove selenium deprecation warnings
allow webdrivers through webmock


christinach:
@eliotjordan I did try the browser options but I didn't see a difference. However I can add them to be consistent with the rest of the repos.

PR #1716, opened 2019-07-15
Update index header

christinach:
update input-group BS4
update select with BS4 custom-select
update index header grid col


christinach:
Screen Shot 2019-07-15 at 8 19 58 PM

christinach:
child of https://github.com/pulibrary/orangelight/issues/1682

PR #1769, opened 2019-08-13
restores json views to pre blacklight 7 format

### LINKED ISSUE #1737: Add more fields to the catalog JSON view

tampakis:
Closes #1737

### DIFF
--- a/app/views/catalog/index.json.jbuilder
+++ b/app/views/catalog/index.json.jbuilder
@@ -1,0 +1,4 @@
+json.response do
+  json.docs @presenter.documents
+  json.facets @presenter.search_facets
+end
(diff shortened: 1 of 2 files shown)


tpendragon:
This is kind of a bummer. Was there no other way? The new API is pretty nice.

tampakis:
My main concern is the library bento box and any other potential integrations that expect the current production behavior. I would rather keep it the same and then gradually introduce the new api. Not sure what others think.

tampakis:
Interested in what work was involved with pulmap. I played around with the api a bit and the thing that is frustrating is that it passes all the values through the helper methods and wraps them in html tags.

PR #1820, opened 2019-08-30
1796 search constraints

### LINKED ISSUE #1796: Fix layout of search constraints

tampakis:
Closes #1796.
Also updates bootstrap display utilities to inline and bumps up 1 breakpoint to match production

### DIFF
--- a/app/assets/stylesheets/components/buttons.scss
+++ b/app/assets/stylesheets/components/buttons.scss
@@ -52,13 +52,19 @@ a.clear-bookmarks {
 }
 
 .btn-outline-secondary {
-  color: $black;
+  color: $dark;
+  border-color: #ccc;
+  &:hover, &:active {
+    background-color: #ddd;
+    border-color: #bbb;
+    color: $dark;
+  }
 }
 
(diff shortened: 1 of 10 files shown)


christinach:
example:
Screen Shot 2019-08-30 at 12 45 10 PM

app/assets/stylesheets/components/buttons.scss
```diff
@@ -51,7 +51,15 @@ a.clear-bookmarks {
...
 
+.btn-outline-secondary {
+  color: $black;
+}
 
+.search-constraint-value:hover,
+.search-constraint-value > .btn-outline-secondary:active {
+  background-color: transparent;
+  color: $black;
+}
```
christinach: maybe it's better to add:
  border-color: #ccc;
  &:hover {
    background-color: #ddd;
    color: $dark;
  }
this way the border color and the hover will be closer to the ones in production.

christinach: ^ this comment is for the `.btn-outline-secondary` line 54 I added in in the wrong line.

PR #1829, opened 2019-09-05
update layout of account page

### LINKED ISSUE #1827: Update layout of Account page
blacklight-7
Screen Shot 2019-08-30 at 1 13 15 PM
production
Screen Shot 2019-09-04 at 1 26 09 PM

tampakis:
Closes #1827

### DIFF
--- a/app/assets/stylesheets/components/account.scss
+++ b/app/assets/stylesheets/components/account.scss
@@ -12,10 +12,6 @@
     }
   }
 
-  .account--sidebar {
-    padding-left: 0;
-  }
-
   .account--main {
     padding-right: 0;
   }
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/components/account.scss
```diff
@@ -32,6 +28,7 @@
 
     li {
       float: none;
+      width: 100%;
```
christinach: The width 100% could also be achieved by adding `w-100` in each class="nav-item". because it is wrapped in a flex element.
https://getbootstrap.com/docs/4.3/utilities/sizing/

tampakis: Cool didn't know about this. I updated the nav-items to include the w-100 class

PR #1883, opened 2019-10-15
Pointing to the new version of locator

carolyncole:
Since the new locator is available over the web no reason to proxy it through the library site.

I believe taking out the proxy will make things simpler to diagnose int he long run if something ever goes down.


tampakis:
@carolyncole this looked good to me in staging. Is this good to merge?

carolyncole:
@tampakis yes this PR should be good to merge

PR #1910, opened 2019-12-04
Adding handling for cases where bibdata env. variables are nil within the CourseReserveRepository

### LINKED ISSUE #1660: [Orangelight/production] NoMethodError: undefined method `values' for ["status", 500]:Array
## Backtrace

    line 109 of [PROJECT_ROOT]/app/repositories/course_reserve_repository.rb: block in courses=
    line 109 of [PROJECT_ROOT]/app/repositories/course_reserve_repository.rb: each
    line 109 of [PROJECT_ROOT]/app/repositories/course_reserve_repository.rb: map

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/54399/faults/36228596)

jrgriffiniii:
This may resolve #1660, although I am not certain why these values wouldn't be passed (or, if they are not, whether or not it would be best to let HoneyBadger log these errors).

### DIFF
--- a/app/repositories/course_reserve_repository.rb
+++ b/app/repositories/course_reserve_repository.rb
@@ -8,11 +8,18 @@ class CourseReserveRepository
 
     private
 
+      def bibdata_url
+        ENV['bibdata_base']
+      end
+
       def courses_response
-        Faraday.get("#{ENV['bibdata_base']}/courses")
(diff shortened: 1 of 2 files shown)


christinach:
The pr got stale and we dont see the error any more.

Issue #1866, opened 2019-09-16
"Bookmark all" button smaller than its friends


hackartisan:
![Screen Shot 2019-09-16 at 12 12 19 PM](https://user-images.githubusercontent.com/845363/64974702-54651b00-d87b-11e9-8f8f-911761382f83.png)

viewed on chrome

hackmastera:
similar issue on the bookmarks page

![Screen Shot 2019-09-16 at 12 13 38 PM](https://user-images.githubusercontent.com/845363/64974762-7494da00-d87b-11e9-8539-81aa091ae556.png)

tampakis:
That's weird. Which version of chrome are you using? This is what I get in `Version 76.0.3809.100 (Official Build) (64-bit)`:
<img width="1439" alt="Screen Shot 2019-09-16 at 12 27 19 PM" src="https://user-images.githubusercontent.com/4650153/64975610-72cc1600-d87d-11e9-85a6-de3247e3b925.png">
<img width="1420" alt="Screen Shot 2019-09-16 at 12 27 28 PM" src="https://user-images.githubusercontent.com/4650153/64975612-72cc1600-d87d-11e9-833b-0750e1180bf2.png">

hackmastera:
mine says Version `76.0.3809.100` hmmm that's the same version as you.

hackmastera:
Hm. the search box is wonky on mine also. hard refresh / incognito mode don't seem to affect it :\
![Screen Shot 2019-09-16 at 12 44 31 PM](https://user-images.githubusercontent.com/845363/64976612-c3447300-d87f-11e9-893e-988dcc42a303.png)

tpendragon:
@hackmastera is going to try to find something different about her setup vs ours so we can see if we can reproduce. If not we will close this. Will work with @christinach to fiddle with things a little.

hackmastera:
I'm not getting any errors on the network tab with cache disabled.

Looking at the search box, for example, the dropdown height is getting calculated to 52 px, but the search box is getting set directly to 42px. 

https://github.com/pulibrary/orangelight/blob/13bae27ba14ebd8a9b0840efc773acb8125b4690/app/assets/stylesheets/components/search.scss#L98

Adding `height:100%` here fixes that particular case
https://github.com/pulibrary/orangelight/blob/13bae27ba14ebd8a9b0840efc773acb8125b4690/app/assets/stylesheets/components/search.scss#L52

[...]

PR #1914, opened 2019-12-16
Ensure search form dropdown and input are same height

hackmastera:
refs #1866 (fixes the search box portion of that ticket but not the issue with
the bookmark buttons)


christinach (approved):
@hackmastera I don't see the issue that you see in your local environment; but I will merge this because it is a good idea the height for `#search_field.custom-select` and `.search-btn` to be calculated in the same way

PR #1935, opened 2020-02-14
Document the fact that reserves core cannot be on a different server

hackmastera:


tpendragon:
Maybe an ADR for this too? We do a cross-core join because the reserves core is updated at the same time as the query, and then instantly committed. Commits on the main core were too slow.

hackmastera:
I would be 👍 on an ADR for this. I wasn't a part of that decision so maybe someone else could write that up. Someone could push it to this branch or it could be separate.

PR #1964, opened 2020-03-18
Covid announcement

eliotjordan:


eliotjordan:
CI is not working currently, but we need to get this up ASAP.

PR #1970, opened 2020-03-19
Home page updates

christinach:
Update the digital collections link with the DPUL link
Remove the Welcome to the Catalog
Remove the about/tour links
Adds temporary message for the Hathi Trust access


christinach:
During the last DSSG meeting @tampakis suggested to place a notice on the top part of the home page where we will prompt the user to check the 'Online content.' I'll add a screenshot.

escowles:
One thing that was discussed yesterday is that once the HathiTrust emergency temporary access is live, we should add some prominent text to the OL home page about it.

Suggested text:
> Many print items are now available online from Hathi Trust. Look for the "Temporary Digital Access from Hathi Trust" link in the Available Online section.

christinach:
I will add this.

christinach:
Screen Shot 2020-04-03 at 10 30 16 AM

christinach:
@escowles can you check? I also removed the Online Content I had added in the 'Other Library Resources'

christinach:
I see an hr line. let me fix this

escowles:
@christinach That looks OK, but I think it would be better to put a border and/or colorful background around it or otherwise set it off from the rest of the page so it doesn't get lost.

christinach:
@escowles when you have a chance 
Screen Shot 2020-04-03 at 11 22 09 AM

PR #1983, opened 2020-03-26
Update to Bundler 2

tpendragon:


config/deploy/staging.rb
```diff
@@ -4,7 +4,7 @@
 set :rails_env, 'staging'
 set :branch, ENV['BRANCH'] || 'master'
 
-server 'catalog-staging1', user: 'deploy', roles: %i[web app db]
+server 'catalog-staging1', user: 'deploy', roles: %i[web app db worker]
```
christinach: do we need the worker on staging?

tpendragon: I don't know, but it's there already

christinach: ok

Issue #1998, opened 2020-04-03
[Orangelight/production] JSON::ParserError: 784: unexpected token at '<html><body><h1>503 Service Unavailable</h1>
No server is available to handle this request.
</body></html>


eliotjordan:
## Backtrace

    line 21 of [PROJECT_ROOT]/app/services/hathi_url.rb: build_hathi_url
    line 7 of [PROJECT_ROOT]/app/services/hathi_url.rb: initialize
    line 643 of [PROJECT_ROOT]/app/controllers/catalog_controller.rb: new

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/54399/faults/62537367)

PR #1999, opened 2020-04-03
Adding a catch and retry for the JSON parse error

carolyncole:
foxes #1998

### DIFF
--- a/app/services/hathi_url.rb
+++ b/app/services/hathi_url.rb
@@ -1,9 +1,12 @@
 # frozen_string_literal: true
 class HathiUrl
-  attr_reader :url
+  attr_reader :url, :maximum_retries, :sleep_duration
 
   # rubocop:disable Lint/UnusedMethodArgument
-  def initialize(oclc_id:, isbn:, lccn:)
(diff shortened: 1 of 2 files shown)


app/services/hathi_url.rb
```diff
@@ -20,6 +21,11 @@ def build_hathi_url(key:, id:)
       json_data = Net::HTTP.get(uri)
       hathi_data = JSON.parse(json_data)
       hathi_url(hathi_data['items'].first['htid']) if hathi_data['items'].count.positive?
+    rescue JSON::ParserError => e
+      Rails.logger.warn(e.message)
+      @retry_count += 1
+      sleep 10
+      retry if @retry_count < 10
```
escowles, line 28: Given the traffic that Hathi might be seeing, should we have a longer wait time and fewer retries? Since this is being checked in the controller, we definitely wouldn't want to wait very long, but I do wonder if 10 rapid requests would get any different response.

carolyncole: @escowles with my querying for data validation I waited 30 seconds, and always got a response on the next try.  Given this was user facing I dropped it to 10 seconds, since 30 seconds is ages for a user.  I would be willing to change either number, but thought if they are really slammed then just not showing the link might be optimal, since they may also not be able to show the user the book.

PR #2002, opened 2020-04-03
Adding the Hathi logo to the home page

carolyncole:
also added an alert icon to the home page
change hathi_ap_url to hathi_api_url also


app/views/catalog/_home_text.html.erb
```diff
@@ -14,9 +14,16 @@
...
-      Look for the "Temporary Digital Access from Hathi Trust" link in the "Available Online" section.
-  </div>
+  <card class="card card-body bg-light">
+      <card-media>
+        <%= link_to "https://www.hathitrust.org/covid-19-response" do %>
+          <%= image_tag("250px-HathiTrust_logo.svg.png") %>
+        <% end %>
+      </card-media>
+      <card-content>    
+        Many print items are now available online from Hathi Trust.
```
christinach, line 24: can we change the icon-warning to somtheing else? maybe icon info?

christinach:
@carolyncole looks great. Can we fix the coverage if it's still failing? It failed earlier.

carolyncole:
@christinach I think it was just failing because the test runs were terminating early. Looks ok now...  Just waiting on 2_6

christinach:
yes it looks green now!

Issue #1944, opened 2020-02-25
Update Ruby to 2.6


tpendragon:
- [x]  Add Ruby 2.6 to CircleCI
- [x]  Fix any failing tests or dependency issues caused by upgrade to 2.6.
- [x]  Deploy code that hopefully works with both 2.4 and 2.6.
- [x]  Provision boxes to use Ruby 2.6.
- [x]  Remove Ruby 2.4 from CircleCI

tpendragon:
Orangelight has some work going on, so putting this back to "To Do". Some in progress work: 

https://github.com/pulibrary/orangelight/tree/update_bundler updates bundler
https://github.com/pulibrary/princeton_ansible/compare/update_orangelight_ruby updates staging to be Ruby 2.6

christinach:
we should also update https://github.com/pulibrary/requests/issues/442

PR #2021, opened 2020-04-21
Update .ruby-version to 2.6.5

hackmastera:
The upgrade on the server is complete.


hackmastera:
Rubocop isn't happy. Maybe we can't really do this without also finishing https://github.com/pulibrary/orangelight/issues/1944.

PR #2034, opened 2020-05-13
Numismatics as a non voyager holding

christinach:


christinach:
@kevinreiss can you check please? I deployed on staging. If it is ok we should merge the requests PR first and then I'll point this one to the new version of the requests gem.

PR #2046, opened 2020-05-21
[Dont merge] Updating to utilize hathi_identfier_t in hathi url

carolyncole:


christinach:
We never indexed the hathi url.

PR #2093, opened 2020-06-09
Adding no bar code option for requests

carolyncole:


kevinreiss:
This is looking good. I think we can deploy. Tested it in staging under several scenarios, ReCAP, On Campus materials as a CAS authenticated and Barcode Authenticated user.

PR #2195, opened 2020-08-21
Paginate solr using cursorMark instead of start

### LINKED ISSUE #1763: Browse lists not generating
## Backtrace

    line 77 of [PROJECT_ROOT]/lib/orangelight/browse_lists.rb: block (2 levels) in browse_cn
    line 73 of [PROJECT_ROOT]/lib/orangelight/browse_lists.rb: times
    line 73 of [PROJECT_ROOT]/lib/orangelight/browse_lists.rb: block in browse_cn

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/54399/faults/50525064)

hackmastera:
This PR:
* Refactors browse_cn into a new class
* Updates all these browse jobs to write to uniquely-named log files (the ones on machine 1 were writing to the same file, confusingly)
* Adds retries and error handling
* Rewrites the solr page fetches to use cursorMark instead of start. This is the recommended way to avoid the deep pagination performance problem.

fixes #1763


hackmastera:
Did a test run and it got all the rows. Completed in 1 hr 5 min. Ready for review.

lib/orangelight/browse_lists/call_number_csv.rb
```diff
@@ -0,0 +1,131 @@
...
+            csv << [sort_cn, mcn, 'ltr', '', "#{record_count} titles with this call number", '', '', "?f[#{facet_field}][]=#{CGI.escape(mcn)}", '', 'Multiple locations']
+          end
+        end
+      end
+
+      # Append the rest of the call numbers to the file
+      def write_remaining_call_numbers
+        cursor_mark = '*'
+        CSV.open(filename, 'ab') do |csv|
+          loop do
```
tpendragon, line 54: Suggestion: It seems like this loop logic along with solr_cursor_page_body could be extracted to an enumerator, so this logic could just become
```ruby
CSV.open(filename, 'ab') do |csv|
  call_number_records.each do |record|
    csv << record.row
  end
end
```

rather than having it all be procedural.

lib/orangelight/browse_lists/call_number_csv.rb
```diff
@@ -0,0 +1,131 @@
...
+            end
+            next_cursor_mark = body['nextCursorMark']
+            break if cursor_mark == next_cursor_mark
+            cursor_mark = next_cursor_mark
+          end
+        end
+      end
+
+      # Get a page of solr results for actual items starting at the given number
+      def solr_cursor_page_body(cursor_mark)
```
tpendragon, line 72: question: Is the cursor mark a number? Curious about the comment here. Maybe a YARD-doc would help here, if it isn't refactored.

lib/orangelight/browse_lists/call_number_csv.rb
```diff
@@ -0,0 +1,131 @@
...
+          req = JSON.parse(resp.body)
+          break if req['response']
+          Rails.logger.error "Call number browse generation failed at iteration with cursor mark #{cursor_mark}. Response from solr was: #{resp}"
+          raise SolrResponseError if retries >= 2
+          retries += 1
+        end
+        req
+      end
+
+      def parse_cn_row(record, cn, sort_cn)
```
tpendragon, line 88: suggestion: Extracting this to a class would make it easier to look at, and would make the above refactors to an enumerator easier as well.

lib/orangelight/browse_lists/call_number_csv.rb
```diff
@@ -0,0 +1,131 @@
...
+              location = holding_record[holding_id]['location']
+            end
+          end
+        end
+        holding_id ||= ''
+        location ||= ''
+        [sort_cn, label, dir, '', title, author, date, bibid, holding_id, location]
+      end
+
+      # determines if there are multiple locations for the same call number and same bib
```
tpendragon, line 125: thought: If you extract a record object above then this method could be a private method on that object.

lib/orangelight/browse_lists/call_number_csv.rb
hackmastera: Oh this comment is wrong since it was changed from using the start number. thanks.

hackmastera: oh, and it's not a number it's an encoded string. example: `AoEoMTAwMDA1NjY=`

hackmastera:
These are great suggestions for further refactoring but I need to turn to some other issues. Created https://github.com/pulibrary/orangelight/issues/2199 and maybe we can get back to it in a refactor meeting.

PR #2196, opened 2020-08-24
Replace Matomo with Google Analytics

### LINKED ISSUE #2168: Use Google Analytics to record analytics in production
Replace Matomo analytics tracking with Google Analytics to record analytics in production.

No analytics should be recorded on staging.

kelynch:
Closes #2168 

I'm coordinating with Axa to make sure deploying this won't disrupt their generating the LLT usage statistics reports, hence "draft."


carolyncole:
Can we add an event in google analytics for when we link out to the library.princeton.edu/resolve and Hathi links?

kelynch:
Sure, I'll add a custom GA event during Maintenance Monday.

In the meantime, I've re-added Matomo as we discussed in yesterday's check-in.

kelynch:
@carolyncole and @axamei, what information would you like to collect on this event?  Just the value of the proxied URL, or something additional?

.rubocop.yml
```diff
@@ -73,6 +73,7 @@ Metrics/CyclomaticComplexity:
     - 'lib/orangelight/browse_lists/call_number_csv.rb'
 
 Metrics/LineLength:
+  Max: 250
```
hackmastera, line 76: Could we use an inline exception around the one test instead of changing this globally?

hackmastera (changes requested):

.rubocop.yml
kelynch: Sure, I can adjust this.

kelynch:
I've removed the GA work from this branch in the interest of the ability to deploy Google Analytics to Orangelight.

I've created a separate issue for this request and the Hathi event, where I'll be happy to address the need - https://github.com/pulibrary/orangelight/issues/2285

.rubocop.yml
kelynch: I've reverted the commits, to be addressed as part of #2285.

hackmastera (approved):

PR #2198, opened 2020-08-25
[DO NOT MERGE] Add hathi url endpoint

### LINKED ISSUE #2184: Create `/catalog/<bibd>/hathi.json` to return the Hathi URL
## Technical Implementation

Use https://github.com/pulibrary/orangelight/blob/master/app/services/hathi_url.rb to get the Hathi URL. It should return a JSON response like
```
{
  hathi_url: "blabla"
}
```

If the URL is nil, it should return a 404.

This will be used by javascript to add a Hathi link to the show page.

hackmastera:
It looks like the hathi api is public in the end, and so we could write this into the javascript side instead. We should discuss before merging.

closes #2184

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -645,4 +645,13 @@ class CatalogController < ApplicationController
   rescue ActionController::BadRequest
     render file: Rails.public_path.join('x400.html'), layout: true, status: :bad_request
   end
+
+  def hathi
+    _, document = search_service.fetch(params[:id])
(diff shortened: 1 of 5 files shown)


hackmastera:
It looks like the hathi api is public, and so we could write this into the javascript side instead. We should discuss before merging.

tpendragon:
Yeah, I was under the impression that the Hathi request used some sort of API key. It doesn't look like it does, so we should just ping that API directly in the JS.

hackmastera:
Closing. This was fun but we'll move it to javascript.

PR #2208, opened 2020-08-27
Handle hathi response with no records or items

### LINKED ISSUE #2203: Add error handling to HathiConnector javascript
In the server-side code we had retries.

hackmastera:
refs #2203

### DIFF
--- a/app/javascript/orangelight/hathi_connector.es6
+++ b/app/javascript/orangelight/hathi_connector.es6
@@ -1,7 +1,7 @@
 import { insert_online_link } from 'orangelight/insert_online_link'
 
 export default class HathiConnector {
-  async insert_hathi_link(oclc_number) {
+  async insert_hathi_link() {
     // The API also offers isbn or lccn but we're just using oclc number
     const url = `https://catalog.hathitrust.org/api/volumes/brief/oclc/${this.oclc_number()}.json?callback=?`
     return $.getJSON(url).promise()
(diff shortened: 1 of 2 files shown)


carolyncole:
@hackmastera I do not think this will catch the actual error.  Is there a way for you to check the status of the response fro Hathi and then just not run the json parse?  When there is an issue it returns HTML not json.

hackmastera:
@carolyncole This is just to handle an empty response. See the last comment on the issue regarding the retry decision.

hackmastera:
Added the html handling and ready for re-review

app/javascript/orangelight/hathi_connector.es6
```diff
@@ -10,9 +10,15 @@ export default class HathiConnector {
 
   process_hathi_data(hathiData) {
     const base_url = "https://babel.hathitrust.org/Shibboleth.sso/Login?entityID=https://idp.princeton.edu/idp/shibboleth&target=https%3A%2F%2Fbabel.hathitrust.org%2Fcgi%2Fpt%3Fid%3D"
-    let mdp_id = hathiData.items[0].htid
-    let url = base_url + mdp_id
-    insert_online_link(url, "hathi_link")
+    try {
```
carolyncole, line 13: @hackmastera  I think the try needs to be around line 7 . I would expect the getJSON to be the thing to throw an exception.

PR #2225, opened 2020-09-01
Google Book Snippet Preview

### LINKED ISSUE #2178: Display "snippets" from Google Books if possible
Displaying Google snippets of non-public items will help users evaluate items, and reduce unnecessary pickup or scanning requests.

This is only for materials that we don't have a better alternative for (eg. Hathi, already-completed CDL).
## Additional implementation details:

* UPDATE 9/10/2020 - Implement this feature using the link out to Google Book Search (example: https://developers-dot-devsite-v2-prod.appspot.com/books/examples/previewwizard-linktogbs), not the embed. 
* Display disclaimer text stating that this is a preview, and that PUL does not recommend/endorse purchasing options on the following page ([text for this](https://docs.google.com/document/d/1xbjEzo4p0vcblY9-KE6Fsv73kNHj151iJysYMCSJnvc/edit)). Initially, display text in line with the link itself. If the text is too long, use a click-through modal that is presented only once per session.
* If possible, link out to the New Google Books view of the page (de-emphasizes purchase). Example: [old view](https://books.google.com/books?id=VpNa9UckT24C&printsec=frontcover&source=gbs_ViewAPI&hl=en#v=onepage&q&f=false) | [new view (preferred)](https://www.google.com/books/edition/_/VpNa9UckT24C?hl=en&gbpv=1&pg=PP1)
* Add the snippet link to the "Show" page in the catalog, where the UV display for items with digital representations
* Make it clear through labeling the link that this is a snippet, not the full book (based on user feedback from a previous implementation)
* Only attempt to match ISBN numbers
* Only display this for items that are print monographs (no serials, no e-resources)
* ~~Link the thumbnail image in the "Show" page to the snippet viewer at the bottom of the page, if possible~~

tpendragon:
Closes #2178 

The only part this doesn't do is "Only display this for items that are print monographs (no serials, no e-resources)." I wasn't sure how to identify these - is there a clearer identifier I should be looking for? Maybe @christinach or @escowles know?

Screen Shot 2020-09-01 at 2 08 26 PM


app/javascript/orangelight/google_books_snippets.es6
```diff
@@ -0,0 +1,28 @@
+export default class GoogleBooksSnippets {
+    get has_figgy_viewer() {
+        return $(".document-viewers > .intrinsic-container").length > 0
```
hackmastera, line 3: I'm realizing this implementation would also return true if it's there's a google books viewer. We're not using it that way but we should either rename it or make it only return true if there's a figgy viewer.

spec/javascript/orangelight/google_books_snippets.spec.js
```diff
@@ -0,0 +1,51 @@
...
+    expect(google_books_snippets.has_figgy_viewer).toBe(false)
+  })
+
+  test("isbn returns the meta isbn value", () => {
+    document.body.innerHTML = '<meta property="isbn" itemprop="isbn" content="9780618643103">'
+    const google_books_snippets = new GoogleBooksSnippets
+    expect(google_books_snippets.isbn).toEqual(["9780618643103"])
+  })
+
+  test('insert_snippet adds the google books snippet inside the document-viewers container', () => {
```
hackmastera, line 29: If we're making other changes can we add "and decorates the thumbnail" or similar

mzelesky:
For no e-resources, you would ignore records that only have 'elf' locations (elf1, elf2, elf4). To ignore serials, you could ignore anything with the 'format' in the solr of 'Journal'. Does that help?

christinach:
@tpendragon I forgot to add that some times I had to refresh the page more than once for the viewer with the google snippet to load.

tpendragon:
Closing this PR. We're going to add a link instead.

PR #2227, opened 2020-09-02
Determine request button label using hathi/access lookup

### LINKED ISSUE #2207: Items in Columbia's ETAS program should not show "Pick-up" in the request button, but others should.
This is a follow-up to https://github.com/pulibrary/orangelight/issues/2206

Use the API call from https://github.com/pulibrary/marc_liberation/issues/758

hackmastera:
To determine whether Columbia items can be requested for pickup

fixes #2207


app/services/physical_holdings_markup_builder.rb
```diff
@@ -293,21 +293,31 @@ def self.location_services_block(adapter, holding_id, location_rules, link, hold
...
     else
       'Request'
     end
   end
 
+  def self.etas_restricted?(location_rules, adapter)
+    return false unless location_rules["code"] == "scsbcul"
+    bibid = adapter.document.fetch("other_id_s")
+    access_records = adapter.hathi_access(bibid)
+    return true if access_records.select { |r| r["origin"] == "CUL" }.select { |r| r["status"] == "DENY" }.present?
```
carolyncole, line 317: @hackmastera I see we used slightly different logic.  https://github.com/pulibrary/requests/blob/master/app/models/requests/request.rb#L422

I wonder if we should both switch to something more like yours 
```
access_records.select { |r| r["origin"] == "CUL" && r["status"] == "DENY"  }.present?
```

There is no need to return true of false since the check would return the correct value.  Are you doing it that was to make it more explicit?

app/services/physical_holdings_markup_builder.rb
```diff
@@ -293,21 +293,31 @@ def self.location_services_block(adapter, holding_id, location_rules, link, hold
 
   # Generate a request label based upon the holding location
   # @param location_rules [Hash] the location for the holding
+  # @param adapter [HoldingRequestsAdapter] adapter for the SolrDocument and Bibdata API
   # @return [String] the label
-  def self.request_label(location_rules)
+  def self.request_label(location_rules, adapter)
```
carolyncole, line 298: Curious about the decision to pass adapter here instead of the just columbia id.   Do you see the adapter being use more in the future?

app/services/physical_holdings_markup_builder.rb
hackmastera: Yep I like this suggestion. The explicit returns were just a holdover from my tdd process.

app/services/physical_holdings_markup_builder.rb
hackmastera: I was being lazy and didn't want to update or write more tests for the calling method. Do you think I should change it?

carolyncole: It's ok, it just makes it look like we are dependent on more than just the id.  I would not worry about it for the moment though...

carolyncole (approved):

carolyncole:
I'm not merging sine it still has bibdata-staging.  I believe the process for putting this live would be 

- [ ] change the PR to bibdata
- [ ] Merge the PR
- [ ] Deploy the code to prod
- [ ] Update Bibdata prod to allow SCSBCUL to circulate

christinach:
In this PR the requests.yml can keep the bibdata-staging. It refers to the staging environment.

christinach (approved):

PR #2248, opened 2020-09-09
Don't add the hathi link if the item is checked out

### LINKED ISSUE #2237: ETAS links should not show if an item is charged
There is a consensus that if we can we should try and suppress ETAS links for items that move into a charged status in Voyager.

hackmastera:
Includes some code reorganization refactors
fixes #2237


christinach:
@hackmastera I'm approving this now but it has a conflict.

christinach:
oh nice you just pushed :)

christinach (approved):

PR #2252, opened 2020-09-11
Make check for checked-out statuses case-insensitive

### LINKED ISSUE #2237: ETAS links should not show if an item is charged

hackmastera:
closes #2237 again


kevinreiss (approved):
Thanks @hackmastera.

PR #2255, opened 2020-09-15
Updating Request to allow for ReCAP In Library Use Requests

carolyncole:
refs https://github.com/pulibrary/requests/issues/792
refs https://github.com/pulibrary/requests/issues/793


kevinreiss (changes requested):
Circulation has asked for a few more changes to the patron confirmation email text. They are getting back with the final version of what they would like.

kevinreiss (approved):

PR #2256, opened 2020-09-15
Google Preview Link

### LINKED ISSUE #2178: Display "snippets" from Google Books if possible

tpendragon:
Closes #2178


spec/javascript/orangelight/insert_online_link.spec.js
```diff
@@ -33,4 +33,15 @@ describe('insert_online_link', function() {
     expect(anchor.getAttribute("href")).toEqual("#view")
     expect(anchor.getAttribute("target")).toEqual("_self")
   })
+
+  test("insert_online_link() can accept a content builder", () => {
```
hackmastera, line 37: 👍

app/javascript/orangelight/google_books_snippets.es6
```diff
@@ -0,0 +1,42 @@
...
+  process_google_response(response) {
+    for(const key in response) {
+      const result = response[key]
+      if(result.preview === "partial" || result.preview === "full")
+      {
+        const url = new URL(result.preview_url)
+        const bookId = url.searchParams.get("id")
+        const link = `https://www.google.com/books/edition/_/${bookId}?hl=en&gbpv=1&pg=PP1`
+        const content = (link, target) => {
+          return `<a href="${link}" target="${target}">Google Preview</a><p>${this.contentStatement}</p>`
```
hackmastera, line 35: Specify "partial" or "full" in the link text?

hackmastera (changes requested):
I think it would be nice to indicate whether it is a full or partial preview.

Also I think the requirement to exclude serials and journals is not yet addressed. Is this desired before the feature is deployed or should we create a follow-up issue? @mzelesky's comment on the previous PR: https://github.com/pulibrary/orangelight/pull/2225#issuecomment-685734042

app/javascript/orangelight/google_books_snippets.es6
tpendragon: If it's full is it even a preview anymore? Possible names fluttering in my head are `Google Books (Partial)`, `Google Preview (Partial)`, `Google Book Viewer (Partial)`, `Google Book Viewer`. Maybe input from @escowles or @jpstroop ?

jpstroop: I thought we we scoping to previews only and that the full views would be covered by Hathi, but you're right, there are probably exceptions. `Google Books (Partial View)` and `Google Books (Full View)` makes sense to me. I'll look at the disclaimer text too and see if that needs adjusting.

jpstroop: Disclaimer updated once again: https://docs.google.com/document/d/1xbjEzo4p0vcblY9-KE6Fsv73kNHj151iJysYMCSJnvc/edit

tpendragon: Screen Shot 2020-09-16 at 10 08 22 AM

tpendragon:
I added not running this code if the item's a Journal. I'm not really sure how to do e-resources only. @mzelesky do you have an example on hand of one in the catalog? I need something in the DOM to key off of, ideally.

tpendragon:
@hackmastera Okay I think this is finally ready.

spec/requests/request_spec.rb
```diff
@@ -347,6 +347,16 @@
     end
   end
 
+  describe 'location metadata' do
```
hackmastera, line 350: I don't think I'd expect to find this test in this file. How did you decide to put it here?

hackmastera (changes requested):
I just have the one question inline

spec/requests/request_spec.rb
tpendragon: This is where the test for the identifiers was - seemed like where the catalog show specs are.

hackmastera (approved):

PR #2257, opened 2020-09-15
Use Lando for the database.

tpendragon:
Provides a more consistent database setup between developers.

### DIFF
--- a/.lando.yml
+++ b/.lando.yml
@@ -12,6 +12,9 @@ services:
     core: orangelight-core-dev
     config:
       dir: "solr/conf"
+  orangelight_database:
+    type: postgres:10
+    portforward: true
 proxy:
   orangelight_test_solr:
     - orangelight.test.solr.lndo.site:8983
--- a/README.md
+++ b/README.md
@@ -27,8 +27,8 @@ rake db:create
 rake db:migrate
 rake db:seed
 ```
-Production credentials: In production you'll need to add production
-credentials to database.yml
+
+You can access the database console via `rails dbconsole -p`.
 
 ### Run the development Environment locally
 **All commands are assumed to be run from your local orangelight directory**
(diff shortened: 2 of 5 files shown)


.lando.yml
```diff
@@ -12,6 +12,9 @@ services:
     core: orangelight-core-dev
     config:
       dir: "solr/conf"
+  orangelight_database:
+    type: postgres:10
+    portforward: true
```
christinach, line 17: can we use postgres 12?

christinach (changes requested):

.lando.yml
tpendragon: Is that what's in prod?

christinach: yes

christinach: Wait. no in production is 10. I was looking the qa it has 12.

tpendragon: Yeah, `PostgreSQL 10.14 (Ubuntu 10.14-1.pgdg16.04+1) on x86_64-pc-linux-gnu, compiled by gcc (Ubuntu 5.4.0-6ubuntu1~16.04.12) 5.4.0 20160609, 64-bit`

christinach (changes requested):
@tpendragon when I use this PR and create the database both in development and in test I'm not able to see them in postgres. Is there another step with lando that I'm missing?

README.md
```diff
@@ -44,7 +30,7 @@ rake db:seed
 Production credentials: In production you'll need to add production
```
christinach, line 30: Before this line 'Production credentials can you add `Connect to the pg console:` `rails dbconsole -p`. Basically what you mentioned in slack.

christinach (changes requested):

README.md
tpendragon: Done. I also removed this line, it didn't really make sense to me.

christinach (approved):

PR #2262, opened 2020-09-21
Remove unused lines from application.css

### LINKED ISSUE #1276: Check for chosen-rails gem upgrade
So we won't need to have a local gem override https://github.com/pulibrary/orangelight/pull/1271/files

christinach:
closes #1276 

With the chosen-rails upgrade to 1.8.7 we no longer need to upload image-url('chosen-sprite.png') and image-url('chosen-sprite@2x.png') before importing the chosen-rails gem.
for more info see: https://github.com/tsechingho/chosen-rails/pull/101/files and https://github.com/tsechingho/chosen-rails/issues/97

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -12,11 +12,10 @@
  *
  *= require  'blacklight_range_limit'
  *= require 'blacklight_advanced_search'
+ *= require chosen
  *= require introjs
  *= require_self
  */
-
-@import 'chosen';
 
 @import 'variables/breaks';
 @import 'variables/colors';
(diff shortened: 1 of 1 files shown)


hackmastera (changes requested):
Should we also go back to using the `require` instead of `import` as in https://github.com/pulibrary/orangelight/pull/1271/files?

It would align with the documented setup: https://github.com/tsechingho/chosen-rails#include-chosen-stylesheet-assets

christinach:
@hackmastera thanks, can you check again?

hackmastera (approved):

Issue #2271, opened 2020-09-24
Borrow Direct Authentication Should Authorize Users Based on the campus_authorized flag


kevinreiss:
Our borrow direct authentication should reject users who aren't eligible to be on campus. If campus_authorized in the patron response from Bibdata is "N" for this value the user should be treated as ineligible for campus service. 

https://github.com/pulibrary/orangelight/blob/d96191e8ea34357f68644f6736dc29727872bb9b/app/controllers/account_controller.rb#L98

PR #2272, opened 2020-09-25
Add campus_authorized check to borrow direct authorization. 

kevinreiss:
For #2271


spec/factories/users.rb
```diff
@@ -11,6 +11,12 @@
     factory :valid_princeton_patron do
     end
 
+    factory :campus_authorized_patron do
```
carolyncole, line 14: @kevinreiss I do not think these are needed.

carolyncole (changes requested):

spec/factories/users.rb
kevinreiss: Updated

carolyncole (approved):

PR #2275, opened 2020-09-28
remove bd campus auth check.

kevinreiss:


hackmastera (changes requested):
Can you create an issue that explains why, and then reference that issue from this PR? What sort of timeline this choice is likely to have? In general deleting code would be preferred over commenting it out. Another issue could be created referencing this PR if we anticipate restoring it in the future.

carolyncole (changes requested):
Doesn't the user at least need to be trained?

Issue #2285, opened 2020-09-30
Add Google Analytics events for proxied online resource links


kelynch:
Can we add an event in google analytics for when we link out to resources in the "Onlline" section?

Here for resolve: https://github.com/pulibrary/orangelight/blob/b6b86181d2df46ada8ec5e4e2fe74517de2eb41b/app/services/online_holdings_markup_builder.rb#L34

Here is an example from stack overfloe from @axamei : https://stackoverflow.com/questions/32553637/rails-link-to-with-onclick-to-google-analytics-event

Hathi has not been implemented yet...

_Originally posted by @carolyncole in https://github.com/pulibrary/orangelight/pull/2196#issuecomment-680071788_

Note: We no longer have Hathi ETAS links

kevinreiss:
Also add this tracking to Google Books rendered in the online access section.

sandbergja:
In Google Analytics 4, events can be created in the Google Analytics UI, rather than in our code.  @kevinreiss and I looked at it today, and found that the best way to approach this ticket is:

1. Follow these steps to migrate to Google Analytics 4: https://support.google.com/analytics/answer/10759417 
1. Follow the steps here to create a new event: https://www.youtube.com/watch?v=FVzJkB4IxEQ

christinach:
@sandbergja and I migrated google analytics to GA4 and added some test events. We will revisit this next Monday.

sandbergja:
There are now three events:

- available_online_google_books (if the link URLs start with `https://www.google.com/books/`)
- available_online_resolve (if the link URLs start with `https://library.princeton.edu/resolve`)
- available_online_alma_portfolio (if the link URLs start with `https://na05.alma.exlibrisgroup.com/view/uresolver/01PRI_INST`)

PR #2354, opened 2021-02-04
Remove Matomo analytics

escowles:


christinach:
@escowles can I change this PR to a draft? Is https://github.com/pulibrary/orangelight/issues/2285 the reason we have both matomo and google analytics in orangelight? I cant remember.

christinach:
@carolyncole said matomo is not active since December. We can remove matomo and keep only google analytics.

christinach (approved):

Issue #2273, opened 2020-09-25
Replace umlaut with solr lookups


hackartisan:
Use the switch created as part of #2362

```
if use_alma 
  include new javascript stuff
else
  included umlaut javascript
end
```

Remove umlaut / get-it code for submitting the open url request and implement the following:

- Index 951 fields
  - ticketed as https://github.com/pulibrary/marc_liberation/issues/1054
  - Index into a new field (Keep indexing 856 fields as-is) because the records that continue to use 856es will only link to supplemental material and shouldn't be pulled in as online links or "other versions".
  - Records with 951 fields are electronic records
  - 951s have a lot more info than 856es, dates of coverage, etc. and you might have to construct a url.
  - Index as json?
  - index into a new core?
- On record show page, query solr index for ISBN, ISSN, and OCLC numbers to find electronic availability from indexed electronic records (records from step 1)
  - This will allow us to display the links from electronic records on the show page of the print records in the "online access" area.
  - Sometimes there’s a 776 field that has a bib id. The current “other versions” feature uses this now.
  - Usually each record will contain the eISBN and the regular ISBN in the same record. These are in 022 (ISSN) 020 (ISBN) 035 (OCL numbers)
  - Concatenate the available displays from the electronic records and put them on the page.
We’ll do this for every catalog show page.
  - Possible implementation would be to index electronic records into their own core and then do a solr join to find them.

Example records:
- https://catalog-alma-qa.princeton.edu/catalog/994264203506421 
- https://catalog-alma-qa.princeton.edu/catalog/99122306151806421 

https://github.com/pulibrary/marc_liberation/blob/6bd9ab6f3c077aa56df129e003101b28e07ff3ce/marc_to_solr/lib/princeton_marc.rb#L138 

Background:
- Initially we planned to replace umlaut with an alma service; see notes at https://docs.google.com/document/d/1YgphqGxTzhKNlV_oDUWTfuI66JY6b9LCOF9lnT8vvA4/edit# for full notes if needed.
- Meeting notes about the current proposal: https://docs.google.com/document/d/1sq_DmkRVLWEQdwf0YY1Dl3KGYlPd_0ldCYYc6PTcOuY/edit#heading=h.eyg5k66hqbqc 

umlaut documentation issue: pulibrary/marc_liberation#1034

kevinreiss:
Confirmed we can start working on this in January.

hackartisan:
comment from @christinach:

[...]

PR #2365, opened 2021-03-09
Disable umlaut integration when using Alma

eliotjordan:
Works towards #2273


christinach (approved):

spec/features/availability_spec.rb
```diff
@@ -146,4 +146,24 @@
       end
     end
   end
+
+  context 'when using Alma' do
+    before do
+      allow(Rails.configuration).to receive(:use_alma).and_return(true)
+    end
+
+    describe 'Electronic Holdings are displayed on a record page', js: true do
```
hackartisan, line 155: should these spec titles be like "don't display"? This one and line 162?

hackartisan (changes requested):
Looks good!! It will be nice to get that big block of javascript out of there.

Just a small question.

spec/features/availability_spec.rb
eliotjordan: Good catch. Fixed.

hackartisan (approved):

PR #2380, opened 2021-03-23
Suppress Marquand from Where to Find it.

kevinreiss:


app/services/stackmap_location_factory.rb
```diff
@@ -28,7 +28,7 @@ def excluded?(library)
         'Online',
         'Rare Books and Special Collections',
         'ReCAP',
-        'Lewis Library' # Temporarily inaccessible
```
carolyncole, line 31: @kevinreiss did you mean to remove Lewis?

carolyncole (approved):

app/services/stackmap_location_factory.rb
kevinreiss: Yes, their stacks are accessible again after the flood a bit ago.

PR #2391, opened 2021-04-02
Add support for Alma availability endpoints

### LINKED ISSUE #2382: Consume new bibdata availability endpoints on search results page, show page
Use the new endpoints created in https://github.com/pulibrary/bibdata/issues/1145 and https://github.com/pulibrary/bibdata/issues/1118 ([see WIP](https://github.com/pulibrary/bibdata/commit/6b131dd814c37de22aac65b98747e9160c69e9ea)) to power availability display on the search results page.

My thought is to use the `use_alma` flag to no longer include https://github.com/pulibrary/orangelight/blob/8720766bcc32fe06eee34094263d85df599fdb83/app/javascript/orangelight/availability.es6 at all, and instead include 2 new javascript files (one to handle search results, one to handle the show page). Some of the existing javascript may need to be copied over / kept, but enough of it will no longer be relevant that trying to use the `use_alma` flag in that already-complex javascript code seems unwise.

It seems likely that these will need to be approached together, at least initially. If it turns out the work can be parallelized we should split this into 2 separate tickets.

Note these endpoints might return an error if the alma concurrency limit has been reached. We need to handle that error.

eliotjordan:
Closes #2382

### DIFF
--- a/spec/features/availability_spec.rb
+++ b/spec/features/availability_spec.rb
@@ -1,5 +1,4 @@
 # frozen_string_literal: true
-
 require 'rails_helper'
 
 describe 'Availability' do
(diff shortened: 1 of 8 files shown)


spec/features/availability_spec.rb
```diff
@@ -18,11 +17,63 @@
...
+
+    describe 'viewing a record for an electronic holding', js: true do
+      before do
+        stub_holding_locations
+        visit '/catalog/857469'
+      end
+
+      xit 'does not render the online section' do
+        expect(page).not_to have_selector('.availability--online')
+      end
```
christinach, line 55: I think this should change to `expect(page).to have_selector('.availability--online')` since there is an online section for this record through google books.

spec/features/availability_spec.rb
```diff
@@ -37,21 +88,8 @@
     end
```
christinach, line 88: Can you update test `An item reserved for controlled digital lending' ` with a record that is on CDL and not deactivate it?  ?

spec/features/availability_spec.rb
```diff
@@ -65,84 +103,52 @@
...
+        visit 'catalog/2238036'
+        expect(page).to have_selector '.availability-icon.badge.badge-success', text: 'See front desk', count: 1
       end
+    end
 
-      xit 'does not render the location' do
-        expect(page).not_to have_selector('.availability--holding')
+    # the item used here is no longer checked out
+    xdescribe 'Checked out item', js: true do
+      it 'shows due date', unless: in_ci? do
```
christinach, line 148: can you update this with an item that is checked out?

christinach (changes requested):

spec/features/availability_spec.rb
eliotjordan: These tests are skipped because of the comment on line 30. Our PR did not break this. We are not sure why this context is testing 'bots' with an iphone driver. Created a follow up ticket: https://github.com/pulibrary/orangelight/issues/2396

spec/features/availability_spec.rb
eliotjordan: We're skipping these because they hit live data that keeps changing. These are not tested in CI and will all eventually fail. They are not failing as a result of our PR. The idea is that we can eventually move the tests for this functionality over to Jest as we build out more of the Alma functionality.

spec/features/availability_spec.rb
eliotjordan: Same comment as above. We're skipping these because they hit live data that keeps changing. Also, IIRC, we will not be able to get access to checked-out status in Alma. The availability will simply be "Unavailable", so this is a "voyager" spec that will not migrate.

christinach:
Yes I understand that it tests live data but I think before we stop testing them we need to make sure they are tested somewhere? I'm talking about the voyager specs.

christinach (approved):
We will have to address https://github.com/pulibrary/orangelight/issues/2398, https://github.com/pulibrary/orangelight/issues/2397, https://github.com/pulibrary/orangelight/issues/2396

PR #2392, opened 2021-04-02
Display AVA/AVE fields in staff view.

tpendragon:
Closes https://github.com/pulibrary/bibdata/issues/712

### DIFF
--- a/spec/views/catalog/librarian_view.html.erb_spec.rb
+++ b/spec/views/catalog/librarian_view.html.erb_spec.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe "catalog/librarian_view.html.erb" do
+  context "when given an Alma marc source with AVA fields" do
+    it "displays the AVA fields" do
+      allow(view).to receive(:params).and_return(id: "9922486553506421")
+      stub_request(:get, "https://bibdata-staging.princeton.edu/bibliographic/9922486553506421")
+        .to_return(status: 200, body: File.open(Rails.root.join("spec", "fixtures", "alma", "9922486553506421_marc.xml")), headers: {})
+
+      render
+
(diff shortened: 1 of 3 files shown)


spec/views/catalog/librarian_view.html.erb_spec.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe "catalog/librarian_view.html.erb" do
+  context "when given an Alma marc source with AVA fields" do
+    it "displays the AVA fields" do
+      allow(view).to receive(:params).and_return(id: "9922486553506421")
+      stub_request(:get, "https://bibdata-staging.princeton.edu/bibliographic/9922486553506421")
+        .to_return(status: 200, body: File.open(Rails.root.join("spec", "fixtures", "alma", "9922486553506421_marc.xml")), headers: {})
```
christinach, line 10: @tpendragon can you change the stub request to `https://bibdata-alma-staging.princeton.edu/bibliographic/9922486553506421` ?

christinach (changes requested):

hackartisan (changes requested):

spec/views/catalog/librarian_view.html.erb_spec.rb
tpendragon: I'd be happy to. Can you show me an example in the code where this happens? I think right now for me to change this I'd have to change every reference that has a BIBDATA_BASE, because that's where this URL comes from, but if that's not the case I just need some guidance. Thanks!

christinach: oh, ok sorry I thought we had set up a bibdata_alma_base for bibdata-alma-staging. This will probably be a different issue.

christinach (approved):

hackartisan (approved):
thank you!

PR #2393, opened 2021-04-02
Display electronic record "Online" badge in search results

### LINKED ISSUE #2388: Electronic records need "online" badge and link in search results
These records have portfolios but no holdings. See screenshots / details in comments below.

tpendragon:
Closes #2388

Requires https://github.com/pulibrary/pul_solr/pull/268

Screen Shot 2021-04-02 at 3 25 14 PM

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -46,6 +46,18 @@ module ApplicationHelper
     urls
   end
 
+  # Returns electronic portfolio links for Alma records.
+  # @param document [SolrDocument]
+  # @return [Array<String>] array containing the links
+  def electronic_portfolio_links(document)
+    return [] if document.electronic_portfolios.blank?
+    document.electronic_portfolios.map do |portfolio|
+      content_tag(:div, class: 'library-location') do
+        link_to(portfolio["title"], portfolio["url"], target: '_blank')
+      end
+    end
+  end
+
   # Retrieve a URL for a stack map location URL given a record, a call number, and the library in which it is held
   # @param location [Hash] location information for the item holding
   # @param document [SolrDocument] the Solr Document for the record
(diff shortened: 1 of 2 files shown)


app/helpers/application_helper.rb
```diff
@@ -179,8 +191,18 @@ def holding_location(holding)
   # @return [String] the markup
   def holding_block_search(document)
     block = ''
-    links = search_links(document['electronic_access_1display'])
-    holdings_hash = JSON.parse(document['holdings_1display'] || '{}')
+    portfolio_links = electronic_portfolio_links(document)
+    # Create stub holdings for Alma's electronic profiles so the below logic will work.
+    # Alma does not have an 'Online' library for holdings, so we fake it to keep
+    # the old logic around.
+    # @TODO: Remove when Voyager is gone.
```
hackartisan, line 198: Which part should be removed when voyager is gone?

app/helpers/application_helper.rb
```diff
@@ -179,8 +191,18 @@ def holding_location(holding)
...
+    portfolio_holdings = Hash[
+      portfolio_links.map do |_portfolio|
+        [SecureRandom.uuid, { 'library' => 'Online' }]
```
hackartisan, line 201: This looks like it's used as the holding id. A big meaningless random number seems like it could be confusing some day. I could imagine debugging something in web tools and trying to track down the target of this id. Maybe we should add the portfolio holding id here https://github.com/pulibrary/bibdata/blob/91666d6c17eb1cb0823d57d5e32c3e986ab92816/marc_to_solr/lib/electronic_portfolio_builder.rb#L24

hackartisan (changes requested):

hackartisan:
Why does it require https://github.com/pulibrary/pul_solr/pull/268? something similar was happening here and I'd love to understand. https://github.com/pulibrary/orangelight/pull/2370/files#diff-f3be7114593cea59759b4f73c1fe286d347833930e56edf12b4561ff96797befL19

tpendragon:
@hackartisan It requires https://github.com/pulibrary/pul_solr/pull/268 because without `electronic_portfolio_s` in the `fl` it doesn't come back in the solr response for search results, so we can't parse it.

app/helpers/application_helper.rb
tpendragon: I'm happy to do that, but it would mean this code won't work until we perform a full reindex, and I think we'll likely remove the requirement of having a holding ID in the future. Is that still what you'd like me to do?

app/helpers/application_helper.rb
tpendragon: I expanded the comment.

hackartisan (approved):
👍🏻 thanks @tpendragon

hackartisan:
oh, @tpendragon, one last problem is that the test is failing in CI.

tpendragon:
@hackartisan It'll fail CI until the pul_solr PR is merged.

hackartisan:
oh, yeah. 👍🏻 it's merged

PR #2399, opened 2021-04-06
Update alma base to point at the alma catalog

carolyncole:
I noticed once configured correctly my local requests was working better with alma than the on out on QA.  After looking at the difference I realized requests was still talking to the non alma catalog for solr documents.

Screen Shot 2021-04-06 at 2 58 23 PM


kevinreiss (approved):
Good catch @carolyncole

PR #2404, opened 2021-04-07
Redirect to Alma ID if it exists.

### LINKED ISSUE #2400: Redirect links with voyager IDs to their equivalent alma id endpoint
@christinach commented on [Wed Jan 20 2021](https://github.com/pulibrary/bibdata/issues/992)

Identify links pointing to existing source systems () that are migrated, so that they can be removed or updated, if necessary.  
Once a resource is migrated to Alma, the links should be updated to point to Alma. 
(will add more information)

---

@mzelesky commented on [Thu Jan 21 2021](https://github.com/pulibrary/bibdata/issues/992#issuecomment-764689328)

Essentially, it is unknown how many external sites (like professor's blogs, LibGuides, etc.) use direct links to Blacklight bib records, so there needs to be a redirect in place to take a user from a Voyager bib ID to the new Alma MMS ID.

---

@hackartisan commented on [Thu Jan 21 2021](https://github.com/pulibrary/bibdata/issues/992#issuecomment-764908861)

Do we want to inspect the ID? or do we want to just try a rewrite if the page isn't found?

A voyager bib id will always be numeric. There is no maximum length, but we could get the last id used in voyager (13million would be the highest). The minimum length of an alma id would be the prefix and suffix with 1 in between.

---

@tpendragon commented on [Tue Apr 06 2021](https://github.com/pulibrary/bibdata/issues/992#issuecomment-814525957)

@hackartisan Is this meant to be an Orangelight ticket?

---

@hackartisan commented on [Wed Apr 07 2021](https://github.com/pulibrary/bibdata/issues/992#issuecomment-814889366)

@tpendragon it must be; I can't think of how it would make sense in the bibdata context, and I don't see a different one for orangelight. I'll move it. Thanks!

tpendragon:
Closes #2400

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -14,6 +14,10 @@ class CatalogController < ApplicationController
   before_action :redirect_browse
 
   rescue_from Blacklight::Exceptions::RecordNotFound do
+    alma_id = "99#{params[:id]}3506421"
+    search_service.fetch(alma_id)
+    redirect_to solr_document_path(id: alma_id)
+  rescue Blacklight::Exceptions::RecordNotFound
     redirect_to '/404'
   end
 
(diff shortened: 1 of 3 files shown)


hackartisan (approved):
🎉

PR #2405, opened 2021-04-07
Add dynamic environment-based queues

### LINKED ISSUE #2319: Create a new queue in rabbitmq for catalog-alma-qa
queue name: catalog_alma_qa, catalog_alma_qa-error, catalog_alma_qa-retry.
The consumers in the queue would be lib-orange-alma-qa1 and lib-orange-alma-qa2.

tpendragon:
Closes #2319 

This also adds a connection pool fix we had to add in DPUL to prevent the workers from stopping up.

### DIFF
--- a/app/workers/event_handler.rb
+++ b/app/workers/event_handler.rb
@@ -2,7 +2,11 @@
 
 class EventHandler
   include Sneakers::Worker
-  from_queue :"catalog_#{Rails.env}"
+  if Rails.configuration.use_alma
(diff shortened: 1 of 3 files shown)


eliotjordan (approved):

hackartisan (changes requested):
I think we should think more about this with respect to the time when we're running alma and voyager machines at the same time.

eliotjordan (approved):

hackartisan (approved):

PR #2444, opened 2021-05-11
Use alma library card account app when use_alma flag is true

### LINKED ISSUE #2401: "Your account" should link to ex libris library card app
Make this change behind the `use_alma` flag.

The link should open in a new window / tab since we can't provide navigation back to the catalog.

library card app is at: https://princeton.alma.exlibrisgroup.com/discovery/account?vid=01PRI_INST:Services&lang=EN&section=overview

hackartisan:
closes #2401 

Screen Shot 2021-05-11 at 4 22 42 PM

Screen Shot 2021-05-11 at 4 22 48 PM

Screen Shot 2021-05-11 at 4 23 09 PM

Screen Shot 2021-05-11 at 4 33 16 PM

Screen Shot 2021-05-11 at 4 22 28 PM


christinach:
Some thoughts:
1. Dropdown with one option: Your account -> `Library Card login` ? (redirects to the exlibris library card view)
Keep `Bookmarks` and `Search history` with the login as it is in this pr.

2. Dropdown with two options: Your account -> `Library Card login` and -> `Catalog Login` (This way we dont need to have separate login for bookmarks and search history).
             
What do you think? I understand the difficulty differentiating between the exlibris view and the catalog. 

Also can we add an external link icon to the link that takes the user to the exlibris library card?

christinach (changes requested):
Can we add an external link icon to the link that takes the user to the exlibris library card?

christinach:
@axamei if/when you have a chance can you provide some feedback?

hackartisan:
I kept the title "Your Account" since that's what was used before for the in-site page and it's the same sort of data. The new page is titled "My Library Card" but I'm not sure if people will know what that means in the context of this UI?

I think adding login buttons to the bookmarks and search history pages gives a little hint that there will be saved data there if they want to log in. If this isn't enough of a hint then maybe a full line of text below the page heading like `<a>Log in</a> to save or view saved bookmarks` would be better. 

If we moved it back into the "your account" dropdown I'm not really sure where it should redirect -- I guess just always to the page you were on? I'm not sure it's clear to the user what is the utility of the catalog login.

axamei:
The double "Your account" labels might be confusing. And, the "My Library Card" page does not give them access to their catalog account to view saved bookmarks and search history. Maybe in the submenu we can have the following options? 
Screen Shot 2021-05-12 at 11 20 46 AM

And asking the patron to click 2 login buttons seems unnecessary. Maybe we can add the 2 login options directly in the bookmarks and search history pages? 
Screen Shot 2021-05-12 at 11 30 16 AM
(fyi did this quickly in devtools to illustrate)

axamei (dismissed):
Commented in PR before requesting changes...

hackartisan:
@axamei Thanks for these suggestions. So the bookmark / search history links under "your account" would just go to the page, without logging in first? And would the links to those pages then be removed from the subnav?

The site does save bookmarks and search history in session even if you're not logged in and even if you don't have an account. I wonder if that's one reason bookmarks was listed separately to begin with. But I'm not sure this distinction is important; I think they're okay under the "Your account" menu.

christinach:
@hackartisan the bookmark / search history links under "your account" would go to the page like in the screenshot. The links in the info text are the ones that will direct the user to login when they click on them. https://catalog.princeton.edu/users/sign_in

axamei:
@hackartisan Yes, they would like to the bookmark/search history pages without having to login first. We could remove the redundant links in the search header. These are usually decisions I make after looking at analytics. Initially, we can leave the redundant links in, add some event tracking, and then evaluate the removal of the subheader links.

hackartisan:
Prior to this PR, "Bookmarks" was in the subnav and "Search History" was in the account dropdown. In this PR I moved "Search History" out to the subnav for symmetry with "Bookmarks" But it sounds like the Information Architecture makes more sense for them to both be in the account dropdown. I'd rather not add Search History to the subnav only to potentially remove it again down the line. In which case Bookmarks would be in both places, but Search History only in the account dropdown. I'm thinking we should just go ahead and remove Bookmarks from the subnav to have that symmetry.

christinach:
I think we should keep bookmarks in the subnav. I believe users are used to see the count in the parenthesis when they bookmark a record. If we remove it from the subnav then they would have to check all the time in the dropdown for the count.

hackartisan:
@christinach @axamei I made these changes, except embedding the login options was bit more complex than I have time to do right now. I agree it's a good idea but with the barcode form there's some controller stuff I just don't have time to figure out, so there are still two clicks needed. I will create a follow-up ticket for that idea. 

This is ready for re-review and I've deployed to orangelight staging and to orangelight-alma-qa if you could take another look.

christinach (approved):
ok, we can merge this PR and follow up on the `login` with the new ticket. https://github.com/pulibrary/orangelight/issues/2446

christinach:
@hackartisan I also opened https://github.com/pulibrary/orangelight/issues/2447 to address the external icon.

hackartisan:
@christinach thanks

Issue #2445, opened 2021-05-12
Read ENV in configuration, use config values in code


hackartisan:
ENV is used directly in a lot of orangelight code. It would be better to centralize its use in configuration and then use config values in the code itself.

- [x] Always use `Requests.config["bibdata_base"]` instead of `ENV["bibdata_base"]`

Related: #1079

PR #2454, opened 2021-05-17
Replace ENV['bibdata_base'] with Requests.config['bibdata_base']

tpendragon:
Progresses #2445


spec/spec_helper.rb
```diff
@@ -18,6 +18,7 @@
 
 RSpec.configure do |config|
   config.include FactoryBot::Syntax::Methods
+  config.example_status_persistence_file_path = "tmp/rspec_examples.txt"
```
eliotjordan, line 21: This looks like it will be useful. 👍

eliotjordan (approved):
Nice job! 💯

PR #2465, opened 2021-05-20
Availability for bib records with temporary locations

hectorcorrea:
Adds a check to make a second availability call to bib data when the initial call reports temporary locations. This is required because Alma does not provide holding information when items are in temporary locations. 

The new code looks to see if Alma reported temporary locations, if it did, then it makes a second availability call to bib data but this time explicitly request to fetch the holding information (via the deep=true query string parameter). This second call is slow and therefore we only make it when we detect temporary locations.

When there are no temporary locations in the initial call to bib data we use the data that we got in the first call. 

This is a good record to test since it has temporary locations: `9959958323506421`

### DIFF
--- a/app/javascript/orangelight/availability.es6
+++ b/app/javascript/orangelight/availability.es6
@@ -24,6 +24,7 @@ export default class AvailabilityUpdater {
     this.process_results_list = this.process_results_list.bind(this);
     this.process_barcodes = this.process_barcodes.bind(this);
     this.process_single = this.process_single.bind(this);
+    this.update_single = this.update_single.bind(this);
     this.process_scsb_single = this.process_scsb_single.bind(this);
   }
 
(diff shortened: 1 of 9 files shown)


app/javascript/orangelight/availability.es6
```diff
@@ -115,6 +116,32 @@ export default class AvailabilityUpdater {
...
+      const availability_info = holding_records[this.id][holding_id];
+      if (availability_info['temp_location'] === true) {
+        dataComplete = false; // The data that we get from Alma for temporary locations is incomplete.
+        break;
+      }
+    }
+
+    if (dataComplete) {
+      // Update the page with the data that we already have.
+      this.update_single(holding_records);
```
hectorcorrea, line 130: Uses data from the first call to bib data. This is good as-is.

app/javascript/orangelight/availability.es6
```diff
@@ -115,6 +116,32 @@ export default class AvailabilityUpdater {
...
+      return;
+    }
+
+    // Make a separate call (with deep=true) to get more information before updating the page.
+    var url = `${this.bibdata_base_url}/bibliographic/${this.id}/availability.json?deep=true`;
+    $.getJSON(url, this.update_single)
```
hectorcorrea, line 136: Makes second call to bib data and uses the result of this call to update the page.

hectorcorrea: The incomplete data looks like this (notice the `fake_id`):

```
{
    "9959958323506421": {
        "fake_id_1": {
            "on_reserve": "N",
            "location": "lewis$resterm",
            "label": "Lewis Library: Term Loan Reserves",
            "status_label": "Available",
            "copy_number": null,
            "cdl": false,
            "temp_location": true,
            "id": "fake_id_1"
        }
    }
}
```

The second call will give the return the proper data (notice the correct holding_ids):
```
{
    "9959958323506421": {
        "22272063570006421": {
            "on_reserve": "N",
            "location": "lewis$resterm",
            "label": "sciresp: Lewis: Term Loan",
            "status_label": "Available",
            "copy_number": "1",
            "cdl": false,
            "temp_location": true,
            "id": "22272063570006421"
        },
        "22272063520006421": {
            "on_reserve": "N",
            "location": "lewis$resterm",
            "label": "sciresp: Lewis: Term Loan",
            "status_label": "Available",
            "copy_number": "2",
            "cdl": false,
            "temp_location": true,
            "id": "22272063520006421"
        }
    }
}
```

hackartisan (changes requested):
It would be great to get a test for this in spec/javascript/orangelight/availability.spec.js

hectorcorrea:
Added unit test.

hackartisan (approved):

PR #2469, opened 2021-05-21
Alma branch

carolyncole:
This pulls the running alma branch into oranglight.  I evaluated having a switch like orangelight, but there are so many changes required it does not seem feasible (more work) to put a switch around them all.  https://github.com/pulibrary/requests/pull/922


hackartisan:
👍🏻 @carolyncole The plan we thought we would try was to have this alma branch here on orangelight to use for deploy to the alma / orangelight servers. In general we plan to continue submitting alma PRs to `main`. The catch is that to deploy we'll have to rebase this branch on main and then deploy this branch. If that becomes too onerous we can just switch to using the branch on orangelight, as well.

christinach (approved):

PR #2475, opened 2021-05-26
Display availability as Online in the search results and Unavailable in the record page for CDL records

hectorcorrea:
First pass at addressing display of status for CDL items (https://github.com/pulibrary/bibdata/issues/1291). With this change instead of displaying an off-putting "Unavailable" status for the records available via CDL we display "Online" as the status. Notice that new logic only kicks in if the Status is Unavailable and the item is on CDL.

Below is a screenshot of how a record like this would be displayed in the Show page

show_page

The Online status has a tooltip (not shown) that says "Available online. Physical copy might not be available."

### This change also handles the display on the Search results. 
search_page

### A note on facet counts
Notice that the "Access" facet counts might not be accurate when these CDL records are on the search results since we calculating this new "Online" status on the fly. This might not be obvious when there are many records in the search results (like on the above screenshot) but it is obvious when this record is the only one on the result set

search_page_single

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -248,8 +248,11 @@ module ApplicationHelper
         info << content_tag(:div, search_location_display(holding, document), class: 'library-location', data: { location: true, record_id: document['id'], holding_id: id })
       end
       block << content_tag(:li, info.html_safe, data: { availability_record: check_availability, record_id: document['id'], holding_id: id, aeon: aeon_location?(location) })
-      cdl_placeholder = content_tag(:span, '', class: 'badge badge-primary', 'data-availability-cdl' => true)
-      block << content_tag(:li, cdl_placeholder.html_safe)
+
+      if Rails.configuration.use_alma
+        cdl_placeholder = content_tag(:span, '', class: 'badge badge-primary', 'data-availability-cdl' => true)
+        block << content_tag(:li, cdl_placeholder.html_safe)
+      end
     end
 
     if scsb_multiple == true
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -247,8 +247,9 @@ module ApplicationHelper
         end
         info << content_tag(:div, search_location_display(holding, document), class: 'library-location', data: { location: true, record_id: document['id'], holding_id: id })
       end
-      info << content_tag(:span, "Online", class: 'badge badge-primary hidden', 'data-availability-cdl' => true)
       block << content_tag(:li, info.html_safe, data: { availability_record: check_availability, record_id: document['id'], holding_id: id, aeon: aeon_location?(location) })
+      cdl_placeholder = content_tag(:span, '', class: 'badge badge-primary', 'data-availability-cdl' => true)
+      block << content_tag(:li, cdl_placeholder.html_safe)
     end
 
     if scsb_multiple == true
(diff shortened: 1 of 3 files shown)


hackartisan (changes requested):
This looks good. Can we add a test?

Also it looks like in production as use a blue badge; `badge-primary`, for this status so we should use that.

christinach (commented):
Can the status be: 'Online. Physical copy might not be available' and not only in the tooltip? Or have we decided that we don't want a more descriptive status for CDL? I'm only saying that because there is some confusion by having online and the physical location display next to the status.
Maybe @kevinreiss and @mzelesky can also give some feedback?

christinach (changes requested):
I added a comment but forgot to request changes at least until we decide that the display is not as confusing.

hectorcorrea:
I have made the changes requested in today's meeting. Below are two screenshots showing how the display renders for the Show page and the Search Results pages

show_page

search_results

If this is the correct approach I'll go ahead and add a test for this.

christinach:
The updated PR follows what we said in the meeting but we are missing a test.

hackartisan (changes requested):
I think a little bit more space between the 2 badges would be good. Here is a search result with multiple rows that shows how it looks in production: https://catalog.princeton.edu/catalog?utf8=%E2%9C%93&search_field=all_fields&q=the+age+of+surveillance+capitalism

christinach (changes requested):
I forgot to request changes

hectorcorrea:
Fixed the spacing issue that @hackartisan noted. This took me a bit longer that I anticipated since I spent time making sure the new HTML elements match the original structure of the page, in particular I am now using an HTML `li + span` rather than just a `span`. 

search_results

Tests coming up soon :)

hectorcorrea:
@hackartisan @christinach I've added a test.

hackartisan (approved):

app/helpers/application_helper.rb
```diff
@@ -248,6 +248,8 @@ def holding_block_search(document)
         info << content_tag(:div, search_location_display(holding, document), class: 'library-location', data: { location: true, record_id: document['id'], holding_id: id })
       end
       block << content_tag(:li, info.html_safe, data: { availability_record: check_availability, record_id: document['id'], holding_id: id, aeon: aeon_location?(location) })
+      cdl_placeholder = content_tag(:span, '', class: 'badge badge-primary', 'data-availability-cdl' => true)
+      block << content_tag(:li, cdl_placeholder.html_safe)
```
christinach, line 252: @hectorcorrea do you think we should add here the use_alma flag `if Rails.configuration.use_alma`? If we don't add it then we have in production an unused cdl element `<span class="badge badge-primary" data-availability-cdl="true"></span>`. It doesn't break anything but it is an unused span. What do you think?

christinach (changes requested):

app/helpers/application_helper.rb
hectorcorrea: yes, good idea

app/helpers/application_helper.rb
```diff
@@ -248,6 +248,11 @@ def holding_block_search(document)
         info << content_tag(:div, search_location_display(holding, document), class: 'library-location', data: { location: true, record_id: document['id'], holding_id: id })
       end
       block << content_tag(:li, info.html_safe, data: { availability_record: check_availability, record_id: document['id'], holding_id: id, aeon: aeon_location?(location) })
+
+      if Rails.configuration.use_alma
```
hectorcorrea, line 252: @christinach I added the check.

christinach (approved):
Thank you @hectorcorrea !

PR #2476, opened 2021-05-27
Update location display on the record page to be consistent between the indexed location and the one that comes from the js call

### LINKED ISSUE #2467: Update record page location display
Update location display in the record page to be consistent before and after js runs. 
follow up on https://github.com/pulibrary/orangelight/pull/2466

christinach:
closes  #2467


christinach:
I'll change this to draft I want to add one more test.

hackartisan (approved):

PR #2483, opened 2021-05-28
Update for alma with new numismatics location

### LINKED ISSUE #2481: Add new numismatics location when USE_ALMA=true
New location is `rare$num`

christinach:
closes #2481

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -223,7 +223,7 @@ module ApplicationHelper
           info << links.shift.html_safe
         end
       else
-        if holding['dspace'] || holding['location_code'] == 'num'
+        if holding['dspace'] || holding['location_code'] == (Rails.configuration.use_alma ? 'rare$num' : 'num')
           check_availability = false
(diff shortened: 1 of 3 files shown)


hackartisan (changes requested):
I wonder if we can give ourselves more flexibility by allowing either `num` or `rare$num` during the transition, since we aren't sure exactly when we're going to switch figgy. What do you think?

christinach:
@hackartisan are you asking me to keep on or the other?

christinach:
I noticed that my test does not include the USE_ALMA maybe that created a confusion.

christinach:
From slack @hackartisan explained to me that she meant to keep all the options for numismatics for more flexibility; I explained that the orangelight/alma branch includes the request version that is looking for rare$num; this will only work when it gets merged and we are in the alma environment. So it wouldn't work to index numismatics earlier with `rare$num` if we are not in alma mode.

hackartisan:
I'm curious about indexing `num` in alma during QA, not indexing `rare$num` before alma. because of the timing of the required figgy change. but I won't hold up this PR over it; it sounds like if we wanted to change it it would have to be done here and in requests.

hackartisan (approved):

Issue #1336, opened 2018-08-22
Research how to trigger invalid "back to search" link


tpendragon:
## Backtrace

    line 39 of [PROJECT_ROOT]/config/initializers/blacklight_url_helper_behavior.rb: link_back_to_catalog
    line 9 of [PROJECT_ROOT]/app/views/catalog/_previous_next_doc.html.erb: _app_views_catalog__previous_next_doc_html_erb___141562302411537421_70097437451980
    line 1 of [PROJECT_ROOT]/app/views/catalog/show.html.erb: _app_views_catalog_show_html_erb__621661626053428788_36727780

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/54399/faults/36655673)

escowles:
Users reported leaving results open for more than a day — maybe session timeouts are the core issue here?

christinach:
check #1458 #1455

escowles:
Anya Bartelmann just reported seeing this issue today: https://app.honeybadger.io/projects/54399/faults/36655673

The error looks like it's happening when trying to calculate the link back to the search results:
https://github.com/pulibrary/orangelight/blob/master/config/initializers/blacklight_url_helper_behavior.rb#L39

We should probably just rescue StandardError to prevent this.

christinach:
I don't think we should rescue StandardError without checking first #1458 and #1455.

christinach:
closing in favor of https://github.com/pulibrary/orangelight/issues/4011

PR #2496, opened 2021-06-09
Prevents error when trying to render the back to catalog link

hectorcorrea:
Prevents error when trying to render the back to catalog link and the user's session has information that results in an invalid link. This fix should probably be pushed eventually to Blacklight but for now a local fix would do.

Fixes https://github.com/pulibrary/orangelight/issues/1336

See also https://github.com/pulibrary/pulmap/pull/950

### DIFF
--- a/spec/helpers/blacklight_helper_spec.rb
+++ b/spec/helpers/blacklight_helper_spec.rb
@@ -5,8 +5,30 @@ require 'rails_helper'
 describe BlacklightHelper do
   let(:blacklight_params) { {} }
 
+  class FakeSearchState
+    def reset(search_session)
+      search_session
+    end
+  end
+
+  class FakeValidCurrentSearchSession
+    def query_params
+      { action: "show", controller: "catalog", id: "123" }
+    end
+  end
+
+  class FakeInvalidCurrentSearchSession
+    def query_params
+      { action: "index", controller: "advanced", id: "123" }
+    end
+  end
+
   before do
     allow(self).to receive(:blacklight_params).and_return(blacklight_params)
+    allow(self).to receive(:current_or_guest_user).and_return(User.new)
+    allow(self).to receive(:search_action_path) do |*args|
+      search_catalog_url(*args)
+    end
   end
 
   describe '#left_anchor_escape_whitespace' do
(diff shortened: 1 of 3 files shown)


spec/helpers/blacklight_helper_spec.rb
```diff
@@ -5,8 +5,30 @@
 describe BlacklightHelper do
   let(:blacklight_params) { {} }
 
+  class FakeSearchState
```
hectorcorrea, line 8: There is probably a way to stub this via FactoryWhatever. Let me know if you know how.

hackartisan:
Oh I thought this was https://github.com/pulibrary/pulmap/pull/950 -- was wondering if @eliotjordan would want to review that one. I'll go ahead and merge this.

hackartisan:
Oh I thought this was https://github.com/pulibrary/pulmap/pull/950 -- was wondering if @eliotjordan would want to review that one. I'll go ahead and merge this.

spec/helpers/blacklight_helper_spec.rb
```diff
@@ -128,4 +150,24 @@
       expect(helper).to have_received(:head).with(:bad_request)
     end
   end
+
+  describe "#link_back_to_catalog_safe" do
+    context "with valid parameters" do
+      let(:search_state) { FakeSearchState.new }
+      let(:current_search_session) { FakeValidCurrentSearchSession.new }
```
hackartisan, line 157: Looking at this closer with Hector because of his question about these mocks. The mocks themselves seem fine, although there is likely some more elegant way of doing them, but these let statements are really confusing me a lot. They seem to be overriding methods in the actual helper namespace -- they're not used anywhere in the tests themselves. I don't really understand how this is the case and it makes me pretty uncomfortable.

I think we could try something like `allow(Blacklight::UrlHelper).to receive(:find_search_session).and_return(valid_session)`

hackartisan (changes requested):

spec/helpers/blacklight_helper_spec.rb
```diff
@@ -128,4 +132,32 @@
       expect(helper).to have_received(:head).with(:bad_request)
     end
   end
+
+  describe "#link_back_to_catalog_safe" do
+    let(:blacklight_config) { CatalogController.new.blacklight_config }
```
hectorcorrea, line 137: @hackartisan I am now using Blacklight built-in classes for `blacklight_config` and `search_state`, this should make the test easier to follow.

spec/helpers/blacklight_helper_spec.rb
```diff
@@ -128,4 +132,32 @@
...
+    let(:search_session) { {} }
+
+    context "with valid parameters" do
+      let(:valid_session) { instance_double(Search) }
+      it "produces a link" do
+        allow(valid_session).to receive(:query_params).and_return(
+          action: "show", controller: "catalog", id: "123"
+        )
+        allow(self).to receive(:current_search_session).and_return(valid_session)
+        expect(link_back_to_catalog_safe).to include("/catalog/123")
```
hackartisan, line 148: Per discussion, we want to try `expect(helper.link_back_to_catalog_safe)` and setting the allows on the `helper` instead of on `self`.

Note also that this helper initialization may obviate a little of the test setup: https://github.com/pulibrary/orangelight/blob/ace3e1ba6188b85a091eac310adb82b18ac1cfe2/spec/support/controller_level_helpers.rb#L16

hackartisan (changes requested):

hackartisan (approved):

Issue #2413, opened 2021-04-16
Handle 429 from bibdata availability calls on show page and search results page


hackartisan:
Bibdata will return a 429 if it gets one from alma (too many requests). Catch the error and retry. This is a highly visible feature and it should take precedence over other uses of the Alma API. It's better for it to take a long time than not show up.

PR #2497, opened 2021-06-09
Retry call to bib data availability endpoint in the show page

hectorcorrea:
Retries call to bib data availability endpoint in the show page when the error returned was an HTTP 429 (too many calls per second). Notice that we wait a second and a half before retrying.

Part of #2413 

Below is a screenshot of how this will look for a user in the event that even the retry failed to fetch the availability. The tooltip (not shown) will say "Cannot determine real-time availability for item at this time."

undetermined

### DIFF
--- a/app/javascript/orangelight/availability.es6
+++ b/app/javascript/orangelight/availability.es6
@@ -25,10 +25,11 @@ export default class AvailabilityUpdater {
     this.process_barcodes = this.process_barcodes.bind(this);
     this.process_single = this.process_single.bind(this);
     this.update_single = this.update_single.bind(this);
+    this.update_single_undetermined = this.update_single_undetermined.bind(this);
     this.process_scsb_single = this.process_scsb_single.bind(this);
   }
 
-  request_availability() {
+  request_availability(allowRetry) {
     let url;
     // a search results page or a call number browse page
     if ($(".documents-list").length > 0) {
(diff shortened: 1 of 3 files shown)


app/javascript/orangelight/availability.es6
```diff
@@ -25,10 +25,11 @@ export default class AvailabilityUpdater {
     this.process_barcodes = this.process_barcodes.bind(this);
     this.process_single = this.process_single.bind(this);
     this.update_single = this.update_single.bind(this);
+    this.update_single_undetermined = this.update_single_undetermined.bind(this);
     this.process_scsb_single = this.process_scsb_single.bind(this);
   }
 
-  request_availability() {
+  request_availability(allowRetry) {
```
hectorcorrea, line 32: In the future, we could make this a number of retries (rather than a boolean) but for now a single retry should be OK.

spec/javascript/orangelight/availability.spec.js
```diff
@@ -151,6 +151,23 @@ describe('AvailabilityUpdater', function() {
     spy.mockRestore()
   })
 
+  // Make sure that the code to handle undetermined availability status updates
```
hectorcorrea, line 154: Notice that this test is pretty shallow, it only makes sure the HTML was updated when we directly call the `update_single_undetermined` method. 

To test the logic that makes that call in the code, we would need to mock the an Ajax call that returns an HTTP 429 on the first try and an HTTP 200 on the second try. That seems more complicated than I think we want to pursue.

app/javascript/orangelight/availability.es6
```diff
@@ -54,6 +55,18 @@ export default class AvailabilityUpdater {
...
         return $.getJSON(url, this.process_single)
           .fail((jqXHR, textStatus, errorThrown) => {
+            if (jqXHR.status == 429) {
+              if (allowRetry) {
+                console.log(`Retrying availability for record ${this.id}`);
+                window.setTimeout(() => {
+                  this.request_availability(false);
+                }, 1500);
+              } else {
+                console.error(`Failed to retrieve availability data for the bib (retry). Record ${this.id}: ${errorThrown}`);
```
jrgriffiniii, line 65: 👍

christinach (approved):

spec/javascript/orangelight/availability.spec.js
christinach: @hectorcorrea do you want to open a new issue for that ^ ?

jrgriffiniii (approved):

spec/javascript/orangelight/availability.spec.js
hectorcorrea: @christinach done: https://github.com/pulibrary/orangelight/issues/2498

christinach: Thank you @hectorcorrea

PR #2501, opened 2021-06-10
Add read-only mode

### LINKED ISSUE #2439: Implement a read-only switch
We need a read-only switch for orangelight that will turn off the ability to request items and other account interactions (e.g. renewals)

We'll have to shut these features down on July 24th for the circulation freeze.

What should the UI look like in read-only mode?
* "request button" is still visible but looks "inactive" and can't be clicked
* Use a banner message to say why requests are deactivated
* Make "my account" route redirect to `/` with a flash message saying that the page is disabled during migration

eliotjordan:
Closes #2439 

In read-only mode:

- Hides the request button. This is the simplest method to implement and I think is the least confusing for the user.
- Adds a special announcement; configurable using and env var.
- Redirects the account page to the root page and flashes a notification.

Screen Shot 2021-06-10 at 4 43 26 PM

Screen Shot 2021-06-10 at 4 43 38 PM

### DIFF
--- a/app/services/physical_holdings_markup_builder.rb
+++ b/app/services/physical_holdings_markup_builder.rb
@@ -237,8 +237,15 @@ class PhysicalHoldingsMarkupBuilder < HoldingRequestsBuilder
   end
 
   def self.requestable_location?(location, adapter, holding)
-    return false if location.nil? || adapter.unavailable_holding?(holding)
-    location[:requestable]
+    if location.nil?
+      false
+    elsif adapter.unavailable_holding?(holding)
+      false
+    elsif Orangelight.read_only_mode
+      false
+    else
+      location[:requestable]
+    end
   end
 
   def self.aeon_location?(location)
--- a/config/initializers/read_only_mode.rb
+++ b/config/initializers/read_only_mode.rb
@@ -6,7 +6,7 @@ module Orangelight
   end
 
   def read_only_message
-    default_msg = "Orangelight is in read-only mode."
+    default_msg = "Catalog is in read-only mode."
     @read_only_message ||= ENV.fetch("ORANGELIGHT_READ_ONLY_MESSAGE", default_msg)
   end
 
--- a/config/initializers/read_only_mode.rb
+++ b/config/initializers/read_only_mode.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+module Orangelight
+  def read_only_mode
+    @read_only_mode ||= ENV.fetch("ORANGELIGHT_READ_ONLY_MODE", false) == "true"
+  end
+
+  module_function :read_only_mode
+end
--- a/config/initializers/read_only_mode.rb
+++ b/config/initializers/read_only_mode.rb
@@ -2,12 +2,12 @@
 
 module Orangelight
   def read_only_mode
-    @read_only_mode ||= ENV.fetch("ORANGELIGHT_READ_ONLY_MODE", false) == "true"
+    @read_only_mode ||= ENV.fetch("OL_READ_ONLY_MODE", false) == "true"
   end
 
   def read_only_message
     default_msg = "Catalog is in read-only mode."
-    @read_only_message ||= ENV.fetch("ORANGELIGHT_READ_ONLY_MESSAGE", default_msg)
+    @read_only_message ||= ENV.fetch("OL_READ_ONLY_MESSAGE", default_msg)
   end
 
   module_function :read_only_mode, :read_only_message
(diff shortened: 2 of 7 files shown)


app/services/physical_holdings_markup_builder.rb
```diff
@@ -237,8 +237,15 @@ def self.open_location?(location)
...
+    if location.nil?
+      false
+    elsif adapter.unavailable_holding?(holding)
+      false
+    elsif Orangelight.read_only_mode
+      false
+    else
+      location[:requestable]
+    end
   end
```
christinach, line 250: can we use the guard clause with multiple ors instead of multiple lines with false?
`return false if location.nil? || adapter.unavailable_holding?(holding) || Orangelight.read_only_mode`
`location[:requestable]`

spec/controllers/account_controller_spec.rb
```diff
@@ -31,6 +31,21 @@
...
+        allow(Orangelight).to receive(:read_only_mode).and_return(true)
+      end
+
+      it 'redirects to root and flashes an explanatory message' do
+        get :index
+        expect(response).to redirect_to(root_path)
+        expect(flash[:notice]).to include("Account page is disabled", "read-only")
+      end
+    end
   end
```
christinach, line 50: 👍🏻

spec/features/request_options_spec.rb
```diff
@@ -24,6 +24,16 @@
...
 
+  describe 'the request button when orangelight is in read-only mode', js: true do
+    it 'does not display a request button' do
+      allow(Orangelight).to receive(:read_only_mode).and_return(true)
+      visit '/catalog/9618072'
+      using_wait_time 5 do
+        expect(page).to have_selector('td[data-requestable="false"]')
+      end
+    end
+  end
```
christinach, line 35: 👍🏻

spec/views/shared/_announcement.html.erb_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+    end
+  end
+
+  context 'in Alma mode' do
+    it 'renders an Alma message' do
+      allow(Rails.configuration).to receive(:use_alma).and_return(true)
+      render
+      expect(rendered).to match(/Catalog Alma QA site/)
+    end
+  end
```
christinach, line 26: 👍🏻

spec/views/shared/_announcement.html.erb_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+    expect(rendered).to match(/Many Library services are available online/)
+  end
+
+  context 'in read-only mode' do
+    it 'renders a read-only message' do
+      allow(Orangelight).to receive(:read_only_mode).and_return(true)
+      allow(Orangelight).to receive(:read_only_message).and_return("test message")
+      render
+      expect(rendered).to match(/test message/)
+    end
```
christinach, line 17: 👍🏻

christinach (approved):

christinach (changes requested):
@eliotjordan I have a question in the comment.

app/services/physical_holdings_markup_builder.rb
hackartisan: To me it's more readable to have each condition on its own line.

@eliotjordan is likely trying this in response to some discussions we've had in refactorama about postfix conditionals. It turns out that they hide test coverage holes because even if the condition is not tested the line is hit. Because of that, I would vote for the form here.

config/initializers/read_only_mode.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module Orangelight
+  def read_only_mode
+    @read_only_mode ||= ENV.fetch("ORANGELIGHT_READ_ONLY_MODE", false) == "true"
+  end
+
+  def read_only_message
+    default_msg = "Orangelight is in read-only mode."
```
hackartisan, line 9: Probably should say "The Catalog" instead of "Orangelight"

hackartisan (changes requested):
See inline. Also, I'm curious what the user story is for putting it into read only mode. Do We need an ansible PR to go along with this to toggle the ENV?

I'm thinking about whether we might have read only mode and alma mode at the same time. Right now I'm not sure we have that use case so I'm happy to leave the conditions mutually exclusive.

config/initializers/read_only_mode.rb
eliotjordan: Yep. Updated.

eliotjordan:
@hackartisan The user story is that we would run an Ansible playbook, like we do for Figgy. I created a follow-on ticket for that: https://github.com/pulibrary/princeton_ansible/issues/2308

app/services/physical_holdings_markup_builder.rb
eliotjordan: That was exactly my thinking here. Granted it's not perfect, but the entire Class needs a refactor and hopefully this logic can be simplified at that time.

christinach (approved):

hackartisan (approved):

PR #2504, opened 2021-06-11
Retry on staff_view

### LINKED ISSUE #2414: Handle 429 from bibdata bibliographic/id endpoint
We should look for the 429 here and raise an error: https://github.com/pulibrary/orangelight/blob/ace3e1ba6188b85a091eac310adb82b18ac1cfe2/app/models/concerns/blacklight/solr/document/marc.rb#L244

The error should be used here to present a message "Too many requests at this time -- please try again later": https://github.com/pulibrary/orangelight/blob/ace3e1ba6188b85a091eac310adb82b18ac1cfe2/app/views/catalog/_marc_view.html.erb

hectorcorrea:
Shows a retry message to the user when we cannot get the MARC data because Alma returns error message "too many requests per second" (HTTP 429). 

Fixes #2414

### DIFF
--- a/app/models/concerns/blacklight/solr/document/marc.rb
+++ b/app/models/concerns/blacklight/solr/document/marc.rb
@@ -129,6 +129,11 @@ module Blacklight
           'unknown'
         end
 
+        # We allow the user to retry in very specific scenarios.
+        def can_retry?
+          @can_retry
+        end
+
         protected
 
           def build_ctx(format = nil)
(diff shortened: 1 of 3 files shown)


app/models/concerns/blacklight/solr/document/marc.rb
```diff
@@ -242,6 +247,7 @@ def marc_record_from_marcxml
             id = fetch(_marc_source_field)
 
             response = Faraday.get("#{Requests.config['bibdata_base']}/bibliographic/#{id}")
+            @can_retry = response.status == 429
```
jrgriffiniii, line 250: 👍

jrgriffiniii (approved):

Issue #2450, opened 2021-05-13
Define database migration plan


hackartisan:
Implementation tasks:
- [x] Add new browse tables for use by alma infrastructure
- [x] Ensure bookmark links will still work
- [x] Hook catalog production DB to catalog-alma production.

Additional tasks:
- [x] provide info for a "what will happen" website
  - note that e-resources (e-journals and some e-books) are not going to have mappable IDs and so a link would not be redirected)

Background:
- Initially we considered pointing to the same database we have been using in production. However, the browse lists cannot be shared between the 2 sites, as indicated in comments below.
- Then we decided to migrate the database, but this is non-trivial due to its size.
- Finally we decided to add additional tables for the alma infrastructure to use.

christinach:
We would have to disable the cron jobs first.

christinach:
maybe batter not deploy them on the alma_qa worker boxes

tpendragon:
I'm wondering about an alternative plan here:

What if we made another set of browse list tables, for alma? If OL is configured to use Alma it uses those, otherwise it uses the old ones.

christinach:
@tpendragon my thought was to create new browse lists tables to close this issue https://github.com/pulibrary/bibdata/issues/1362. Now that we have a separate alma branch I dont think there is a problem to have the cron jobs running on the new pul indexer boxes https://github.com/pulibrary/orangelight/blob/af08d2c343e81c7be89c5ac91d65f5305bce72db/config/schedule.rb.

The migration should be considered for other tables. For example for bookmarks.

hackartisan:
Will doubling the browse lists be a database size issue? Look on postgres and see how big the tables are.

tpendragon:
I've done this in QA.  A couple things:

1. In catalog-alma-qa there's no way to login to see your bookmarks, except via the link that appears in the "temporary bookmarks" piece. The page you get at that point has barcode login - is that still a thing?
2. The bookmark count is right, but after clicking you get no bookmarks. This is because it does a solr search on all the stored IDs, which are now wrong. I think we can patch the barcode controller to replace all those IDs with alma equivalents and this feature should work fine.
3. Search history doesn't migrate. This is because searches aren't attaching to users, it's just displaying the searches from your session. I don't know if this is a bug or I'm misunderstanding the feature or what. Are there still "saved searches"? In any case, it might be that these are just lost.

hackartisan:
There's a link to log in on both the bookmarks page and the saved searches page.

[...]

PR #2517, opened 2021-06-16
Search for Alma equivalent IDs for bookmark display.

tpendragon:
Work towards #2450

### DIFF
--- a/app/controllers/bookmarks_controller.rb
+++ b/app/controllers/bookmarks_controller.rb
@@ -13,9 +13,15 @@ class BookmarksController < CatalogController
       }
   end
 
+  # Copied from
+  # https://github.com/projectblacklight/blacklight/blob/040933c7a383cd0c5be5895d51ab1004ef3ad5e1/app/controllers/concerns/blacklight/bookmarks.rb#L40-L57
   def index
     @bookmarks = token_or_current_or_guest_user.bookmarks
     @response, deprecated_document_list = search_service.fetch(bookmark_ids)
+    # <Princeton Modifications>
+    # Commented out to use the instance method instead, which adds alma IDs.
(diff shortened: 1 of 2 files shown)


app/controllers/bookmarks_controller.rb
```diff
@@ -26,12 +43,22 @@ def csv
...
       _, @documents = search_service.fetch(bookmark_ids, rows: bookmark_ids.length, fl: '*')
     end
 
+    def alma_ids(bookmark_ids)
+      bookmark_ids.map do |id|
+        "99#{id}3506421"
+      end
+    end
+
     # byte-order-mark declaring our output as UTF-8 (required for non-ASCII to be handled by Excel)
```
christinach, line 62: This Pr looks ok for the voyager ids. I have a question for @mzelesky about the SCSB ID extraction. If the princeton recap ids are different then we would need to consider this transformation as well for the bookmarks. But I'm not sure.

christinach (approved):

app/controllers/bookmarks_controller.rb
```diff
@@ -13,6 +13,23 @@ class BookmarksController < CatalogController
       }
   end
 
+  def index
```
hackartisan, line 16: Assuming this was copied from somewhere and modfiied, can you add a comment linking to where it was copied from and describing the modification? This sort of thing can be helpful during upgrades.

hackartisan (changes requested):

app/controllers/bookmarks_controller.rb
christinach: @mzelesky said that the Ids of the Princeton records we send to scsb will not change.

hackartisan (approved):

PR #2527, opened 2021-06-22
Availability for records in temporary locations

### LINKED ISSUE #2524: Availability in search results for bib records in temporary locations
(this issue is a spin off https://github.com/pulibrary/orangelight/issues/2417) 

The search results page does not show availability for records in temporary locations. For example:

https://catalog-alma-qa.princeton.edu/catalog?utf8=%E2%9C%93&search_field=all_fields&q=9959958323506421

bib_in_temp_location

Notice that availability information is reported correctly for this record in the Show page (because we make an extra call there to request `deep=true` when fetching availability.)

hectorcorrea:
Use 'Check record' as the label for records in temp locations since otherwise they stay with the 'Loading...' label. Notice that we cannot determine the availability for these records because Alma does not return it to us on the normal bib level API call. In the show page we make a second (and slow-ish) API call to fetch the complete information and hence the hint to "Check Record" to allow the user to view all the information if they are indeed interested in this record.

Fixes #2524

temp_location_av

### DIFF
--- a/app/javascript/orangelight/availability.es6
+++ b/app/javascript/orangelight/availability.es6
@@ -129,7 +129,7 @@ export default class AvailabilityUpdater {
         // In this case we set all of them to "Check record" because we can get this
         // information in the Show page.
         const badges = $(`*[data-availability-record='true'][data-record-id='${record_id}'] span.availability-icon`);
-        badges.text("Check record")
+        badges.text("Check record for availability")
         return true;
       }
 
--- a/app/javascript/orangelight/availability.es6
+++ b/app/javascript/orangelight/availability.es6
@@ -123,15 +123,25 @@ export default class AvailabilityUpdater {
   // search results
   process_result(record_id, holding_records) {
     for (let holding_id in holding_records) {
-      const availability_info = holding_records[holding_id];
+      if (holding_id.startsWith('fake_id_')) {
+        // In this case we cannot correlate the holding data from the availability API
+        // (holding_records) with the holding data already on the page (from Solr).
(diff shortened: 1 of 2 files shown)


app/javascript/orangelight/availability.es6
```diff
@@ -123,15 +123,25 @@ export default class AvailabilityUpdater {
...
       // In Alma the label from the endpoint includes both the library name and the location.
+      const availability_info = holding_records[holding_id];
       if (availability_info['label']) {
         const location = $(`*[data-location='true'][data-record-id='${record_id}'][data-holding-id='${holding_id}'] .results_location`);
         location.text(availability_info['label']);
       }
       const availability_element = $(`*[data-availability-record='true'][data-record-id='${record_id}'][data-holding-id='${holding_id}'] .availability-icon`);
 
-      if (availability_info['temp_loc']) {
+      if (availability_info['temp_location']) {
```
hectorcorrea, line 144: This is unrelated to this PR and the `if` block is probably not even needed anymore, but at least now it uses the correct property name (`temp_location`)

hackartisan (changes requested):
Thank you for including the screen shot. When I looked at it, my first thought was "check record for what?" since it's the first result and I didn't have the context of "Available" or "Not Available" to slot the new value into. I wonder if it should say "Check record for Availability"?

Also you know I am going to ask! How feasible would it be to add a test for this change?

hectorcorrea:
@hackartisan I changed the label (see picture below) and added a test for it.

temp_location_av

hackartisan (approved):
Great, thanks!

Issue #2431, opened 2021-03-22
Display boundwith host holdings on boundwith constituent records


christinach:
As part of the show page logic, a boundwith constituent record (i.e a record with a `contained_in_s` value) should pull physical holdings from its host record (given by the mms id in that field).

Availability needs to populate correctly on the show page for these records. In search results, availability should be "View Record for Full Availability", a value that we're already using in other places.

Example records: the host record 99121886293506421 with the constituent records 9929455783506421, 9929455793506421 and 9929455773506421.
https://catalog-alma-qa.princeton.edu/catalog/99121886293506421

hackartisan:
As an alternative to actually inserting the host holdings on the page, we could bump the link to the host record to a more prominent position.

Note there might be some constituent records that do have holdings.

PR #2536, opened 2021-06-25
Pseudo code to show the flow to fetch holdings from a host record

hectorcorrea:
Prototype to address #2431

This is a rather crude proof of concept to show how we could detect that we need to fetch holdings from a host record. YThe basic idea is to (1) detect that this record is container in another, (2) fetch the data from the host record, and (3) merge the holdings from the host into the current record. 

Only the first part is shown in this draft PR.

Unfortunately the field that indicates the host record is currently not being populated because [we are indexing the wrong field](https://github.com/pulibrary/bibdata/blob/alma/marc_to_solr/lib/traject_config.rb#L586) (we are indexing `773$w` but we need `773$x`) so the logic is hard-coded but it shows the general idea.

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -223,6 +223,7 @@ class CatalogController < ApplicationController
     config.add_index_field 'pub_created_display', label: 'Published/Created'
     config.add_index_field 'format', label: 'Format', helper_method: :format_icon
     config.add_index_field 'holdings_1display', show: false
+    config.add_index_field 'contained_in_s', show: false
     config.add_index_field 'isbn_t', show: false
     config.add_index_field 'score', show: false
     config.add_index_field 'marc_relator_display', show: false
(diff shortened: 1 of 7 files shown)


hackartisan:
Hm, the description of the PR says we are indexing `w` but want `x`, and the code comment says we are indexing `x` but want `w`. I am pretty sure `w` is what we want, right?

This seems good and we would still need to figure out the search results availability piece.

app/controllers/catalog_controller.rb
```diff
@@ -673,4 +673,26 @@ def index
...
+
+  def show
+    # Let Blacklight do the normal flow
+    value = super
+
+    host_id = nil
+    if @document.id == "9929455793506421" || @document["contained_in_s"]
+      # This value should come from @document["contained_in_s"] but right now the value
+      # it's empty because we are indexing subfield x instead of subfield w.
+      host_id = "9929455793506421" || @document["contained_in_s"]
```
christinach, line 685: We are indexing 773w. https://github.com/pulibrary/bibdata/blob/35e81e844b6491cdedd65f20313838364eaf31b7/marc_to_solr/lib/traject_config.rb#L586

christinach:
I think that records that have 773x were part of a test. @mzelesky can provide more input on that.

mzelesky:
Yes, that is the old version of a record. There is a new version of that record that uses the final structure. The subfield x was a mistake.

hectorcorrea:
Great, thanks for the clarification @christinach @mzelesky

hectorcorrea:
Replaced with https://github.com/pulibrary/orangelight/pull/2553

PR #2576, opened 2021-07-16
Remove Request button if item is on reserve.

hectorcorrea:
@kevinreiss this should remove the request button from the Show page for items that are on reserves. This PR is on draft since I am still testing it but it looks like the change to support this will be minimal.

### DIFF
--- a/app/javascript/orangelight/availability.es6
+++ b/app/javascript/orangelight/availability.es6
@@ -330,6 +330,9 @@ export default class AvailabilityUpdater {
     if (availability_label_text === "Reserved for digital lending") {
       location_services_element.remove();
     }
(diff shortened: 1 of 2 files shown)


kevinreiss (approved):
@hectorcorrea as you say this looks pretty clear and low risk. Hiding the button was the preferred outcome for circ staff so I think this should be good to get integrated after go live.

carolyncole:
Seems like it works, but it is a little bit confusing seeing the button there and then go away.  Maybe we should ticket the UX of this change.  Seeing available with no button may be confusing to the user.  Maybe instead of the button just disappearing it should say "On Site Access" but that is for another time...
Screen Shot 2021-07-16 at 3 29 01 PM

carolyncole (approved):

PR #2583, opened 2021-07-22
Set all the prod boxes to be workers

hackartisan:
So that sneakers will restart

### DIFF
--- a/config/deploy/alma-production.rb
+++ b/config/deploy/alma-production.rb
@@ -13,11 +13,11 @@ set :branch, ENV['BRANCH'] || 'main'
 # server list. The second argument is a, or duck-types, Hash and is
 # used to set extended properties on the server.
 
-server 'catalog1', user: 'deploy', roles: %i[web app db]
-server 'catalog2', user: 'deploy', roles: %i[web app db sitemap]
-server 'catalog-indexer1', user: 'deploy', roles: %i[cron_prod1 cron_db]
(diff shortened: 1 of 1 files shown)


config/deploy/alma-production.rb
```diff
@@ -13,11 +13,11 @@
 # server list. The second argument is a, or duck-types, Hash and is
 # used to set extended properties on the server.
 
-server 'catalog1', user: 'deploy', roles: %i[web app db]
-server 'catalog2', user: 'deploy', roles: %i[web app db sitemap]
-server 'catalog-indexer1', user: 'deploy', roles: %i[cron_prod1 cron_db]
-server 'catalog-indexer2', user: 'deploy', roles: %i[cron_prod2]
-server 'catalog-indexer3', user: 'deploy', roles: %i[cron_prod3]
+server 'catalog1', user: 'deploy', roles: %i[web app db worker]
```
christinach, line 16: why are you also setting catalog1 and catalog2 and not only the 3 workers?

christinach: Is this because of the numismatics that we want to make sure the sneakers restart in catalog1 and catalog2?

hackartisan: Because they're all running sneakers workers. If we don't allow them to restart they'll be running outdated code.

christinach (approved):

PR #2597, opened 2021-07-30
Updates the show page to make an extra Javascript call to get availab…

hectorcorrea:
This is the Orangelight complement for PR https://github.com/pulibrary/bibdata/pull/1565

Since now bib data will return "Unavailable" for CDL records unless we request a deep check this PR makes a second call to bib data with `deep=true` for any record that is Unavailable. Notice that we only make this second call from the Show page.

This improves the Search Results page performance significantly. The drawback is that if a record is Unavailable according to Alma but available through CDL the user will not notice this unless they go into the Show page for that record. 

~~We could try to make the second call with `deep=true` from the Search results after the initial call with `deep=false` if there are Unavailable records in the search results. I have not tried that.~~

### DIFF
--- a/app/javascript/orangelight/availability.es6
+++ b/app/javascript/orangelight/availability.es6
@@ -199,8 +199,11 @@ export default class AvailabilityUpdater {
     var url = `${this.bibdata_base_url}/bibliographic/${mms_id}/availability.json?deep=true`;
     $.getJSON(url, (data) => { this.update_single(data, mms_id); })
       .fail((jqXHR, textStatus, errorThrown) => {
-        // Notice that this console.error() message shows in the console output when running `yarn test`
-        return console.error(`Failed to retrieve deep availability data for bib. record ${mms_id}: ${errorThrown}`);
+        // Notice that this console.error() message shows in the Terminal output when running `yarn test`
+        // even while all the tests pass. The only way I have found to prevent it from showing up in the
+        // Terminal is by declaring this gratuitous variable WITHOUT var/let/const.
+        dummy_var_1 = true
+        return console.error(`Failed to retrieve deep availability data for bib. record ${mms_id}: ${errorThrown}`)
       });
 
     return;
--- a/app/javascript/orangelight/availability.es6
+++ b/app/javascript/orangelight/availability.es6
@@ -183,7 +183,7 @@ export default class AvailabilityUpdater {
         dataComplete = false; // The data that we get from Alma for temporary locations is incomplete.
         break;
       }
-      if (availability_info['status_label'] == 'Unavailable') {
+      if ((availability_info['status_label'] == 'Unavailable') && (availability_info['cdl'] !== true)) {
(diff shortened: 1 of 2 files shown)


hackartisan (commented):
Will this mean "unavailable" items continue to have status "Loading..."?

hectorcorrea:
@hackartisan no, they will show "Unavailable". As soon as the first JavaScript calls returns (and this is the "fast" API call) all records will have they availability updated (Available or Unavailable). 

I am about to push an update to this PR that will retry the Availability for "Unavailable" items with the slow API call. More to follow.

hectorcorrea:
I've pushed a new commit that has the retry logic (via the slow-ish API call) for Unavailable items. This will keep the same user experience as our original design, but the slow API call will __not__ block the initial display of Available/Unavailable items. 

The only thing that might be disconcerting to the user is that if an item is Unavailable initially but then we recognize that  it's available through CDL they will see the availability badge toggle from "Unavailable" to "Available" once the slow-ish call returns. Not ideal but not horrible either, particularly since the toggle will only happen for those (few?) items in CDL that were reported as Unavailable in the first call.

hectorcorrea:
I've reverted the code to make to make the second call in the Search page. 

The only second call to bibdata will happend in the *show* page if the record is unavailable according to Alma (the idea being that on the second call will ask bibdata to check deeper to see if it is available via CDL.)

Still pending to update the test suite to match the new code path.

hectorcorrea:
This is all ready to go, but if there is no rush, I would like to wait another day. I am puzzled by a `console.log()` error that `yarn test` outputs to the Terminal even as all test pass. See screenshot below. 

The fact that the message is being shown in the terminal is harmless, but I would like to better understand what test is causing it. I though I had it but it seems to jump from one test to another. 

yarn_test

Sincerely,
Puzzled Dev in CT

app/javascript/orangelight/availability.es6
```diff
@@ -195,7 +199,11 @@ export default class AvailabilityUpdater {
     var url = `${this.bibdata_base_url}/bibliographic/${mms_id}/availability.json?deep=true`;
     $.getJSON(url, (data) => { this.update_single(data, mms_id); })
       .fail((jqXHR, textStatus, errorThrown) => {
-        return console.error(`Failed to retrieve deep availability data for bib. record ${mms_id}: ${errorThrown}`);
+        // Notice that this console.error() message shows in the Terminal output when running `yarn test`
+        // even while all the tests pass. The only way I have found to prevent it from showing up in the
+        // Terminal is by declaring this gratuitous variable WITHOUT var/let/const.
+        dummy_var_1 = true
```
hectorcorrea, line 205: This is a horrible hack, but it was the only way could avoid the behavior I described here: https://github.com/pulibrary/orangelight/pull/2597#issuecomment-891329066

hectorcorrea:
Not needed, this was taken care of by https://github.com/pulibrary/orangelight/pull/2601

PR #2611, opened 2021-08-03
Minor fix for CDL check

hectorcorrea:
Moved the `deep=true` parameter to the front of the URL so that we don't build a bad URL if the record is a bound-with and we request availability information for more than one bib. 

See also https://github.com/pulibrary/orangelight/pull/2601

### DIFF
--- a/app/javascript/orangelight/availability.es6
+++ b/app/javascript/orangelight/availability.es6
@@ -27,6 +27,7 @@ export default class AvailabilityUpdater {
     this.update_single = this.update_single.bind(this);
     this.update_availability_undetermined = this.update_availability_undetermined.bind(this);
     this.process_scsb_single = this.process_scsb_single.bind(this);
+    this.availability_url_show = this.availability_url_show.bind(this);
   }
 
   request_availability(allowRetry) {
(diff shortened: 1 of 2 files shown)


hackartisan (commented):
What would a bad url look like?

hectorcorrea:
@hackartisan The line below the fix adds a MMS_ID to the URL for bound-with records and it would produce a URL like this: `/bibliographic/availability.json?bib_ids=444&deep=true,555` 

With this PR the URL will be `/bibliographic/availability.json?deep=true&bib_ids=444,555`

hackartisan:
Ah, I see! I think a test would be good to protect against regression. It may be easy to test if it's extracted into a helper function?

hectorcorrea:
Honestly I don't think it is worth spending the time refactoring the code to account for this, but if  you think we should I'll do it.

hackartisan:
I understand. I won't block this PR.

hackartisan:
I do think extracting it would also make it easier to read / understand.

christinach:
> @hackartisan The line below the fix adds a MMS_ID to the URL for bound-with records and it would produce a URL like this: `/bibliographic/availability.json?bib_ids=444&deep=true,555`
> 
> With this PR the URL will be `/bibliographic/availability.json?deep=true&bib_ids=444,555`

@hectorcorrea can you give an example record where you've seen that this `/bibliographic/availability.json?bib_ids=444&deep=true,555` is happening (if there are more than one ids)? I haven't pulled this pr to test it locally.

hectorcorrea:
OK. I will add a test to this so that we can all move on.

hectorcorrea:
@hackartisan moved the logic to build the URL to its own method and added a test for it. Thanks for the push, this was indeed an improvement to the code.

tpendragon (approved):

tpendragon:
Y'all rock.

tpendragon (approved):

christinach (approved):

christinach:
@hectorcorrea thanks! Can you add @hackartisan in the last commit?

hackartisan:
Thank you!

Issue #2621, opened 2021-08-04
Update orangelight deploy targets


hackartisan:
Remove alma-production and alma-qa. Update production and qa to point to the correct machines. Comment out servers from staging config file.

Make corresponding changes in pulbot. Announce to developers when this is happening.

PR #2633, opened 2021-08-06
Disable staging deploy

hackartisan:
Until there are staging boxes for alma

refs #2621


hackartisan:
actually I think https://github.com/pulibrary/pulbot/pull/76 is probably enough.

PR #2636, opened 2021-08-06
Handles bad session

### LINKED ISSUE #2635: Error with nil UID
NoMethodError: undefined method `empty?' for nil:NilClass

## Backtrace

    line 77 of [PROJECT_ROOT]/app/controllers/users/omniauth_callbacks_controller.rb: barcode_token_in_request_params
    line 38 of [PROJECT_ROOT]/app/controllers/users/omniauth_callbacks_controller.rb: barcode_user
    line 42 of [PROJECT_ROOT]/app/controllers/users/omniauth_callbacks_controller.rb: patron

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/54399/faults/80779922)

Visiting URL  https://catalog.princeton.edu/users/auth/barcode/callback (with a bad/empty session?) causes the error.

hectorcorrea:
This PR handles the bad session so that now instead of crashing and showing an HTTP 500 

http_500

we handle the error and that results in the (expected?) workflow to take place: redirect the user and show a flash warning indicating that the session in invalid.

redirect_flash

Fixes #2635

### DIFF
--- a/spec/controllers/omniauth_callbacks_controller_spec.rb
+++ b/spec/controllers/omniauth_callbacks_controller_spec.rb
@@ -26,6 +26,21 @@ RSpec.describe Users::OmniauthCallbacksController do
     end
   end
 
+  describe 'bad omniauth session' do
+    let(:omniauth_response_nil_uid) do
+      OmniAuth::AuthHash.new(provider: 'barcode', uid: nil, info: { last_name: "" })
+    end
+    before do
+      controller.request.env['omniauth.auth'] = omniauth_response_nil_uid
+    end
+    it 'handles nil UID' do
+      allow(User).to receive(:from_barcode) { valid_barcode_user }
+      allow(Bibdata).to receive(:get_patron) { guest_response }
+      get :barcode
+      expect(response).to redirect_to(account_path)
+    end
+  end
+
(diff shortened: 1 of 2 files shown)


spec/controllers/omniauth_callbacks_controller_spec.rb
```diff
@@ -26,6 +26,21 @@
...
 
+  describe 'bad omniauth session' do
+    let(:omniauth_response_nil_uid) do
+      OmniAuth::AuthHash.new(provider: 'barcode', uid: nil, info: { last_name: "" })
+    end
+    before do
+      controller.request.env['omniauth.auth'] = omniauth_response_nil_uid
+    end
+    it 'handles nil UID' do
+      allow(User).to receive(:from_barcode) { valid_barcode_user }
```
christinach, line 37: If the message is '...could not authenticate because barcode or last name did not match active patron..', can we use an invalid_barcode_user in the test?

hectorcorrea: The user itself is irrelevant for this test, what causes the error is the `nil` UID in the OmniAuth session.

christinach (approved):

PR #2641, opened 2021-08-09
Loads Rails environment so that Rails.logger.error() is available

### LINKED ISSUE #2639: Error in rake browse:call_numbers
[Orangelight] NoMethodError: undefined method `error' for nil:NilClass

## Backtrace

    line 81 of [PROJECT_ROOT]/lib/orangelight/browse_lists/call_number_csv.rb: block in solr_cursor_page_body
    line 76 of [PROJECT_ROOT]/lib/orangelight/browse_lists/call_number_csv.rb: loop
    line 76 of [PROJECT_ROOT]/lib/orangelight/browse_lists/call_number_csv.rb: solr_cursor_page_body

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/54399/faults/80786210)

hectorcorrea:
Fixes #2639 

Although the real error behind 2639 was a Solr error, the error reported by Honeybadger was because it crashed when it tried to log the error. This PR addresses only the error reported by Honeybadger (i.e. not the Solr error)

I did not add a test since the problem was the declaration of the Rake task, let me know if I should add a test for that kind of mis-declaration of a rake task.

### DIFF
--- a/lib/tasks/browse.rake
+++ b/lib/tasks/browse.rake
@@ -26,7 +26,7 @@ namespace :browse do
   end
 
   desc 'Pull data for call numbers browse'
-  task :call_numbers do
+  task call_numbers: :environment do
     _sql_command, facet_request, conn = BrowseLists.connection
     BrowseLists::CallNumberCSV.new(facet_request, conn).write
   end


hackartisan (approved):
👍🏻 looks good, thank you!

hackartisan:
And yes we normally don't test rake tasks themselves. If there's enough logic in them we would pull out the logic into a function somewhere and test that function.

christinach (approved):

PR #2650, opened 2021-08-11
Caches the Solr data for the home page

### LINKED ISSUE #2637: Cache the home page
We shouldn't have to wait for it to load; there's nothing particularly dynamic.

hectorcorrea:
Fixes #2637

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -670,8 +670,52 @@ class CatalogController < ApplicationController
   end
 
   def index
-    super
+    if home_page?
+      render_empty_search
+    else
+      super
+    end
   rescue ActionController::BadRequest
     render file: Rails.public_path.join('x400.html'), layout: true, status: :bad_request
   end
+
+  private
+
+    def render_empty_search
+      # This code is a copy of Blacklight::Catalog.index() method but adapted to use
+      # a cached version of the data rather than requesting the data from Solr.
+      # See https://github.com/projectblacklight/blacklight/blob/v7.0.1/app/controllers/concerns/blacklight/catalog.rb#L25-L41
+      @response = empty_solr_response
+      @document_list = @response.documents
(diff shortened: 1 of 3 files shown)


app/controllers/catalog_controller.rb
```diff
@@ -670,8 +670,52 @@ def render_search_results_as_json
...
+        end
+        additional_response_formats(format)
+        document_export_formats(format)
+      end
+    end
+
+    def home_page?
+      # When only the "controller" and "action" keys are in the request (i.e. no query or facets)
+      # we consider it the home page.
+      params.keys.count == 2
```
hectorcorrea, line 705: We could tweak this logic to make sure that neither `q` nor `f` (facet) are present. But as-is it should catch the vast majority of calls.

spec/controllers/catalog_controller_spec.rb
```diff
@@ -112,4 +112,18 @@
       expect(assigns(:hathi_url)).to be_nil
     end
   end
+
+  describe 'home page' do
+    it 'renders empty searches for the home page' do
+      expect(controller).to have_received(:render_empty_search)
```
hectorcorrea, line 118: I am only testing that the `render_empty_search` method was called in the Controller but not the logic inside of it nor the code inside `empty_solr_response` and `empty_raw_response`. Let me know if you think we should test deeper.

hackartisan (changes requested):
This looks good to me but rspec is saying the test needs mocks or spies and I don't think you want to mock the controller under test. I wonder if a mock of the Rails cache would work, and then you could test that it calls fetch. And/or a test that that logger message is written?

hectorcorrea:
@hackartisan I fixed the tests and now they pass (sorry about that!) and in addition they now check whether the Rails cache was used which is a better test for this scenario.

christinach (approved):

PR #2651, opened 2021-08-12
Use read only mode for database maintenance windows

### LINKED ISSUE #2648: Repurpose READ_ONLY flag for database downtime
Before Alma, the READ_ONLY flag was used to redirect from the account page. But the account page has changed quite a bit and we don't have a use case currently for redirecting away from the digitization requests page, per se. We do have use cases (see #2600 and #2371) for a database-level read only mode.

hackartisan:
closes #2648


hectorcorrea:
"digitization" is misspelled in the banner

banner

hackartisan:
> "digitization" is misspelled in the banner

Thank you!!

christinach:
Who will have the responsibility to switch the catalog in the read only mode during a database maintenance window?

PR #2652, opened 2021-08-13
Added library name in email and print view

### LINKED ISSUE #2616: Update Bookmarks to include both the library name and the location name in the email and print view
In the bookmarks view there are tools to print and to email the bookmarks. With the change in the location display we show only the location in these views. We should update the code so that we display both the library and the location name. 

From catalog-feedback: `Locations are set to display differently in the catalog now ("Firestone Library - Stacks", etc.), and one consequence seems to be that they don't carry over fully when using Email or Print function from Bookmarks; e.g. the e-mailed list of bookmarked items only shows "Stacks" or "Reference", etc., not Firestone or Mendel `

XanderGardner:
closes #2616

### DIFF
--- a/spec/models/email_spec.rb
+++ b/spec/models/email_spec.rb
@@ -9,13 +9,13 @@ RSpec.describe Blacklight::Document::Email do
     describe 'record with vernacular script' do
       let(:properties) do
         {
-          id: '9153359',
+          id: '9991533593506421',
           title_display: 'al-Amāzīgh : mawsūʻat tārīkh duwal al-Maghrib al-ʻArabī',
           title_vern_display: 'الأمازيغ : موسوعة تاريخ دول المغرب العربي.',
           pub_created_display: ['al-Muhandisīn, al-Jīzah : Dār Halā lil-Nashr wa-al-Tawzīʻ, 2016.',
                                 'المهندسين، الجيزة : دار هلا للنشر والتوزيع، 2016.'],
           format: ['Book'],
-          holdings_1display: '{"9034559":{"location":"ReCAP","library":"ReCAP","location_code":"rcppa",'\
+          holdings_1display: '{"9034559":{"location":"Remote Storage","library":"ReCAP","location_code":"recap$pa",'\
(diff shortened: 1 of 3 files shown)


spec/models/email_spec.rb
```diff
@@ -15,7 +15,7 @@
           pub_created_display: ['al-Muhandisīn, al-Jīzah : Dār Halā lil-Nashr wa-al-Tawzīʻ, 2016.',
                                 'المهندسين، الجيزة : دار هلا للنشر والتوزيع، 2016.'],
           format: ['Book'],
-          holdings_1display: '{"9034559":{"location":"ReCAP","library":"ReCAP","location_code":"rcppa",'\
+          holdings_1display: '{"9034559":{"location":"Remote","library":"ReCAP","location_code":"rcppa",'\
```
christinach, line 18: It would be nice if you could update the Voyager id with an Alma id. In this case it would be: '9991533593506421' vs "9034559" . Alma is the new system we started using. You can get the 'holdings_1display' from https://catalog.princeton.edu/catalog/9991533593506421/raw if you search for 'holdings_1display'. The information that is different in the holdings_1display is that "location": "Remote Storage" and "location_code": "recap$pa".

spec/models/email_spec.rb
```diff
@@ -72,7 +72,7 @@
       it 'includes format' do
         expect(document).to match('Format: Senior Thesis')
       end
-      it 'includes library name when location field is not present in holdings' do
+      it 'includes only library name when location field is not present in holdings' do
```
christinach, line 75: @XanderGardner can you please update this test as well? I'm sorry I missed it earlier. If you'd like to check the record:  https://catalog.princeton.edu/catalog/dsp01zk51vk08g/raw. It doesn't have holdings_1display. You can remove it. You will have to adjust your test and remove the last one `..to match('Location: Online')`.

christinach: @XanderGardner yes if you can update this. It would be great.

christinach (approved):
Thank you @XanderGardner! 💯

PR #2676, opened 2021-08-20
Updating orangelight to allow undergraduates access to pickup books

carolyncole:


christinach:
@carolyncole I compared the two examples you provided in slack. The one in catalog-qa after deploying this PR. https://catalog-qa.princeton.edu/requests/99124624583506421?mfhd=22707996030006421&source=pulsearch

Screen Shot 2021-08-20 at 4 58 56 PM

vs the one in production https://catalog.princeton.edu/requests/99124624583506421?mfhd=22707996030006421&source=pulsearch

Screen Shot 2021-08-20 at 4 59 26 PM

You mentioned in slack that the one in production should only allow EDD.
They both seem to allow physical item pickup. I dont see a difference.

carolyncole:
@christinach The difference should only have been for undergraduate students.  Were you checking as an undergraduate?  I'm unsure how to become a student.  If you have a trick I would love to know it.

Although I am realizing that I did not quite create this branch correctly, so maybe the restrictions for undergraduates were removed earlier and this is just clean up.  It is hard to test because anyone authorized to be on campus has always had pickup rights.  This change should only impact undergraduates that were never authorized to be on campus.

christinach:
@carolyncole I don't have a way to test as an undergraduate student. I tested as me :). Ignore my screenshots.

kevinreiss (approved):

PR #2677, opened 2021-08-20
Removed umlaut

### LINKED ISSUE #2623: Remove umlaut_base: https://getit.princeton.edu from the requests.yml
We dont use getit we can remove this. It is loading an umlaut js script. 

@kevinreiss can you confirm?

### LINKED ISSUE #2644: Remove umlaut specs
https://github.com/pulibrary/orangelight/blob/main/spec/views/catalog/umlaut_services_spec.rb

We don't use umlaut

XanderGardner:
closes #2644 
closes #2623


christinach:
Thank you @XanderGardner !

christinach (dismissed):

christinach (changes requested):
@XanderGardner I looked again. Thank you for removing the spec and the other reference in the application helper. The spec is testing some methods that they are still in the code. Can you first remove the methods and see the spec failing and then remove the spec as well? The methods I'm referring to are in https://github.com/pulibrary/orangelight/blob/8998c32af47876285885fda357edf23920bcfcc6/app/helpers/application_helper.rb#L15-L30. Thank you again for taking the time to work on this issue.

christinach (approved):

Issue #2680, opened 2021-08-23
Refactor PhysicalHoldingsMarkupBuilder


tpendragon:

kevinreiss:
This class has largely been re-factored to use the view components pattern. Closing.

PR #2681, opened 2021-08-23
Move to instance methods in PhysicalHoldingsMarkupBuilder

hackartisan:
refs #2680

Co-authored-by: Trey Pendragon <tpendragon@princeton.edu>
Co-authored-by: Bess Sadler <bess@ibiblio.org>


app/services/physical_holdings_markup_builder.rb
```diff
@@ -379,6 +376,9 @@ def self.request_placeholder(adapter, holding_id, location_rules, holding)
     markup
   end
 
+  attr_reader :adapter
```
carolyncole, line 379: My only comment is the order of this class is now strange.  Aka usually the attr readers and initializer is at the top of the class.

carolyncole (approved):
I'm not going to block this PR for the class order, but I also will not merge it so you can decide if you care.

app/services/physical_holdings_markup_builder.rb
hackartisan: Yes, good point, it's worth reorganizing it a bit first. I didn't move them around because I thought it would make the diff harder to read but I'll add it in another commit.

Issue #1616, opened 2019-03-25
Monogram display for coins


eliotjordan:

christinach:
![Screen Shot 2019-04-16 at 3.12.32 PM.png](https://zube.io/files/pulibrary/9b26e45b4b2c8b96783a9bacbd06db8a-screen-shot-2019-04-16-at-3-12-32-pm.png)

christinach:
Adding monogram/s thumbnail below coin-thumbnail in the left column.

PR #2691, opened 2021-08-26
Availability display in Show page for CDL items

### LINKED ISSUE #2669: Record page - Control Digital Lending Items have status Unavailable
CDL: https://catalog.princeton.edu/catalog/9999490563506421
https://bibdata.princeton.edu/bibliographic/9999490563506421/availability.json?deep=true

Check why we don't display the status? https://github.com/pulibrary/orangelight/blob/f2c10c3d585162dfad7b20fd96e9cbd2a88f6a6b/app/javascript/orangelight/availability.es6#L376

hectorcorrea:
Display "Online" as the availability for items in CDL instead of "Unavailable". It also *removes* the Request button from the page. Fixes #2669

The only drawback with the current implementation is that the user will initially see the "Request" button next to each item, and after a few seconds (when the call to BibData/Alma comes back) the Request buttons will be removed. This is a known issue due to the slowness of the Alma API to return when we request this level of information (see #1616) and could be tackled separately.

Below are two screenshots showing the page before and after the change.

 ## Before

cdl_before

## After

cdl_after

### DIFF
--- a/app/javascript/orangelight/availability.es6
+++ b/app/javascript/orangelight/availability.es6
@@ -367,14 +367,26 @@ export default class AvailabilityUpdater {
     availability_element.text(status_label);
     availability_element.attr('title', '');
     if (status_label.toLowerCase() === 'unavailable') {
-      availability_element.addClass("badge-danger");
-      if (isCdl && addCdlBadge) {
-        // The _physical_ copy is not available but we highlight that the _online_ copy is.
-        availability_element.attr('title', 'Physical copy is not available.');
-        let cdlPlaceholder = availability_element.parent().next().find("*[data-availability-cdl='true']");
-        cdlPlaceholder.text('Online');
-        cdlPlaceholder.attr('title', 'Online copy available via Controlled Digital Lending');
-        cdlPlaceholder.addClass('badge badge-primary');
+      if (isCdl) {
+        if (addCdlBadge) {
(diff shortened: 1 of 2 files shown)


christinach:
In the past we displayed for cdl item status: 'Reserved for digital lending'.  https://github.com/pulibrary/orangelight/blob/a1a88135b5331b30b087d97263c32da2fba208ac/app/javascript/orangelight/availability.es6#L22 https://github.com/pulibrary/orangelight/blob/a1a88135b5331b30b087d97263c32da2fba208ac/app/javascript/orangelight/availability.es6#L330
@kevinreiss, @kelynch do we want to display this status again instead of 'Online'?

kevinreiss:
I approve of making this change where if I understand it the user will never see a request button for these items at any point no matter how slow the extra availability checks take.

hectorcorrea:
@kevinreiss The current implementation **will initially show** the Request button for a few seconds (4-5 seconds) like it does for Reserves and then, **after Alma comes back with CDL info we hide them**. This is not ideal.

We could try to implement this the other way around: Initially hide the Request button and only show if after the Alma slow API comes back. But this will be a more complicated and riskier change so I would vote to do that separately so that we can test it extensively...because if we mess that one up the Request button won't show up if there is an Alma error or bug in our code.

christinach (approved):

PR #2694, opened 2021-08-31
Prevents the no holdings display when there electronic access links

### LINKED ISSUE #2668: Display issue of online holdings
@tventimi reported:
On 08/17/2021 he loaded a batch of records for an electronic collection:
https://catalog.princeton.edu/?f%5Baccess_facet%5D%5B%5D=Online&f1=in_series&q1=%E4%B8%AD%E5%9C%8B%E8%97%9D%E8%A1%93%E6%96%87%E7%8D%BB%E5%8F%A2%E5%88%8A.&search_field=advanced

Although these appear in Blacklight now, they appear in search results with
"No holdings available for this record".  The same is true for the record
for the electronic collection itself.

https://catalog.princeton.edu/catalog?utf8=%E2%9C%93&search_field=left_anchor&q=+%E4%B8%AD%E8%8F%AF%E7%B6%93%E5%85%B8%E5%8F%A4%E7%B1%8D%E5%BA%AB.+

However, this is not true for all records in this set.  For example, this
record was also loaded as part of the same batch but does show electronic
holdings:

https://catalog.princeton.edu/catalog?utf8=%E2%9C%93&search_field=left_anchor&q=+%E4%B8%A4%E4%BA%AC%E6%96%B0%E8%AE%B0%E8%BE%91%E6%A0%A1+%2F+%5B%E5%94%90%5D%E9%9F%A6%E8%BF%B0%E6%92%B0+%3B+%E8%BE%9B%E5%BE%B7%E5%8B%87%E8%BE%91%E6%A0%A1.+

It looks like a display issue in the search results page for online records.

hectorcorrea:
This takes care of the issue in the Search results where we incorrectly display "No holdings available for this record" and instead it displays the Online badge and the electronic access link

show_elink

Compare with the [current display in production](https://catalog.princeton.edu/catalog?utf8=%E2%9C%93&f%5Baccess_facet%5D%5B%5D=Online&f1=in_series&q1=%E4%B8%AD%E5%9C%8B%E8%97%9D%E8%A1%93%E6%96%87%E7%8D%BB%E5%8F%A2%E5%88%8A.&search_field=all_fields&q=99101494203506421)

With this fix the online badge and link will only be added if all other checks for holdings/portfolios come up empty but we do have an "electronic access" link for the record. 

Fixes #2668

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -265,6 +265,12 @@ module ApplicationHelper
                                         class: 'availability-icon more-info', title: 'Click on the record for full availability info',
                                         'data-toggle' => 'tooltip').html_safe, class: 'empty', data: { record_id: document['id'] })
     end
+
+    if block.empty? && links.count > 0
+      # All other options came up empty but since we have electronic access let's show Online rather than no holdings.
+      block << content_tag(:span, 'Online', class: 'availability-icon badge badge-primary', title: 'Electronic access', 'data-toggle' => 'tooltip')
+    end
+
     if block.empty?
       content_tag(:div, t('blacklight.holdings.search_missing'))
     else
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -266,7 +266,7 @@ module ApplicationHelper
                                         'data-toggle' => 'tooltip').html_safe, class: 'empty', data: { record_id: document['id'] })
     end
 
-    if block.empty? && links.count > 0
+    if block.empty? && links.present?
       # All other options came up empty but since we have electronic access let's show Online rather than no holdings.
       block << content_tag(:span, 'Online', class: 'availability-icon badge badge-primary', title: 'Electronic access', 'data-toggle' => 'tooltip')
     end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -267,8 +267,12 @@ module ApplicationHelper
     end
 
     if block.empty? && links.present?
-      # All other options came up empty but since we have electronic access let's show Online rather than no holdings.
-      block << content_tag(:span, 'Online', class: 'availability-icon badge badge-primary', title: 'Electronic access', 'data-toggle' => 'tooltip')
+      # All other options came up empty but since we have electronic access let's show the
+      # Online badge with the electronic access link (rather than a misleading "No holdings")
+      info = ''
+      info << content_tag(:span, 'Online', class: 'availability-icon badge badge-primary', title: 'Electronic access', 'data-toggle' => 'tooltip')
+      info << links.shift.html_safe
+      block << content_tag(:li, info.html_safe)
     end
 
     if block.empty?
--- a/spec/helpers/holding_block_spec.rb
+++ b/spec/helpers/holding_block_spec.rb
@@ -245,16 +245,16 @@ RSpec.describe ApplicationHelper do
         {
           id: '1',
           format: ['Book'],
-          electronic_access_1display: '{"https://purl.fdlp.gov/GPO/LPS40377":["purl.fdlp.gov"]}'
+          electronic_access_1display: '{"https://library.princeton.edu/resource/28076":["library.princeton.edu"]}'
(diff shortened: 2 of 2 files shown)


hectorcorrea:
If this approach looks like a good solution I'll go ahead and add tests to the PR but for now I am just seeing what others think of the approach.

app/helpers/application_helper.rb
```diff
@@ -265,6 +265,12 @@ def holding_block_search(document)
                                         class: 'availability-icon more-info', title: 'Click on the record for full availability info',
                                         'data-toggle' => 'tooltip').html_safe, class: 'empty', data: { record_id: document['id'] })
     end
+
+    if block.empty? && links.present?
+      # All other options came up empty but since we have electronic access let's show Online rather than no holdings.
+      block << content_tag(:span, 'Online', class: 'availability-icon badge badge-primary', title: 'Electronic access', 'data-toggle' => 'tooltip')
+    end
```
christinach, line 272: I think this is ok for now. Can you add like: 
```
https://github.com/pulibrary/orangelight/blob/main/app/helpers/application_helper.rb#L222-L223
```
We can update lines https://github.com/pulibrary/orangelight/blob/main/app/helpers/application_helper.rb#L213-L224 later. I would need to look closer.

hectorcorrea:
Test have been added.

app/helpers/application_helper.rb
hectorcorrea: Notice that the changes in this PR are **outside** of the holdings loop because there are no holdings (only links) for this record. 

The way I implement this fix I let all the previous logic take place and, if at the end of all the existing logic the (HTML) block is empty but there are links available then it adds the badge. I implement it this way since I did not want to mess with the existing logic and potentially break other holdings/links scenarios.

christinach: https://github.com/pulibrary/orangelight/blob/9bde667d4f25588cd7a17e4dc996754233cbcb32/app/helpers/application_helper.rb#L426

hectorcorrea: @christinach this was a fantastic suggestion. Reusing the code logic that you suggested displays both the Online badge and the text of the link (see screenshot above)

https://github.com/pulibrary/orangelight/blob/issue_2668_online_holdings/app/helpers/application_helper.rb#L269-L276

christinach: Great @hectorcorrea !

spec/helpers/holding_block_spec.rb
```diff
@@ -240,6 +240,24 @@
       end
     end
 
+    context '#holding_block_search with links only' do
+      let(:document) do
```
christinach, line 244: Can you update the test title :...links from the electronic_access_1display.. 
because we have links from other fields as well.

christinach: Ignore the title is ok. We can add more tests later regarding other fields that are also used in the online display.

spec/helpers/holding_block_spec.rb
```diff
@@ -240,6 +240,24 @@
...
+        {
+          id: '1',
+          format: ['Book'],
+          electronic_access_1display: '{"https://library.princeton.edu/resource/28076":["library.princeton.edu"]}'
+        }.with_indifferent_access
+      end
+
+      before { stub_holding_locations }
+
```
christinach, line 253: Can you update this with the `stub_alma_holding_locations` https://github.com/pulibrary/orangelight/blob/57511ef6189e437fd311f2ad0f5ed2c8fb4649de/spec/support/webmock_stubs.rb#L9 ?

hectorcorrea: All the existing tests in this file use `stub_holding_locations`. If we want to refactor the code and leave only one stub method (either `stub_holding_locations` or `stub_alma_holding_locations`) I would suggest we handle that via a separate PR.

christinach: That's fine. It can be done at another time. I have already updated in some places to use the stub_alma_holding_locations.

christinach (approved):

PR #2703, opened 2021-09-01
Export bookmarks to csv with id wrapped in single quotes

### LINKED ISSUE #2695: Exporting Bookmarks as a CSV returns the wrong ID number
When I export my bookmarks as CSV, the ID ends in 420 instead of 421.  e.g., 9912140633506421 comes out as 9912140633506420. 

Reported by Dave Jenkins

christinach:
closes #2695

### DIFF
--- a/spec/controllers/bookmarks_controller_spec.rb
+++ b/spec/controllers/bookmarks_controller_spec.rb
@@ -35,9 +35,11 @@ RSpec.describe BookmarksController do
       expect(body).to include(headers)
       data1.each do |value|
         expect(body).to include(value)
+        expect(body).to include("'9997412163506421'")
       end
       data2.each do |value|
         expect(body).to include(value)
+        expect(body).to include("'9935444363506421'")
       end
 
       expect(body).not_to include(bad_value)
--- a/app/controllers/bookmarks_controller.rb
+++ b/app/controllers/bookmarks_controller.rb
@@ -97,7 +97,13 @@ class BookmarksController < CatalogController
     end
 
     def csv_values(doc, field)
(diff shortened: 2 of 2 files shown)


christinach:
@escowles if you have time

spec/controllers/bookmarks_controller_spec.rb
```diff
@@ -35,9 +35,11 @@
       expect(body).to include(headers)
       data1.each do |value|
         expect(body).to include(value)
+        expect(body).to include("'9997412163506421'")
```
carolyncole, line 38: @christinach I would expect this as a separate check from the set of values, or a change to the data.  Not a check within the check for each value.
Maybe this line https://github.com/pulibrary/orangelight/blob/50c9cc9e4ea17d102e033507c3f5abf233aaa114/spec/controllers/bookmarks_controller_spec.rb#L22. should just change to contain the quotes?

spec/controllers/bookmarks_controller_spec.rb
```diff
@@ -35,9 +35,11 @@
...
       end
       data2.each do |value|
         expect(body).to include(value)
+        expect(body).to include("'9935444363506421'")
```
carolyncole, line 42: similar comment to above

carolyncole (changes requested):

escowles (approved):
👍🏻 Thanks @christinach !

spec/controllers/bookmarks_controller_spec.rb
christinach: @carolyncole I updated data id to include the single quotes

PR #2710, opened 2021-09-02
Fixes problem in Chicago citations when MARC 502|d is missing

### LINKED ISSUE #1874: Citations link broken for record 9946263
[Orangelight/production] ActionView::Template::Error: undefined method `[]' for nil:NilClass

## Backtrace

    line 79 of [PROJECT_ROOT]/app/models/concerns/blacklight/solr/document/marc.rb: export_as_chicago_citation_txt
    line 12 of [PROJECT_ROOT]/lib/orangelight/middleware/invalid_parameter_handler.rb: call

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/54399/faults/47616219)

hectorcorrea:
Override method `clean_end_punctuation()` to prevent crash in Chicago style citations when field MARC `502|d` is `nil`. 

Fixes #1874 

The culprit is buried in a method inside Blacklight MARC gem ([code](https://github.com/projectblacklight/blacklight-marc/blob/db85fe39922f556c7dd44126865321848e4b882c/app/models/concerns/blacklight/solr/document/marc_export.rb#L303-L305)) that I was able to override thanks to Ruby's built-in prepend feature ([more info](https://stackoverflow.com/questions/5944278/overriding-method-by-another-defined-in-module)). Notice that I disabled  a Rubocop rule in the overridden method so that its code mimics the structure of the [original implementation in Blacklight MARC](https://github.com/projectblacklight/blacklight-marc/blob/db85fe39922f556c7dd44126865321848e4b882c/app/models/concerns/blacklight/solr/document/marc_export.rb#L493-L498).

### DIFF
--- a/spec/models/concerns/blacklight/solr/document/marc_export_override_spec.rb
+++ b/spec/models/concerns/blacklight/solr/document/marc_export_override_spec.rb
@@ -2,15 +2,17 @@
 require 'rails_helper'
 
 RSpec.describe Blacklight::Solr::Document::MarcExportOverride do
-  # See https://mixandgo.com/learn/how-to-test-ruby-modules-with-rspec for information on
-  # testing methods defined in modules.
-  let(:dummy_class) do
-    Class.new do
-      extend Blacklight::Solr::Document::MarcExportOverride
+  describe "#clean_end_punctuation" do
+    # See https://mixandgo.com/learn/how-to-test-ruby-modules-with-rspec for information on
+    # testing methods defined in modules.
+    let(:sample_class) do
+      Class.new do
+        include Blacklight::Solr::Document::MarcExportOverride
+      end
     end
-  end
 
-  it 'handles nil values' do
-    expect(dummy_class.clean_end_punctuation(nil)).to eq ""
+    it 'handles nil values' do
+      expect(sample_class.new.clean_end_punctuation(nil)).to eq ""
+    end
   end
 end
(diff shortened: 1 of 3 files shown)


spec/models/concerns/blacklight/solr/document/marc_export_override_spec.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+require 'rails_helper'
+
+RSpec.describe Blacklight::Solr::Document::MarcExportOverride do
+  # See https://mixandgo.com/learn/how-to-test-ruby-modules-with-rspec for information on
+  # testing methods defined in modules.
+  let(:dummy_class) do
+    Class.new do
```
christinach, line 8: Can you update the word 'dummy_class' with 'sample_class'? There was a discussion a long time ago where dummy was decided as a deprecated word. The proposed inclusive term was 'Placeholder value, sample value'. https://docs.google.com/document/d/1QXxC7hyhjUmsJyrZhkQn9rxv3Y32iIuVXzvqpMg7Ikw/edit#

hectorcorrea: Fixed. Thank you catching this and making me aware of the document with the discussion.

christinach: Thank you @hectorcorrea!  Maybe at some point we can go ahead and fix these terms across orangelight and bibdata.

christinach (approved):

PR #2713, opened 2021-09-08
Change request initializer to match Gem template

carolyncole:
Also update requests to use Factory-bot instead of factory girl

### DIFF
--- a/config/initializers/requests_initializer.rb
+++ b/config/initializers/requests_initializer.rb
@@ -19,7 +19,7 @@ module Requests
       end
 
       begin
-        YAML.safe_load(requests_erb, [], [], true)[Rails.env]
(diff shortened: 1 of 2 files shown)


config/initializers/requests_initializer.rb
```diff
@@ -19,7 +19,7 @@ def config_yaml
       end
 
       begin
-        YAML.safe_load(requests_erb, [], [], true)[Rails.env]
+        YAML.safe_load(requests_erb, aliases: true)[Rails.env]
```
carolyncole, line 22: This change is not necessary, but seems less mysterious than the other syntax.

bess (approved):

bess:
Closing this PR in favor of one that only has the initializer update, not the Gemfile.lock update. The update to the SHA for requests has been superseded.

PR #2717, opened 2021-09-10
Ensuring that the `log` directory is created for deployments

### LINKED ISSUE #2718: Directory log: /opt/appname/shared/log/sneakers.log does not exist
symlink is missing https://github.com/pulibrary/orangelight/blob/a7a071d5fb7e9e40ec650c46f8b6e4d1ada73a9b/config/deploy.rb#L37 for log directory.
Sneakers cannot create log file.

See: https://github.com/pulibrary/pulmap/blob/9c2ea3593466609df7da69ebc7709b896e858cb3/config/deploy.rb#L28

jrgriffiniii:
Ensuring that the `log` directory is created for deployments (this may otherwise block the sneakers workers from running). Resolves #2718

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -34,7 +34,7 @@ shared_path = "#{:deploy_to}/shared"
 # config/redis.yml config/devise.yml config/resque_pool.yml, config/recipients_list.yml, log/resque-pool.stderr.log log/resque-pool.stdout.log
 
 # set :linked_files, %w{config/database.yml}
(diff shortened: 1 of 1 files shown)


jrgriffiniii:
This was tested on the QA environment with a deployment, and it was confirmed with @hectorcorrea and @christinach that this will indeed create the `/opt/orangelight/shared/log` directory as needed. However, in observance of the practice in which team members do *not* deploy on Fridays, this shall be reviewed next week.

christinach (approved):

PR #2740, opened 2021-09-20
Adjusted values to new values now that we have migrated to Alma

### LINKED ISSUE #2608: Stackmap not working for Architecture
Architecture locations test successfully within Stackmap itself, but seem to be failing from the catalog. Example:

From Stackmap test: https://princeton.stackmap.com/view/?callno=NA1080.R85%20B54%202019&location=arch$stacks&library=Architecture%20Library

From the catalog: https://catalog.princeton.edu/catalog/99118878653506421/stackmap?loc=arch$stacks&cn=NA1080.R85%20B54%202019

Locations from other libraries appear to be working.

hectorcorrea:
Fixes #2608 

This is an example of a book in the Architecture Library now displaying the stack map

arch_stacks

### DIFF
--- a/app/services/stackmap_service.rb
+++ b/app/services/stackmap_service.rb
@@ -74,20 +74,20 @@ class StackmapService
       end
 
       def stackmap_libs
-        %w[architecture eastasian engineering lewis mendel plasma stokes]
+        %w[arch eastasian engineer lewis mendel plasma stokes]
       end
 
       def missing_stackmap_reserves
         {
-          'ueso' => 'https://library.princeton.edu/architecture',
-          'piaprr' => 'https://library.princeton.edu/stokes',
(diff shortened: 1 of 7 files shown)


kevinreiss (approved):

kevinreiss:
These look good to me @hectorcorrea, I think one of the failing tests related to the hathiurl can be removed since we deprecated ETAS.

hectorcorrea:
@kevinreiss Thanks for the feedback. I had to update one of the fixtures to include the new Alma library/holding codes (instead of Voyager codes) but that caused a bunch of other tests (unrelated to Stackmap but that use the location codes) to break. Hunting them down and tracing what the equivalent Alma code should be took me a bit longer than I thought but I think this is finally ready.

bess (approved):

PR #2758, opened 2021-09-30
Availability label in Search Results

### LINKED ISSUE #2671: We use two labels for full availability
- [ ] Ask @kevinreiss if we want to do this change now and display only 'View Record for Full Availability' which is basically the label that we use for SCSB.

`View Record for Full Availability` when it is a scsb record or when it has more than 2 holdings. This is added through the application helper.
`View Record for Availability'`. It is added by javascript when the holding uses a 'fake_id' - when it is a temp location.

Since the purpose for both is to visit the record page to see more holdings can we use one label, in preference 'View Record for Full Availability' ?

Screen Shot 2021-08-18 at 4 29 07 PM

hectorcorrea:
Use the same text to let the user that they can view complete availability in the record page regardless of what kind of holding the availability is for. 

With this PR we display "View record for Full Availability" rather than sometimes "View record for Full Availability" and other times "View record for availability"

The following screenshots show before and after. 

## Before
old_display

## After
av_display_new

Fixes #2671

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -236,10 +236,7 @@ module ApplicationHelper
     if scsb_multiple == true
       block << content_tag(:li, link_to('View Record for Full Availability', solr_document_path(document['id']), class: 'availability-icon badge badge-secondary more-info', title: 'Click on the record for full availability info', 'data-toggle' => 'tooltip').html_safe)
     elsif holdings_hash.length > 2
-      block << content_tag(:li, link_to('View Record for Full Availability', solr_document_path(document['id']),
-                                        class: 'availability-icon badge badge-secondary more-info', title: 'Click on the record for full availability info',
-                                        'data-toggle' => 'tooltip').html_safe)
-
+      block << content_tag(:span, "View record for information on additional holdings", "style" => "font-size: small; font-style: italic;")
     elsif !holdings_hash.empty?
       block << content_tag(:li, link_to('', solr_document_path(document['id']),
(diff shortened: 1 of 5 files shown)


hectorcorrea:
@kevinreiss I am 99.9% sure this new text won't affect Request since the change is only for "Search Results", and we don't have a Request button there, but could you confirm?

kevinreiss:
@hectorcorrea I can't link of any reason why this would impact requests. The only feedback I have is I know the third badge is because this item has more than two holdings, it does look a little to have the same wording on all three badges.

christinach:
@hectorcorrea @kevinreiss would it be ok if we remove the third label that indicates that holdings are more than 2 or that it is a scsb?

christinach:
> @hectorcorrea @kevinreiss would it be ok if we remove the third label that indicates that holdings are more than 2 or that it is a scsb?

So in this example in the search results we would display only the 2 badges and not the 3rd one.

kevinreiss:
I think that would make sense.

hectorcorrea:
@kevinreiss @christinach I've replaced the third redundant badge with a simpler text when there are more than two holdings on the record. See updated "before/after" screenshot at the top of this PR.

This way we still display the relevant information to the user but the display is not (as) repeatitive.

I can remote the text altogether if you prefer, though.

christinach:
@hectorcorrea it looks intuitive to me. Can you also test it with a scsb record?

christinach:
It looks good:
Screen Shot 2021-10-01 at 4 28 58 PM

bess (approved):

PR #2776, opened 2021-10-20
IIIF viewer for related records

hectorcorrea:
Adds a separate viewer for a related MMS_ID to handle instances where we only have a manifest in Figgy for a related MMS_ID but not for the current MMS_ID

Fixes https://github.com/pulibrary/bibdata/issues/1659

In the case of the original issue reported the current MMS_ID is `9956597633506421` but in Figgy we only have a manifest for the related MMS_ID `9956630863506421`

The code in this PR will detect the presence of the related MMS_ID in the `electronic_access_1display` field and use that to add an extra viewer for this related MMS_ID.

I still need to add a test for it, but if you want to test it you can import with Traject the attached fixture in your local Solr and view the record in Orangelight.

The attached screenshot how the problematic record now shows the viewer as soon as the record is loaded.

Screen Shot 2021-10-20 at 5 37 15 PM


app/views/catalog/_show_digital_content.html.erb
```diff
@@ -1 +1,7 @@
+<!-- Viewer for the current MMS_ID -->
 <div id="view" class="document-viewers" data-bib-id="<%= @document.id %>"></div>
+
+<!-- Viewer for related MMS_IDs -->
+<% @document.related_bibs_iiif_manifest.each_with_index do |related_bib, ix| %>
```
hectorcorrea, line 5: This new viewer will be picked up by Javascript in https://github.com/pulibrary/orangelight/blob/main/app/javascript/orangelight/orangelight_ui_loader.es6#L37-L43 because it has the `document-viewers` CSS class.

app/models/solr_document.rb
```diff
@@ -106,6 +106,28 @@ def iiif_manifest_uri
...
+    end
+  rescue => ex
+    Rails.logger.error "Error calculating related_bibs_iiif_manifest for #{id}: #{ex.message}"
+    []
+  end
+
+  # Makes sure an ID is an Alma ID or converts it to one if it is not.
+  def ensure_alma_id(id)
+    return id if id.length > 7 && id.start_with?("99")
+    "99#{id}3506421"
```
christinach, line 128: Why are you checking id to be > 7 ? Do you consider a potential id of 99106421?
I don't think we need this method since we have in the controller https://github.com/pulibrary/orangelight/blob/f3877565aa40cc3c744728b054531fd4c80a0723/app/controllers/catalog_controller.rb#L17. 
Also if we keep it can you rename it to: ensure_voyager_to_alma_id . An alma id can also end to 06421 if it was created in Alma.

hectorcorrea: Method has been renamed.

PR #2787, opened 2021-10-27
Concurrent Alma API calls (with smaller batches)

### LINKED ISSUE #2780: Search results batch API calls in groups of 10
The search results page makes a call to bibdata to request availability information for as many records as the user request per page. Our default value is 20 records per page but users can select up to 100 records per page.

We have noticed a significant performance degradation in Alma response times when requests have 50 record per page. ExLibris indicated that we should do batches of 10 instead.

We should update the Search Javascript to batch the IDs in the page in groups of 10 to prevent the performance degradation. Although this will increase the number of API calls that we make to Alma this should be OK since (1) we are way under the limit and (2) ExLibris said we could negotiate this if it becomes an issue.

This will also increase the number of calls to bibdata, that should be OK, but we should keep an eye on it. 

Note: Even if we end-up using SRU as indicated in https://github.com/pulibrary/orangelight/issues/2779 we might still want to batch the SRU calls, in other words, the work in this issue should be a good investment regardless of the way we go.

hectorcorrea:
Make calls to bibdata/Alma in batches of 10 MMS_IDs, instead of using the page size that defaults to 20 but than can be as large as 100. Large batches (50 or 100 MMS_IDs) results in extremely slow response times from Alma.

For the default page size of 20 in Search results the code in this PR will result in two concurrent calls to bibdata (and hence Alma) of 10 MMS_IDs each. It seems that we get better performance this way at the cost of increasing our number of API call to Alma, but since we are way under the limit this is something that ExLibris recommended that we try.

Another thing to watch for is our **concurrent** number of calls to Alma since we are limited to (I believe) 25 calls per second. But again, since our current concurrency is not high this should not be a problem, but we should keep an eye on it after deploying this to production.

Fixes #2780

This PR also takes care of the problem where we were requesting more than 100 MMS_IDs from Alma at once and Alma was rejecting the request (Alma has a hard limit of 100 MMS_IDs). This issue happens if user requests page size of 100 and there are more than 100 **holdings** in the results. Not a common occurrence but something that showed up in our logs a few times.

### DIFF
--- a/app/javascript/orangelight/availability.es6
+++ b/app/javascript/orangelight/availability.es6
@@ -35,45 +35,23 @@ export default class AvailabilityUpdater {
     // a search results page or a call number browse page
     if ($(".documents-list").length > 0) {
       const bib_ids = this.record_ids();
-      if (bib_ids.length < 1) { return; }
-
-      // Default batch size to the largest page size in OL. The number of actual bib_ids
-      // is usually the same as the number of bibs on the page.
-      var batch_size = 100;
-      var urlParams = new URLSearchParams(window.location.search);
-      if (urlParams.has('alma_batch_size')) {
-        // Break down the bibdata/Alma API calls in batches if and only if we have our
-        // super secret query string parameter in the URL.
-        batch_size = parseInt(urlParams.get('alma_batch_size'), 10);
+      if (bib_ids.length < 1) {
+        return;
       }
 
+      const batch_size = 10;
       const batches = this.ids_to_batches(bib_ids, batch_size);
-      console.log(`Batches requested at ${new Date().toISOString()}`);
-      console.log(`size: ${batch_size}, batches: ${batches.length}, ids: ${bib_ids.length}`);
-      for(var i= 0; i < batches.length; i++) {
+      console.log(`Requested at ${new Date().toISOString()}, batch size: ${batch_size}, batches: ${batches.length}, ids: ${bib_ids.length}`);
 
+      for(var i= 0; i < batches.length; i++) {
         var batch_url = `${this.bibdata_base_url}/bibliographic/availability.json?bib_ids=${batches[i].join()}`;
         console.log(`batch: ${i}, url: ${batch_url}`);
-
         $.getJSON(batch_url, this.process_results_list)
-          .fail((jqXHR, textStatus, errorThrown) => {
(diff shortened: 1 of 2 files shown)


app/javascript/orangelight/availability.es6
```diff
@@ -35,25 +35,24 @@ export default class AvailabilityUpdater {
...
+
+      for(var i= 0; i < batches.length; i++) {
+        var batch_url = `${this.bibdata_base_url}/bibliographic/availability.json?bib_ids=${batches[i].join()}`;
+        console.log(`batch: ${i}, url: ${batch_url}`);
+        $.getJSON(batch_url, this.process_results_list)
+          .fail((jqXHR, _textStatus, errorThrown) => {
+            // Log that there were problems fetching a batch. Unfortunately we don't know exactly
+            // which batch so we cannot include that information.
+            console.error(`Failed to retrieve availability data for batch. HTTP status: ${jqXHR.status}: ${errorThrown}`);
+          });
```
hectorcorrea, line 54: Notice that I removed the retry logic for HTTP 429 (too many requests) but this kind of error does not seem to have been triggered (at least it does not show in our logs) and removing it should not be an issue. Additionally, even if we get an error, with this new version the error will be limited to a given batch not to the entire set of MMS_IDs on the page. 

Plus, the batch logic now prevents the of requesting more than 100 MMS_IDs at once from Alma -- and this error was indeed in our logs.

app/javascript/orangelight/availability.es6
```diff
@@ -35,25 +35,24 @@ export default class AvailabilityUpdater {
...
-          }
-          return console.error(`Failed to retrieve availability data for bibs. Records ${bib_ids.join(", ")}: ${errorThrown}`);
-        });
+      if (bib_ids.length < 1) {
+        return;
+      }
+
+      const batch_size = 10;
+      const batches = this.ids_to_batches(bib_ids, batch_size);
+      console.log(`Requested at ${new Date().toISOString()}, batch size: ${batch_size}, batches: ${batches.length}, ids: ${bib_ids.length}`);
```
hectorcorrea, line 44: I left a few `console.log()` in place on purpose to be able to validate the logic in production for a few days after we deploy this to production. We should be able to remove them later on.

christinach (approved):

PR #2819, opened 2021-12-08
2816 host id in holding block

### LINKED ISSUE #2816: Request button and holding block for host records should have the host id and not the record id.
follow up on https://github.com/pulibrary/orangelight/issues/2811
working branch: https://github.com/pulibrary/orangelight/tree/2816-host-id-in-holding-block

christinach:
closes #2816

Adds the host id in each request button so that the button works for host records. 
Fixes the status display in availability.js

### DIFF
--- a/app/javascript/orangelight/availability.es6
+++ b/app/javascript/orangelight/availability.es6
@@ -173,7 +173,7 @@ export default class AvailabilityUpdater {
   // information for a single bib since we are on the Show page. But occasionally the record
   // that we are showing is bound with another (host) record and in those instances
   // `holding_records` has data for two bibs: `this.id` and `this.host_id`.
-  process_single(holding_records) { 
+  process_single(holding_records) {
     this.process_single_for_bib(holding_records, this.id)
     // Availability response in bibdata should be refactored not to include the host holdings under the mms_id of the record page.
     // problematic availability response behaviour for constituent record page with host records. 
--- a/app/javascript/orangelight/availability.es6
+++ b/app/javascript/orangelight/availability.es6
@@ -7,7 +7,6 @@
  * DS207: Consider shorter variations of null checks
  * Full docs: https://github.com/decaffeinate/decaffeinate/blob/master/docs/suggestions.md
  */
-import { thisExpression } from '@babel/types';
 import { insert_online_link } from 'orangelight/insert_online_link'
 
 export default class AvailabilityUpdater {
@@ -215,14 +214,14 @@ export default class AvailabilityUpdater {
 
   update_single(holding_records, id) {
     return (() => {
-      const result = [];
+      let result = [];
       for (let holding_id in holding_records[id]) {
-        const availability_info = holding_records[id][holding_id];
+        let availability_info = holding_records[id][holding_id];
         // case :constituent with host ids.
         // data-record-id has a different this.id when there are host ids.
-        const availability_element = $(`*[data-availability-record='true'][data-record-id='${this.id}'][data-holding-id='${holding_id}'] .availability-icon`);
+        let availability_element = $(`*[data-availability-record='true'][data-record-id='${id}'][data-holding-id='${holding_id}'] .availability-icon`);
         if (availability_info['label']) {
-          const holding_location = $(`*[data-location='true'][data-holding-id='${holding_id}']`);
+          let holding_location = $(`*[data-location='true'][data-holding-id='${holding_id}']`);
           holding_location.text(availability_info['label']);
         }
         this.apply_availability_label(availability_element, availability_info, false);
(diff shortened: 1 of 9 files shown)


app/javascript/orangelight/availability.es6
```diff
@@ -7,6 +7,7 @@
  * DS207: Consider shorter variations of null checks
  * Full docs: https://github.com/decaffeinate/decaffeinate/blob/master/docs/suggestions.md
  */
+import { thisExpression } from '@babel/types';
 import { insert_online_link } from 'orangelight/insert_online_link'
```
christinach, line 11: I dont remember importing this I need to check.

christinach:
Screen Shot 2021-12-08 at 6 53 28 PM

christinach:
Example with more than one host ids: https://catalog-qa.princeton.edu/catalog/9923427953506421

examples provided from @mzelesky 
More than one 773:
https://catalog-qa.princeton.edu/catalog/99676243506421
https://catalog-qa.princeton.edu/catalog/99684283506421
https://catalog-qa.princeton.edu/catalog/99692953506421
https://catalog-qa.princeton.edu/catalog/99740703506421

One 773:
https://catalog-qa.princeton.edu/catalog/9914403506421
https://catalog-qa.princeton.edu/catalog/9947863506421
https://catalog-qa.princeton.edu/catalog/99640703506421

To check the host id -> check the staff_view example: https://catalog-qa.princeton.edu/catalog/9947863506421/staff_view for 773 field(s): **773w = host id**

The catalog record page will include the host as long as the host has holdings.
The request button should direct the user to the correct url with the correct mmsId and mfhd. 
The catalog record page should display the status for the host as well.

example: https://catalog-qa.princeton.edu/catalog/99701333506421/. The record page displays only the holdings of the record id (constituent record). The host record (99124920733506421) has no holdings and the holding does not display.

app/javascript/orangelight/availability.es6
```diff
@@ -86,10 +86,11 @@ export default class AvailabilityUpdater {
       }
     }
   }
-
+  /* example with three host ids: https://bibdata.princeton.edu/bibliographic/availability.json?deep=true&bib_ids=9923427953506421,99125038613506421,99125026373506421,99124945733506421 */
```
carolyncole, line 89: 🎉

app/javascript/orangelight/availability.es6
```diff
@@ -172,10 +173,15 @@ export default class AvailabilityUpdater {
   // information for a single bib since we are on the Show page. But occasionally the record
   // that we are showing is bound with another (host) record and in those instances
   // `holding_records` has data for two bibs: `this.id` and `this.host_id`.
-  process_single(holding_records) {
+  process_single(holding_records) {
```
carolyncole, line 176: I'm guessing this space is unwanted?

mzelesky:
Looks good.

app/services/physical_holdings_markup_builder.rb
```diff
@@ -276,10 +276,21 @@ def self.listify_array(arr)
     arr.join
   end
 
+  def doc_id(holding)
+    host_id = adapter.host_id
```
carolyncole, line 280: Curious here why this is dependent on adapter host_id and not just the holding information.  I might have expected to see something simpler like `holding.dig("mms_id") || adapter.doc_id` replace this entire method.  It looks to me like you set up for that at line 190 of solr_document.

config/requests.yml
```diff
@@ -19,7 +19,7 @@ defaults: &defaults
 development:
   <<: *defaults
   voyager_ub_id: 1@DB
-  bibdata_base: <%= ENV['BIBDATA_BASE'] || "https://bibdata-staging.princeton.edu" %>
+  bibdata_base: <%= ENV['BIBDATA_BASE'] || "https://bibdata.princeton.edu" %>
```
carolyncole, line 22: Was this intended to be committed?  It is what I use locally, so I have not issue with it changing...

carolyncole (changes requested):
Looking great!  Just a couple of questions/ suggestions.

config/requests.yml
christinach: It was unintentional but I can keep the change since we both use it locally.

app/services/physical_holdings_markup_builder.rb
christinach: You are right. I think it was because initially I append the mmsid only in the host holdings and later I decided to append it in the record holding as well.  The holding should include all the mms ids; but I will also leave the one you suggested || adapter.doc_id just to be safe

christinach:
@carolyncole thank you for reviewing I added the updates.

carolyncole (approved):

PR #2829, opened 2022-01-11
2821 facet link

### LINKED ISSUE #2821: The accordion toggles for the filters present their role as "link" instead of button, and do not specify their state.
Assign role="button" and aria-expanded="[true/false]", and make sure the element reacts to spacebar as well as enter.

https://www.w3.org/WAI/WCAG21/quickref/#name-role-value

https://github.com/projectblacklight/blacklight/issues/1905

christinach:
closes #2821


christinach:
Following notes in https://github.com/pulibrary/orangelight/issues/2821#issuecomment-1009118329 and @itmaybejj recommendation `Warning: Be careful when marking up links with the button role. Buttons are expected to be triggered using the Space or Enter key, while links are expected to be triggered using the Enter key. In other words, when links are used to behave like buttons, adding role="button" alone is not sufficient. It will also be necessary to add a key event handler that listens for the Space key in order to be consistent with native buttons.` need to complete the 3rd checkbox before merging.

christinach:
I miss one test for the event handlers function. But I can always add it later.

app/javascript/orangelight/accessible_facets.js
```diff
@@ -0,0 +1,27 @@
...
+    const cardFacets = document.querySelectorAll('#facet-panel-collapse div.card-header.collapse-toggle.facet-field-heading > a')
+    cardFacets.forEach(function(card) {
+      card.addEventListener('keydown', (event) => {
+        // "Spacebar" for IE11 support
+        /* https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code
+           KeyboardEvent: key=' ' | code='Space'
+           KeyboardEvent: key='Enter' | code='Enter'  */
+        // I do check for event.key === 'Enter' event though it is a link
+        if (event.key === " " || event.key === "Enter" || event.key === "Space" || event.key === "Spacebar") {
+          // if space is keydown then do what click does;
```
christinach, line 18: If Space or Enter is keydown

christinach:
chnaging to a draft to refactor a function.

christinach:
Thannk you for testing @carolyncole I'll give it a try.

christinach:
@carolyncole I updated the PR. When tabkey or when letting screen reader to read, it announces if the button is collapsed or expanded. When keydown 'Enter' or Space' it collapses and expands accordingly. I also added role='button' in the facets layout. I commented the other code in case we want to use .js for the role="button". I also added a feature test to show the toggle of aria-expanded attribute. 

https://user-images.githubusercontent.com/9905193/149433961-98e4e417-ef90-4eba-8fa3-f7d9d45e28d7.mov

kevinreiss (approved):

carolyncole (approved):

Issue #2827, opened 2022-01-03
Type ahead matching field type - screen readers only say "edit text" when tabbing into fields that have no associated label 


christinach:
There is a type-ahead item-matching field type found on several pages (Advanced search, Numismatic Search, etc) wherein the fields are not associated with a label, so screen readers only say "edit text" when tabbing into them.

Use the label's for= attribute or the input's aria-labelledby attribute to cross-reference the label.

https://www.w3.org/WAI/WCAG21/quickref/#label-in-name

rladdusaw:
This issue seems to have been resolved already. I was unable to find any fields that did not already have an associated label for the text field.

PR #2834, opened 2022-01-13
Change advanced search facet selection to bootstrap-select for accessibility

carolyncole:
refs #2827 

This is the first page in the ticket.  I took this approach from the Temple advanced search page: https://librarysearch.temple.edu/catalog/advanced

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -57,6 +57,7 @@ gem 'yajl-ruby', '>= 1.3.1', require: 'yajl'
 
 gem 'babel-transpiler'
 gem 'bootstrap', '~> 4.6'
+gem 'bootstrap-select-rails'
 gem 'capybara'
 gem 'coveralls', require: false
 gem 'ddtrace'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -166,6 +166,7 @@ GEM
       autoprefixer-rails (>= 9.1.0)
       popper_js (>= 1.14.3, < 2)
       sassc-rails (>= 2.0.0)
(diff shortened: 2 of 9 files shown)


christinach:
@carolyncole I tested it on staging. The tab key announces as expected. Where I'm confused is when the screen reader goes over the 'Limit results by' for every combo box it says 'list box' then it reads the content and in the end it announces 'Type or select access types, list box pop up, Menu pop-up combobox'. I think this message should come before reading the content? Similar behaviour with the tab key.

carolyncole:
@christinach I updated the PR to not read the select before it is opened.  I believe that makes the most sense.  I also added a feature test for advanced search.

christinach (commented):
@carolyncole the script might have overwritten some css? the search bar with the keyword are dispositioned. Can you check?  I deployed on staging.

carolyncole:
@christinach good catch.  I removed the bootstrap styling...

christinach:
The only difference that I see is that the width of the specific Holding Library field expands based on the content. 
Screen Shot 2022-01-18 at 10 23 17 AM

christinach:
Other than that the PR advances the accessibility and addresses the specific issue.

christinach:
approved by DSSG

christinach (approved):

carolyncole:
@christinach my only comment is that you can use below to not change all the spacing
```
Layout/IndentationConsistency:
  EnforcedStyle: indented_internal_methods
```

PR #2840, opened 2022-01-18
Upgrading ruby to 2.7.5

carolyncole:
Also add capistrano passenger gem


christinach:
@carolyncole I noticed that this PR has the UI advanced search changes. The DSSG meeting is tomorrow afternoon. Since it is in the agenda, we should wait to get a confirmation for the UI change.

christinach:
I tested indexing the subjects table and it looks ok https://catalog-staging.princeton.edu/browse/subjects?search_field=browse_subject&q=

christinach (approved):

Issue #2860, opened 2022-02-17
click View Bookmarks triggers: Blacklight::Exceptions::InvalidRequest


christinach:
reported in the catalog channel: 
I bookmarked a bunch of records in Blacklight, but when I click to view my bookmarks, I get a message that says "sorry, I don't understand your search"

https://app.honeybadger.io/projects/54399/faults/81114343

christinach:
The error happens when the user is logged in or logged out.

carolyncole:
This is the line that we do this: https://github.com/pulibrary/orangelight/blob/main/app/controllers/bookmarks_controller.rb#L20

sandbergja:
closed by #2870

carolyncole:
The maximum ids that allow the page to work is 1024.  We should limit the number of bookmarks to 1024, otherwise we will continue to get this strange error.

kevinreiss:
Failpoint for bookmarks is 1024 items.

kevinreiss:
I think this one can be closed.

PR #2870, opened 2022-02-23
Allow old ids to be update individually

carolyncole:
Currently an entire set of new ids gets added to the old ids and the query is getting too large.
This change only updates the ids that need to be updated, so if a user has 100 bookmarks 100 ids get sent not 200

refs #2860


christinach (approved):

sandbergja (approved):
Thanks for finding this issue, @carolyncole!

PR #2876, opened 2022-02-24
Add subject_facet to catalog controller (get subject browse working again)

sandbergja:
Blacklight now needs facets to be defined in the catalog controller before it can display them. See https://github.com/projectblacklight/blacklight/issues/2401

This facet is linked from the subject browse and from show pages, so we also need to add it to the catalog controller.


kevinreiss (approved):

carolyncole (approved):
Tested with "Undocumented immigrants" on QA.  Looks good to me.  Thanks @sandbergja

PR #2880, opened 2022-02-28
Reorder numismatics fields

### LINKED ISSUE #2826: Reorder fields on Numismatics pages
Reorder fields to match the following screenshot (reviewed and approved by Alan, and then by DSSG on Slack):

141860292-5702896b-e3c9-41a3-9dd1-fb4699d50f02

sandbergja:
This starts to use the new Blacklight DocumentMetadataComponent and the Blacklight Rendering Pipeline (see https://github.com/projectblacklight/blacklight/wiki/value-rendering)

Also adds a new coin fixture

Closes #2826

### DIFF
--- a/app/components/coin_description_component.html.erb
+++ b/app/components/coin_description_component.html.erb
@@ -1,0 +1,7 @@
+<h2>Coin Description</h2>
+<dl class="dl-horizontal dl-invert coin-details">
+<% fields.each do |field| -%>
+  <%= field %>
+<% end -%>
+<%= render 'statement_on_language_in_description' %>
+</dl>
--- a/app/views/catalog/_show_coin_description.html.erb
+++ b/app/views/catalog/_show_coin_description.html.erb
@@ -1,0 +1,2 @@
+<% doc_presenter = show_presenter(document) %>
+<%= render CoinDescriptionComponent.new fields: doc_presenter.field_presenters.select { |field| field.field_config&.coin_description }, show: true %>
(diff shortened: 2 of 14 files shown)


app/views/catalog/_show_coin_description.html.erb
```diff
@@ -0,0 +1,2 @@
+<% doc_presenter = show_presenter(document) %>
+<%= render CoinDescriptionComponent.new fields: doc_presenter.field_presenters.select { |field| field.field_config&.coin_description }, show: true %>
```
carolyncole, line 2: It would be nice for the ruby CoinDescriptionComponent class to do this filtering, vs putting it in the view.

app/components/coin_description_component.html.erb
```diff
@@ -0,0 +1,7 @@
+<h2>Coin Description</h2>
+<dl class="dl-horizontal dl-invert coin-details">
+<% fields.each do |field| -%>
+  <%= field %>
```
carolyncole, line 4: @sandbergja does the data list `<dd> <dt> `fields get output by field?

carolyncole (changes requested):
Nice work!  I just had a question and a stylistic change, but otherwise nice work.  I assume you are leaning into the new blacklight way of doing things!

app/components/coin_description_component.html.erb
sandbergja: Yes!  fields is a ViewComponent "slot", which is filled with `Blacklight::MetadataFieldComponent`s, which output the `<dd> <dt>` fields.  Here is where the slot is defined: https://github.com/projectblacklight/blacklight/blob/v7.22.2/app/components/blacklight/document_metadata_component.rb#L5

carolyncole (approved):

PR #2882, opened 2022-03-02
Patch availability to recognize new temp_location key and make the second call

christinach:
cond call, we need this until we swap the index with the temp locations

### DIFF
--- a/app/javascript/orangelight/availability.es6
+++ b/app/javascript/orangelight/availability.es6
@@ -133,7 +133,7 @@ export default class AvailabilityUpdater {
   // search results
   process_result(record_id, holding_records) {
     for (let holding_id in holding_records) {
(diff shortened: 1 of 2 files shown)


christinach:
it should look both fake_id and the new temp key. Just thought about it. I will update. the indexed records are a mix right now and unitl we swap the index with the temp locations indexed.

christinach:
I deployed in qa. https://catalog-qa.princeton.edu/catalog?utf8=%E2%9C%93&search_field=all_fields&q=

catalog-qa has records from the previous index. for example https://catalog-qa.princeton.edu/catalog/99125410371706421 has holding_id indexed in permanent but the availability response has the new temp key.

christinach:
turning to a draft and deploy to production temporarily

carolyncole (approved):

PR #2884, opened 2022-03-02
Avoid constraint display error when facets applied in certain order

### LINKED ISSUE #2874: Advanced Search fails when right hand side choices are added to an existing search (has facets)
ActionView::Template::Error: undefined method `to_sym' for ["Mendel Music Library", "ReCAP"]:Array
Did you mean?  to_s
               to_set

honeybadger error:
https://app.honeybadger.io/projects/54399/faults/84204121

https://github.com/pulibrary/orangelight/blob/652a275746d36b11f07b10d044259268d97fc2eb/app/helpers/application_helper.rb#L392

def render_location_code(value)
  location = Bibdata.holding_locations[value.to_sym]
  ocation.nil? ? value : "#{value}: #{location_full_display(location)}"
end

sandbergja:
closes #2874 

Here is an easy way to trigger this issue on production:
1. Go to [the production catalog](https://catalog.princeton.edu/)
2. Select a Format Facet
3. Push Edit Search button
4. Select one or many holding locations
5. Push search button

Without the fix, you will get "We're sorry, but there was a problem processing the page you requested."  With the fix, you will at least get results, just with goofy-looking duplicative constraints view.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -389,12 +389,11 @@ module ApplicationHelper
   end
 
   def render_location_code(value)
-    if value.is_a?(Array)
-      value.map { |v| render_location_code(v) }
-    else
-      location = Bibdata.holding_locations[value.to_sym]
-      location.nil? ? value : "#{value}: #{location_full_display(location)}"
+    values = Array(value).map do |loc|
+      location = Bibdata.holding_locations[loc.to_sym]
+      location.nil? ? loc : "#{loc}: #{location_full_display(location)}"
     end
+    values.count == 1 ? values.first : values
   end
 
   def holding_location_label(holding)
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -389,8 +389,12 @@ module ApplicationHelper
   end
 
   def render_location_code(value)
-    location = Bibdata.holding_locations[value.to_sym]
-    location.nil? ? value : "#{value}: #{location_full_display(location)}"
+    if value.is_a?(Array)
+      value.map { |v| render_location_code(v) }
+    else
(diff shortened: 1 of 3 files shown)


app/helpers/application_helper.rb
```diff
@@ -389,8 +389,12 @@ def bibdata_location_code_to_sym(value)
   end
 
   def render_location_code(value)
-    location = Bibdata.holding_locations[value.to_sym]
-    location.nil? ? value : "#{value}: #{location_full_display(location)}"
+    if value.is_a?(Array)
+      value.map { |v| render_location_code(v) }
```
carolyncole, line 393: @sandbergja Recursion is a bit complex in this situation.  I might have just cast the single value to an Array with 
```
Array(value).map do |loc|
   location = Bibdata.holding_locations[loc.to_sym]
   location.nil? ? value : "#{value}: #{location_full_display(location)}"
end
```

spec/system/searching_spec.rb
```diff
@@ -121,4 +121,9 @@
     visit "/catalog?f[subject_facet][]=Japan%E2%80%94History"
     expect(page).to have_content '1 entry found'
   end
+
```
carolyncole, line 124: This test is great.  There is also a test for the helper, which would be a great place to add a test for the syntax of the output https://github.com/pulibrary/orangelight/blob/34bf790fd5fb9aab0cdbbd7f514ff0b5d2c39fa4/spec/helpers/locations_spec.rb#L49

carolyncole (changes requested):
Nice work tracking down a quick fix.  I had a couple comments on the changes.  If making these updates does not make sense because this is not the final code just let me know.

carolyncole (approved):

christinach (approved):

PR #2887, opened 2022-03-03
Migrate constraints to viewcomponents

sandbergja:
Use the new stock blacklight constraints when possible; should help to avoid some issues where the same constraint is repeated multiple times in the "breadcrumbs" section.


sandbergja:
Closing in favor of #3361

PR #2896, opened 2022-03-15
Display indexed public notes for portfolios

sandbergja:


sandbergja:
Per discussion with @kevinreiss, we switched to display the note in parentheses after the link, like this:
portfolio displays link, dates, then a note

I've confirmed that it does display any HTML that staff had added to a note.  For example, this public note included `<br>` tags and an `<a>` tag:


note includes line breaks and a link

christinach:
left side is https://catalog-staging.princeton.edu/catalog/99125446419606421
right side : https://catalog.princeton.edu/catalog/99125446419606421
Screen Shot 2022-03-17 at 12 30 59 PM

sandbergja:
Here are some more IDs:

- 99125209435106421
- 99125209087406421
- 99125207348406421
- 99125204926806421

christinach (approved):

PR #2903, opened 2022-03-22
Adding alma as a login option

carolyncole:
This is working in general.  We plan on demoing this option on Wednesday April 13th at the Circulation Checkin


app/controllers/account_controller.rb
```diff
@@ -62,10 +62,11 @@ def verify_user
 
     def cas_user
```
sandbergja, line 63: Should this method be renamed now, since it could be a CAS or non-CAS user?

app/models/requests/patron.rb
```diff
@@ -170,11 +170,11 @@ def api_request_patron(id:)
       def current_patron(uid)
```
sandbergja, line 170: This method is so much more readable now!  Thank you!

app/views/devise/sessions/new.html.erb
```diff
@@ -1,51 +1,3 @@
...
-          </div>
-        </div>
-      <% end %>
-      <div class='offset-md-2 col-md-8'>
-        <p><%=t('blacklight.login.barcode_help').html_safe%></p>
-      </div>
-    </div>
-  </div>
-</div>
+  <%= render 'shared/login' %>
```
sandbergja, line 2: I really like how you DRYed these views out!

spec/features/requests/request_spec.rb
```diff
@@ -1596,6 +1596,120 @@
...
+      end
+
+      it "allows a physical pickup request of ReCAP Item" do
+        stub_scsb_availability(bib_id: "9941151723506421", institution_id: "PUL", barcode: '32101050751989')
+        visit 'requests/9941151723506421?mfhd=22492702000006421'
+        expect(page).not_to have_content 'Electronic Delivery'
+        expect(page).to have_content 'Physical Item Delivery'
+      end
+
+      it "allows only physical pickup to enumerated annex item" do
```
christinach, line 1625: @carolyncole is this test specific to annex or to annex enumerated item? Will it allow physical pickup if it is annex item and not enumerated?

carolyncole: @christinach this item is enumerated so I put it in the description.  The code should allow access to annex items either enumerated or not.

app/models/requests/request_decorator.rb
```diff
@@ -96,6 +97,11 @@ def holding
       holdings[mfhd]
     end
 
+    def alma_no_services?
```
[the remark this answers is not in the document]
carolyncole: @christinach would the name alma_user_no_services? be better?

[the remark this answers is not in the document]
carolyncole: @christinach this got reworked a bit. Hopefully it makes more sense

christinach (approved):

sandbergja (approved):

PR #2905, opened 2022-03-24
Record page - Add lang attribute in title and vernacular title. (when vernacular exists)

### LINKED ISSUE #1129: Relevant fields need language attribute values
Key fields would be Titles, Publisher, Series, Notes.
Roughly half our entire catalog (i.e., around 6 million non-english book titles) is inaccessible to screen reader users if we don't use lang attributes.

Related to: https://github.com/pulibrary/requests/issues/349

Screen readers need to know the language of text to read it correctly. If we know the language of a record, we should assign that language to the record title. https://www.w3.org/WAI/WCAG21/quickref/#language-of-parts

christinach:
closes #1129


sandbergja (approved):
Looks great!  Thanks, @christinach

PR #2918, opened 2022-04-04
Adding a no services concept to requestable

carolyncole:
This is a step in trying to detangle the router from the the codebase
Also removed the bd_and_ill as they are really one option now.

In general the PR is trying to remove calls to the services array outside of the Requestable.  That way when we add additional constraints based on the logged in user we will only need to add them at the requestable level and everything above it in the UI should just work.  The services array is generated in the router and only considers the item, not who is requesting the item.

This will be necessary when we add the restricted Alma user.

### DIFF
--- a/app/helpers/requests/application_helper.rb
+++ b/app/helpers/requests/application_helper.rb
@@ -33,33 +33,18 @@ module Requests
       if requestable.on_shelf?
         display_on_shelf(requestable, mfhd_id)
       else
-        display_requestable_list(requestable.services)
+        display_requestable_list(requestable)
       end
     end
 
     def show_service_options(requestable, _mfhd_id)
-      if requestable.services.empty?
-        content_tag(:div, "#{requestable.title} #{enum_copy_display(requestable.item)} #{I18n.t('requests.no_services.brief_msg')}", class: 'sr-only') +
-          content_tag(:div, I18n.t("requests.no_services.brief_msg"), class: 'service-item', aria: { hidden: true })
+      if requestable.no_services?
+        content_tag(:div, "#{requestable.title} #{enum_copy_display(requestable.item)} #{ActionView::Helpers::SanitizeHelper.sanitize(I18n.t('requests.no_services.brief_msg'))}", class: 'sr-only') +
+          content_tag(:div, ActionView::Helpers::SanitizeHelper.sanitize(I18n.t("requests.no_services.brief_msg")), class: 'service-item', aria: { hidden: true })
       elsif requestable.charged? && !requestable.aeon? && !requestable.ask_me?
         render partial: 'checked_out_options', locals: { requestable: requestable }
       else
-        display_requestable_list(requestable.services)
-      end
-    end
-
-    def show_service_options_fill_in(requestable)
-      content_tag(:ul, class: "service-list") do
-        brief_msg = if requestable.annex?
-                      I18n.t("requests.annex.brief_msg")
-                    elsif requestable.preservation?
-                      I18n.t("requests.pres.brief_msg")
-                    elsif requestable.services.include? 'recap_no_items'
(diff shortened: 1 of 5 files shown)


app/helpers/requests/application_helper.rb
```diff
@@ -353,28 +336,26 @@ def display_urls(requestable)
...
-          services_to_filter = ["on_shelf_edd", "recap_edd"]
-          if services.include?('bd') && services.include?('ill')
-            services_to_filter << 'bd' << 'ill'
-            concat content_tag(:li, I18n.t("requests.bd_and_ill.brief_msg"), class: "service-item")
-          end
-          filtered_services = services.reject { |val| services_to_filter.include?(val) }
-          filtered_services.each do |service|
-            brief_msg = I18n.t("requests.#{service}.brief_msg")
-            concat content_tag(:li, brief_msg, class: "service-item")
+          if requestable.borrow_direct? || requestable.ill_eligible?
```
christinach, line 342: 👍🏻

app/helpers/requests/application_helper.rb
```diff
@@ -353,28 +336,26 @@ def display_urls(requestable)
 
     private
 
-      def display_requestable_list(services)
-        return if services.blank? # || services.include?('recap_edd') # || services.include?(recap)
+      def display_requestable_list(requestable)
```
christinach, line 339: 👍🏻

app/helpers/requests/application_helper.rb
```diff
@@ -353,28 +336,26 @@ def display_urls(requestable)
...
-            concat content_tag(:li, I18n.t("requests.bd_and_ill.brief_msg"), class: "service-item")
-          end
-          filtered_services = services.reject { |val| services_to_filter.include?(val) }
-          filtered_services.each do |service|
-            brief_msg = I18n.t("requests.#{service}.brief_msg")
-            concat content_tag(:li, brief_msg, class: "service-item")
+          if requestable.borrow_direct? || requestable.ill_eligible?
+            concat content_tag(:li, ActionView::Helpers::SanitizeHelper.sanitize(I18n.t("requests.ill.brief_msg")), class: "service-item")
+          else
+            # there are no instances where more than one actual service is available to an item, so we are going to take the first service that is not edd
```
christinach, line 345: @carolyncole should this comment say: ...is not available to an item..?

app/helpers/requests/application_helper.rb
```diff
@@ -33,33 +33,18 @@ def show_pick_up_service_options(requestable, mfhd_id)
...
+        content_tag(:div, "#{requestable.title} #{enum_copy_display(requestable.item)} #{ActionView::Helpers::SanitizeHelper.sanitize(I18n.t('requests.no_services.brief_msg'))}", class: 'sr-only') +
+          content_tag(:div, ActionView::Helpers::SanitizeHelper.sanitize(I18n.t("requests.no_services.brief_msg")), class: 'service-item', aria: { hidden: true })
       elsif requestable.charged? && !requestable.aeon? && !requestable.ask_me?
         render partial: 'checked_out_options', locals: { requestable: requestable }
       else
-        display_requestable_list(requestable.services)
-      end
-    end
-
-    def show_service_options_fill_in(requestable)
```
christinach, line 51: where was this called? `def show_service_options_fill_in(requestable)`

christinach (commented):
@carolyncole ~~I have a question in a `comment` you've added~~; and one more question that I don't understand where the method `def show_service_options_fill_in` was called. Other than that the specs look ok to me and the Pr looks ok.

app/helpers/requests/application_helper.rb
carolyncole: @christinach No where, which is why I removed it.

PR #2935, opened 2022-04-07
Move build into non parallel steps

carolyncole:


carolyncole:
Got this working, but does not seem to improve time overall.

PR #2945, opened 2022-04-11
load temp holding form and list items

### LINKED ISSUE #2929: Request form errors
https://catalog.princeton.edu/requests/99125453129606421?mfhd=22917265440006421
Screen Shot 2022-03-23 at 12 25 18 PM

If the temporary location is sent as a mfhd the form does not error, but also does not work correctly: https://catalog.princeton.edu/requests/99125453129606421?mfhd=arch$newbook

christinach:
closes #2929

### DIFF
--- a/app/models/requests/request.rb
+++ b/app/models/requests/request.rb
@@ -284,10 +284,14 @@ module Requests
         item_loc = item_current_location(item)
         current_location = get_current_location(item_loc: item_loc)
         item['status_label'] = barcodesort[item['barcode']][:status_label] unless barcodesort.empty?
-        holding = holdings[holding_id] # || holdings[item_loc]
+        calculate_holding = if item["in_temp_library"]
+                              { holding_id.to_sym.to_s => holdings[item_loc] }
+                            else
+                              { holding_id.to_sym.to_s => holdings[holding_id] }
+                            end
         params = build_requestable_params(
           item: item.with_indifferent_access,
-          holding: { holding_id.to_sym.to_s => holding },
+          holding: calculate_holding,
           location: current_location
         )
         # sometimes availability returns items without any status
--- a/app/models/requests/request_decorator.rb
+++ b/app/models/requests/request_decorator.rb
@@ -87,13 +87,22 @@ module Requests
 
     def location_label
       return "" if holding.blank?
-      label = holding["library"]
-      label += " - #{holding['location']}" if holding["location"].present?
+      if any_items? && requestable.first.item["in_temp_library"]
+        label = holding["current_library"]
+        label += " - #{holding['current_location']}" if holding["current_location"].present?
+      else
+        label = holding["library"]
+        label += " - #{holding['location']}" if holding["location"].present?
+      end
       label
     end
 
     def holding
-      holdings[mfhd]
+      if any_items? && requestable.first.item["in_temp_library"]
+        holdings[requestable.first.item["temp_location_code"]]
+      else
+        holdings[mfhd]
+      end
     end
 
     private
--- a/app/models/requests/request_decorator.rb
+++ b/app/models/requests/request_decorator.rb
@@ -87,7 +87,7 @@ module Requests
 
     def location_label
       return "" if holding.blank?
-      if any_items? && requestable.first.item["in_temp_library"]
+      if any_items? && requestable.first.item.temp_loc?
         label = holding["current_library"]
         label += " - #{holding['current_location']}" if holding["current_location"].present?
       else
@@ -98,7 +98,7 @@ module Requests
     end
 
     def holding
-      if any_items? && requestable.first.item["in_temp_library"]
+      if any_items? && requestable.first.item.temp_loc?
         holdings[requestable.first.item["temp_location_code"]]
       else
         holdings[mfhd]
--- a/app/models/requests/requestable/item.rb
+++ b/app/models/requests/requestable/item.rb
@@ -45,7 +45,7 @@ class Requests::Requestable
     end
 
     def temp_loc?
-      self[:temp_loc].present?
+      self[:in_temp_library]
     end
 
     def on_reserve?
@@ -161,7 +161,7 @@ class Requests::Requestable
       end
 
       def temp_loc?
-        false
+        ""
       end
 
       def on_reserve?
--- a/app/views/requests/request/_requestable_list_form.html.erb
+++ b/app/views/requests/request/_requestable_list_form.html.erb
@@ -1,10 +1,10 @@
-<% return unless holdings[mfhd].present? %>
+<% return unless @request.holding.present? %>
 <%= hidden_fields_mfhd holdings[mfhd] %>
 <div class="table-responsive">
   <div id="holding<%= mfhd %>" class="table items table">
     <h2><%= current_location_label(@request.location_label, requestable_list) %> </h2>
 
-      <% unless holdings[mfhd]["location_has"].nil? %>
+      <% unless @request.holding["location_has"].nil? %>
           <div>
             <h3>Location has: </h3>
           </div>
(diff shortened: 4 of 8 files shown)


christinach:
I'm able to place an electronic delivery request.
Screen Shot 2022-04-11 at 5 51 07 PM

christinach:
@carolyncole when you have a chance can you please review this pr?

christinach:
Items in lewis$resterm could be in a temp location. For example: http://localhost:3000/requests/9960861053506421?mfhd=22753114530006421. When you pick an example check the availability response to make sure it is in a temp library.

spec/features/requests/request_spec.rb
```diff
@@ -1283,6 +1283,52 @@
...
+            stub_request(:get, patron_url)
+              .to_return(status: 200, body: responses[:found], headers: {})
+            stub_request(:post, transaction_url)
+              .with(body: hash_including("Username" => "jstudent", "TransactionStatus" => "Awaiting Request Processing", "RequestType" => "Loan", "ProcessType" => "Borrowing", "WantedBy" => "Yes, until the semester's", "LoanAuthor" => "Zhongguo xin li xue hui", "LoanTitle" => "Xin li ke xue = Journal of psychological science 心理科学 = Journal of psychological science", "LoanPublisher" => nil, "ISSN" => "", "CallNumber" => "BF8.C5 H76", "CitedIn" => "https://catalog.princeton.edu/catalog/9941150973506421", "ItemInfo3" => "no.217-218", "ItemInfo4" => nil, "CitedPages" => "COVID-19 Campus Closure", "AcceptNonEnglish" => true, "ESPNumber" => nil, "DocumentType" => "Book", "LoanPlace" => nil))
+              .to_return(status: 200, body: responses[:transaction_created], headers: {})
+            # stub_request(:post, transaction_note_url)
+            #   .with(body: hash_including("Note" => "Loan Request"))
+            #   .to_return(status: 200, body: responses[:note_created], headers: {})
```
sandbergja, line 1299: Can we remove the commented-out stub?

app/models/requests/requestable/item.rb
```diff
@@ -161,7 +161,7 @@ def copy_value
       end
 
       def temp_loc?
-        false
+        ""
```
sandbergja, line 164: Is changing `false` to `""` necessary?

spec/features/requests/request_spec.rb
```diff
@@ -1283,6 +1283,52 @@
...
+            expect(page).to have_content 'Electronic Delivery'
+            expect(page).to have_content "Architecture Library- Librarian's Office NA682.B7 S673 2017b"
+            expect(page).to have_content 'Available - Item in place'
+            expect(page).to have_content 'vol.1'
+            check('requestable_selected_23514405150006421')
+            choose('requestable__delivery_mode_23514405150006421_edd')
+            fill_in 'requestable__edd_art_title_23514405150006421', with: 'some text'
+            expect { click_button 'Request this Item' }.to change { ActionMailer::Base.deliveries.count }.by(1)
+            expect(a_request(:post, transaction_url)).to have_been_made
+            # expect(a_request(:post, transaction_note_url)).to have_been_made
```
sandbergja, line 1315: Can we remove this comment-out line?

app/models/requests/request.rb
```diff
@@ -284,10 +284,14 @@ def build_requestable_mfhd_item(_requestable_items, holding_id, item, barcodesor
...
-        holding = holdings[holding_id] # || holdings[item_loc]
+        calculate_holding = if item["in_temp_library"]
+                              { holding_id.to_sym.to_s => holdings[item_loc] }
+                            else
+                              { holding_id.to_sym.to_s => holdings[holding_id] }
+                            end
         params = build_requestable_params(
           item: item.with_indifferent_access,
-          holding: { holding_id.to_sym.to_s => holding },
+          holding: calculate_holding,
```
sandbergja, line 294: Nice!  I think this looks much cleaner than it did before.

app/models/requests/request_decorator.rb
```diff
@@ -87,13 +87,22 @@ def only_aeon?
...
+      else
+        label = holding["library"]
+        label += " - #{holding['location']}" if holding["location"].present?
+      end
       label
     end
 
     def holding
-      holdings[mfhd]
+      if any_items? && requestable.first.item["in_temp_library"]
```
sandbergja, line 101: Just to keep it consistent with other checks, what about:
```suggestion
      if any_items? && requestable.first.item.temp_loc?
```

sandbergja (changes requested):
Looks great to me; thanks for fixing this, @christinach!  I had a few small questions and suggestions -- nothing that I'd consider a blocker if we'd like to get this in soon.

app/models/requests/request_decorator.rb
christinach: thank you! I updated another one I missed !

app/models/requests/requestable/item.rb
christinach: Previously temp_loc? was checking if [self[:temp_loc].present?](https://github.com/pulibrary/orangelight/pull/2945/files#diff-fc52fa5fdd9a5b0b2265bfd72297d0e560cb1099fbe904a25c5e4adcac488655R48). Based on the api response when there is an item it will always return an 'in_temp_library' either true or false. For a null item we won't get anything. This is why I set it to `""`.

sandbergja (approved):

christinach:
{
"99125452874506421": {
"arch$newbook": {
"on_reserve": "N",
"location": "arch$newbook",
"label": "Architecture Library - New Book Shelf",
"status_label": "Unavailable",
"copy_number": null,
"cdl": false,
"temp_location": true,
"id": "arch$newbook"
}
}
} 

In the request form has Plasma Physics. @kevinreiss tested this. This is a Resource Sharing request and it should have Pickup Location: 'Firestone'. Per @kevinreiss the resource sharing items aren't delivered to branches at the moment.
Another one https://catalog-qa.princeton.edu/requests/99125436761106421?mfhd=22912023020006421 
{
"99125436761106421": {
"arch$newbook": {
"on_reserve": "N",
"location": "arch$newbook",
"label": "Architecture Library - New Book Shelf",
"status_label": "Unavailable",
"copy_number": null,
"cdl": false,
"temp_location": true,
"id": "arch$newbook"
}
}
}

app/views/requests/request/_requestable_list_form.html.erb
```diff
@@ -1,10 +1,10 @@
-<% return unless holdings[mfhd].present? %>
+<% return unless @request.holding.present? %>
```
carolyncole, line 1: You should remove the local from the calling partial since you are no longer utilizing it.  https://github.com/pulibrary/orangelight/blob/main/app/views/requests/request/_request_form.html.erb#L25. and this line should be changed to also utilize the holding method https://github.com/pulibrary/orangelight/blob/3452a91e2fc1d3d2ac9a50c60fbb26faca601690/app/views/requests/request/_requestable_list_form.html.erb#L2. It's a bit odd to be utilizing both...

carolyncole (changes requested):
The hidden fields are utilized during the submission, so having those synced with what is displaying on the form is important.

I'm surprised that we are allowing request for temporary items.  @kevinreiss is this the desired behavior?  I thought Inter Library Loan was the desired option for an item in a temporary location.

carolyncole:
Per discussion in slack.  Even though this location is labeled reserves it is not really a reserved location.

https://catalog-qa.princeton.edu/catalog/9960861053506421

Only updated that is needed is to the hidden fields

christinach:
@carolyncole can you check again?

carolyncole (approved):

PR #2952, opened 2022-04-14
Updating help me get it message to remove migration

### LINKED ISSUE #1123: Strip whitespace from barcodes
## Backtrace

    line 9 of [PROJECT_ROOT]/app/services/bibdata.rb: get_patron
    line 15 of [PROJECT_ROOT]/app/controllers/users/omniauth_callbacks_controller.rb: barcode

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/54399/faults/36109686)

carolyncole:
fixes #1123

I should also note the html_safe and rubocop exception was added to fix:
Screen Shot 2022-04-14 at 9 17 14 AM

### DIFF
--- a/app/models/requests/requestable_decorator.rb
+++ b/app/models/requests/requestable_decorator.rb
@@ -139,7 +139,7 @@ module Requests
             else
               "digital_access"
             end
-      I18n.t("requests.help_me.brief_msg.#{key}_html")
+      I18n.t("requests.help_me.brief_msg.#{key}_html").html_safe # rubocop:disable Rails/OutputSafety
     end
 
     def aeon_url(request_ctx)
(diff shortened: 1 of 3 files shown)


app/models/requests/requestable_decorator.rb
```diff
@@ -139,7 +139,7 @@ def help_me_message
             else
               "digital_access"
             end
-      I18n.t("requests.help_me.brief_msg.#{key}_html")
+      I18n.t("requests.help_me.brief_msg.#{key}_html").html_safe # rubocop:disable Rails/OutputSafety
```
sandbergja, line 142: It seems like the `.html_safe` won't be needed once we get to Rails 7 (see https://github.com/rails/rails/pull/43415).  But this seems like the best approach until then.

sandbergja (approved):

sandbergja:
Thanks, @carolyncole !

Issue #2963, opened 2022-04-21
Tests with Js:true start to fail


christinach:
There was an intermittent failure which was fixed by https://github.com/pulibrary/orangelight/pull/2959

All the viewer tests started to fail yesterday: https://app.circleci.com/pipelines/github/pulibrary/orangelight/2422/workflows/929a7232-7f0d-4892-99f6-43b2b97a1885/jobs/5290/tests

I changed the default time to what it was -60- and also played around with different wait time but nothing seemed to work. There are also two tests that don't check the viewer and are not needed. working branch https://github.com/pulibrary/orangelight/tree/handle-wait-time-viewer

christinach:
In order to see the page errors, in https://github.com/pulibrary/orangelight/blob/5f17327633c93ae6ac08664fad2bedb1e2fd5663/spec/support/capybara_selenium.rb#L66 I  replaced Capybara.javascript_driver = :selenium_headless with Capybara.javascript_driver = :selenium. Then in the console and while the test was still running I could see: `Access to fetch at 'https://figgy.princeton.edu/graphql' from origin 'http://127.0.0.1:50248' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.`

 I found a reference that with Rails 6 there is a change `DNS Guard: hosts whitelist` see : https://getaround.tech/rails-6-unnoticed-features/. 
Following this reference, in https://github.com/pulibrary/orangelight/blob/main/config/environments/test.rb, I set `config.hosts << "https://figgy.princeton.edu/graphql"`, `config.hosts.clear`, `config.hosts << "http://127.0.0.1"` but it didn't seem to work.

PR #2968, opened 2022-04-26
Updates js test in orangelight loader- mark as pending viewer rspec tests- removes redundant tests

christinach:
Add js test in orangelight loader when there is an id it calls buildT…humbnailSet, it calls buildViewers
Update js test when there is no id it doesnt call buildThumbnailSet or buildMonogramThumbnails
Mark as pending viewer rspecs
Removed redundant and incorrect rspecs

part of https://github.com/pulibrary/orangelight/issues/2963

### DIFF
--- a/app/javascript/orangelight/figgy_manifest_manager.js
+++ b/app/javascript/orangelight/figgy_manifest_manager.js
@@ -89,7 +89,6 @@ class FiggyViewer {
     this.element.appendChild(viewerElement)
   }
 }
-
 class FiggyViewerSet {
   constructor(element, query, variables, arks, jQuery) {
     this.element = element
@@ -184,7 +183,6 @@ class FiggyThumbnailSet {
       const bibId = this.constructor.buildBibId(orangelightId)
       this.thumbnails[bibId] = resource.thumbnail
     }
-
     return this.resources
   }
 
@@ -265,7 +263,7 @@ class FiggyManifestManager {
   static buildThumbnailSet($elements) {
     return new FiggyThumbnailSet($elements, loadResourcesByOrangelightIds, window.jQuery)
   }
-
+  // See: https://github.com/pulibrary/orangelight/issues/2967
   static buildMonogramThumbnails($monogramIds) {
     return new FiggyThumbnailSet($monogramIds, loadResourcesByFiggyIds, window.jQuery)
   }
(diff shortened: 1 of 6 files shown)


app/javascript/orangelight/figgy_manifest_manager.js
```diff
@@ -265,8 +263,9 @@ class FiggyManifestManager {
   static buildThumbnailSet($elements) {
     return new FiggyThumbnailSet($elements, loadResourcesByOrangelightIds, window.jQuery)
   }
-
+  // See: https://github.com/pulibrary/orangelight/issues/2967
   static buildMonogramThumbnails($monogramIds) {
+    console.log(loadResourcesByFiggyIds)
```
sandbergja, line 268: Should this log statement be removed?

spec/javascript/orangelight/orangelight_ui_loader.spec.js
```diff
@@ -28,7 +30,26 @@ describe('OrangelightUiLoader', function() {
     const spy2 = jest.spyOn(FiggyManifestManager, 'buildMonogramThumbnails')
     let l = new loader
     l.setup_viewers()
-    expect(spy).not.toHaveBeenCalled()
-    expect(spy2).not.toHaveBeenCalled()
+    expect(spy).toHaveBeenCalledTimes(0)
+    expect(spy2).toHaveBeenCalledTimes(0)
```
sandbergja, line 34: What does this change do?  I would have expected these to be the same as `not.toHaveBeenCalled`.

sandbergja (changes requested):
Those two new jest tests look great, and really nice to have CI working again.  Thanks, @christinach!  Just a small comment and question.

spec/javascript/orangelight/orangelight_ui_loader.spec.js
christinach: @sandbergja thank you . It makes sure that it calls it once or twice etc. Otherwise toHaveBeenCalled is only checking if it is called. It's good to have an idea of how many times it's been called.

app/javascript/orangelight/figgy_manifest_manager.js
christinach: yes, I missed this

christinach:
@sandbergja can you check again?

spec/javascript/orangelight/orangelight_ui_loader.spec.js
sandbergja: 👍

sandbergja (approved):

PR #2976, opened 2022-04-29
removing recap_pf from offsite?

carolyncole:
since recap_pf has to have recap as true to be true it is really a redundant check

### DIFF
--- a/app/models/requests/requestable_decorator.rb
+++ b/app/models/requests/requestable_decorator.rb
@@ -83,7 +83,7 @@ module Requests
     end
 
     def off_site?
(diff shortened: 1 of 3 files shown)


app/models/requests/requestable_decorator.rb
```diff
@@ -83,7 +83,7 @@ def in_library_use_required?
     end
 
     def off_site?
-      recap? || recap_pf? || annex? || item_at_clancy? || held_at_marquand_library?
```
christinach, line 86: 👍🏻

christinach (approved):

Issue #2889, opened 2022-03-09
The maximum bookmark ids that will allow a page to work is 1024


christinach:
if the bookmarks are more than 1024 the page will error. See: https://github.com/pulibrary/orangelight/issues/2860

sandbergja:
It's been backported and released in 7.25.2.

kevinreiss:
Jane will take a look and see if this can split into only a ruby only PR that can be merged with BL 7. Javascript should wait until BL8.

kevinreiss:
Jane tried this without Javascript helpers, didn't work out well.

sandbergja:
For a Blacklight 7-compatible work-in-progress branch, see https://github.com/pulibrary/orangelight/pull/3363.

caroldh:
low priority feature, and this might be resolved with 
blacklight upgreade?

kevinreiss:
@caroldh and @ellen-aa, we occasionally get feedback from users (almost always Subject Librarians) that they've amassed a huge number of bookmarks. Do either of you have a sense if folks are doing this in the course of their research or they are using this feature for work that could be better served by using something like the reporting features that Alma has for PUL collections?

ellen-aa:
@kevinreiss, if you have a few names handy, we could check in with the folks who are amassing a lot of bookmarks. I suspect it's research-related, but it would be good to know for sure.

christinach:
I tested my account in catalog-qa where I have 1299 bookmarks and it works.

PR #3047, opened 2022-06-22
Upgrade blacklight to 7.25.3

sandbergja:
part of working on #2889 

It's currently deployed to staging, but we should generate the browse lists on staging and double check them before merging.


christinach:
The browselists have been populated.

christinach (approved):

PR #3060, opened 2022-07-12
Wrapping text and forcing arrow to the right

### LINKED ISSUE #2844: Update the width of the dropdown in advances search page
so that it does not expand based on the content. We want the text to wrap

rladdusaw:
Resolves #2844


sandbergja (approved):
+1 to not having half of the text off screen!  Thanks @rladdusaw

PR #3061, opened 2022-07-12
i2889 don't allow users to add more than 1000 bookmarks

sandbergja:
Provides an alert if they attempt this:

alert message reads Bookmarks You have exceeded the maximum number of bookmarks! You can only save up to 1000 bookmarks

Includes a second commit to migrate from uglifier to terser, without which this branch does not deploy.


christinach (approved):

spec/models/bookmarks_spec.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+require 'rails_helper'
+
+describe Bookmark do
+  it 'allows user to create 5 bookmarks' do
+    user = user_with_many_bookmarks(5)
+    expect { user.save! }.not_to raise_error(ActiveRecord::RecordInvalid)
+    expect(user.bookmarks.count).to eq(5)
+  end
+  it 'does not allow user to create more than 1000 bookmarks' do
```
maxkadel, line 10: Is it possible to override `MAXIMUM_BOOKMARKS` just for this test, so that we don't have to create so many objects in the database (slowing down the test suite)?

maxkadel (approved):
Looks good! Left a couple comments.

maxkadel (changes requested):
Oops, meant to request changes to be sure my comments are read before merging

christinach:
@maxkadel can you please review again?

maxkadel (approved):
Looks good!

christinach (changes requested):
I did a manual testing of the UI. The PR breaks the position of the dropdown caret icon:
Screen Shot 2022-07-27 at 2 48 27 PM
Screen Shot 2022-07-27 at 2 48 22 PM

christinach:
> I did a manual testing of the UI. The PR breaks the position of the dropdown caret icon: Screen Shot 2022-07-27 at 2 48 27 PM Screen Shot 2022-07-27 at 2 48 22 PM

@sandbergja  I think you need to rebase so that it includes commit https://github.com/pulibrary/orangelight/commit/55665f7dc66e20e89918f32b4c84f3dab35cc133. Then I'm sure it will look fine but would still like to deploy on staging and test it.

christinach:
@sandbergja these are my observations from deploying in catalog-qa and playing around with bookmarks:

1. When I'm logged in and have 1000 bookmarks if I select one at a time I get an alert.  When I close the alert the bookmark stays with a button caption 'Saving...'. Going to a different page removes the 'Saving...'

2. If a user clicks on 'Bookmark all', they'll get the same alert like before and they would need to close it for all the bookmarks that are on the results page. The bookmark button has the 'Saving...' caption, because it can't complete the action to bookmark the record. Going to a different page removes the 'Saving...'

3. If I have already more than 1000 saved bookmarks and I click 'Bookmarks', when I try to visit the bookmarks page I still get the 'Sorry I don't understand your search'. If I try to save a bookmark I get the **1**. If I try to save many bookmarks at once I get the **2**.  In both cases the record cannot be bookmarked and 'Saving...' remains as a caption. Going to a different page removes the 'Saving...
4. The alert looks different than the one you have in the screenshot. I've attached a screenshot.
5. In production there are currently 4 user_ids with more than 1000 saved bookmarks. Bookmarks who belong to a specific user_id that does not exist in Users table should be removed. (This is a different ticket. I believe we have many users that are not in the users table anymore but their bookmarks exist)
The active users that have more than 1000 bookmarks are 3. We should discuss in our meeting to reach out to the users and help them to reduce their bookmarks. I believe as it is right now they can't access the bookmarks page at all.
In order to get the users that have more than 1000 bookmarks I run in catalog1 in the rails console : `Bookmark.group(:user_id).count.select{|k,v| v > 999}`

Screen Shot 2022-07-29 at 2 45 20 PM

Issue #3072, opened 2022-07-15
Provide Record Responses in jsonld


kevinreiss:
Allow the catalog show page controller to respond with jsonld. Initial use case is to allow figgy to consumer the metadata of records in this form when needed. See current bibdata implementation of a jsonld response for an example (ex: https://bibdata.princeton.edu/bibliographic/99108564513506421/jsonld). 

See for motivation and the context. 
https://github.com/pulibrary/figgy/issues/5229
https://github.com/pulibrary/figgy/pull/5230

Bibdata's current implementation of this controller is at: https://github.com/pulibrary/bibdata/blob/main/app/controllers/bibliographic_controller.rb#L118-L122. A version of this class https://github.com/pulibrary/bibdata/blob/e92ba310e4cc859a8b40494eee018f5e06156eb2/app/models/jsonld_record.rb could be used to package up the catalog's JSON documents into jsonld.

maxkadel:
I feel like I'm making fairly good progress on this by hewing pretty closely to the bibdata code. One place I'm confused right now is [local_identifier](https://github.com/pulibrary/bibdata/blob/e92ba310e4cc859a8b40494eee018f5e06156eb2/app/models/jsonld_record.rb#L139). 

From my reading of the code, it looks like it's 
- only displaying cicognara numbers 
  - This is especially confusing because I don't think those identifiers are local?
- local_identifier doesn't seem to be included in https://bibdata.princeton.edu/context.json, which I would expect.

In addition, I can't seem to find any examples of works with cicognara numbers in the catalog, although that may just be because I don't know how to search for them correctly.

maxkadel:
@tpendragon - the above may be a question for you?

PR #3088, opened 2022-07-28
Add jsonld as option for record export

maxkadel:
This code is heavily based on the jsonld code in bibdata - see https://github.com/pulibrary/bibdata/blob/e92ba310e4cc859a8b40494eee018f5e06156eb2/app/models/jsonld_record.rb.

Connected to #3072


maxkadel:
There are three values that I'm not sure about
- local_identifier (currently not implemented)
  - From my reading of the code, bibdata is only displaying cicognara numbers 
  - I can't find an example of a record that uses these, and cicognara numbers don't seem like local_identifiers
  - local_identifier isn't included in https://bibdata.princeton.edu/context.json 
- identifier
  - It looks like the bibdata implementation is only displaying `iiif_manifest_paths`, so that's how I've implemented it here, but I'm not positive that is what's wanted 
- electronic_locations
  - Currently pulling from `electronic_access_1display` and displaying all of them - it does not appear that that's what's happening in bibdata - see, for example, `9946093213506421` https://bibdata.princeton.edu/bibliographic/9946093213506421/jsonld vs the "with a digitized manuscript" example in the tests on this branch.

tpendragon:
@maxkadel 

1. We definitely use local identifier. You can see it imported here: https://figgy.princeton.edu/catalog/c2e1f051-73c6-43ba-a8cb-651a89fdd8f1, and these identifiers are used for synchronizing with Getty for https://cicognara.org

2. `identifier` should be ARKs that are in the 856, I believe? I'm not sure where that'd be in the solr record, exactly. The idea is that if there's already an ARK in there, it should import it. I think it'll only be a IIIF ARK if it was put in there from someone importing it from Figgy. I'm working from memory, but I think there used to be some logic in there to try to pick the right 856 if there were multiple, but I could be wrong.

3. I suspect this is just a side effect of the Alma -> Solr direct pipeline in Bibdata being pretty rough, since it didn't have all the data that the bulk export has. I think it's fine as you have it.

maxkadel:
@tpendragon - am I understanding correctly that right now local_identifier should only include the Cicognara number?

tpendragon (approved):
I tested this in a local Figgy branch and it seems to work great.

kevinreiss (approved):

PR #3099, opened 2022-08-02
Fix bug with Orangelight to Alma login

### LINKED ISSUE #3005: My Account Link Sends Already Logged in Users into a loop
If a user signs into the catalog to create a bookmark or make a request for an item and they then visit the my account link that redirects one to the Alma my account page you get stuck in a loop and are never actually redirecting to the Account page hosted on the Ex Libris site.

maxkadel:
- Document login process
- Test and refactor after_sign_in_path_for

Closes #3005

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -14,14 +14,7 @@ class ApplicationController < ActionController::Base
   protect_from_forgery with: :exception
 
   def after_sign_in_path_for(_resource)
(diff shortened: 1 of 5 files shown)


maxkadel:
NetID and Alma user login work fine.

We do not have credentials to test barcode login, so it has not been tested yet.

kevinreiss (approved):

PR #3110, opened 2022-08-04
Change badge on request form

### LINKED ISSUE #3044: For items that are in resource sharing change the status in the Request form to be Not Available
For items that are in resource sharing change the status in the Request form to be `Not Available`.
Remove the `- Resource sharing Request`.

Screen Shot 2022-06-15 at 1 01 59 PM

maxkadel:
- Remove "Resource Sharing" from badge

Closes #3044

### DIFF
--- a/app/models/requests/requestable_decorator.rb
+++ b/app/models/requests/requestable_decorator.rb
@@ -123,12 +123,16 @@ module Requests
                   else
                     "badge-danger"
                   end
-      status = if requestable.status_label.nil? || requestable.status == requestable.status_label
-                 requestable.status
-               else
-                 requestable.status + ' - ' + requestable.status_label
-               end
+      status = status_for_badge
       content_tag(:span, status, class: "availability--label badge #{css_class}")
+    end
+
+    def status_for_badge
+      if requestable.status_label.nil? || requestable.status == requestable.status_label || requestable.status_label == "Resource Sharing Request"
+        requestable.status
+      else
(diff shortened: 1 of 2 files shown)


app/models/requests/requestable_decorator.rb
```diff
@@ -123,14 +123,18 @@ def status_badge
                   else
                     "badge-danger"
                   end
-      status = if requestable.status_label.nil? || requestable.status == requestable.status_label
-                 requestable.status
-               else
-                 requestable.status + ' - ' + requestable.status_label
-               end
+      status = status_for_badge
```
sandbergja, line 126: Is this assignment needed? (couldn't we just call status_for_badge on the next line?)

app/models/requests/requestable_decorator.rb
```diff
@@ -123,14 +123,18 @@ def status_badge
...
       content_tag(:span, status, class: "availability--label badge #{css_class}")
     end
 
+    def status_for_badge
```
sandbergja, line 130: Thanks for refactoring this out into its own method!

sandbergja (changes requested):

app/models/requests/requestable_decorator.rb
maxkadel: So true. I feel silly.

sandbergja (approved):

sandbergja:
Thanks @maxkadel

PR #3134, opened 2022-08-16
Flipflop with admins

### LINKED ISSUE #3130: Create users with elevated privileges in Orangelight
In order to implement the flipflop gem, we need to be able to protect the dashboard, so that only certain users of the Princeton community (probably starting with only this team) can turn features on and off. To do that, we need a concept of users with elevated privileges, and users without elevated privileges. 

Acceptance:
* Application can determine whether a user is an admin or not, and protect or allow a dummy page on that basis
* We have a way to quickly and easily update who the admins are in a given environment

maxkadel:
To verify this PR 
- Go to catalog-staging (assuming this branch is still deployed)
- log in, and notice if there's a banner with "Flipflop test_header feature is on!" is at the top of the page
- Go to https://catalog-staging.princeton.edu/features and flip the feature the other way (if you can see this page, it means you are an admin)
- Go to the home page and see if the banner is there - it should be the opposite of what it was. If it's not, try a hard refresh. 

Closes #3132 
Closes #3130


sandbergja (commented):
Is it possible to change the term "Active Record" to "Current Status" or similar on the dashboard?  Since it will mainly be devs using this dashboard, it will be easy to confuse with Rails ActiveRecord.

maxkadel:
I think what that is showing *is* the Rails ActiveRecord - you can turn on different strategies, and right now the only two active strategies are active_record and default. If we added another strategy - cookies, sequel, redis, etc., those columns would be added to the dashboard, in the order listed in the `config/features.rb` file

docs/features.md
```diff
@@ -0,0 +1,16 @@
+### Flipflop
+Flipflop is a gem that allows you to use different strategies to turn on and off different features.
+
+The dashboard for flipflop is at `/features`, and only administrators for the application can access it. This dashboard shows each feature, and what it is set to using different strategies. Right now, "Active record" (essentially the database) and "Default" are the only strategies, but in the future we could use strategies such as Redis, Cookies, or Sessions, and if we enable these, they will also appear on the dashboard (for configuration of these strategies see `config/features.rb`). Currently, the "Active record" strategy overrides the "Default" strategy.
+
+Unless the feature impacts a controller or initializer (probably better not to use Flipflop for changes in these areas), it is not be necessary to restart the server or deploy in order for these changes to take effect.
```
sandbergja, line 6:
```suggestion
Unless the feature impacts a controller or initializer (probably better not to use Flipflop for changes in these areas), it is not necessary to restart the server or deploy in order for these changes to take effect.
```

lib/tasks/admins.rake
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+namespace :admins do
+  desc 'Update the list of admin users for the app'
+  task update: :environment do
+    if ENV["ORANGELIGHT_ADMIN_NETIDS"]
```
sandbergja, line 5: Now that I'm looking more closely, do we even need to record the admin status in the database in the first place?  It seems like there is potential for the envvar and the database to get out of sync if folks forget to run this rake task.

sandbergja (changes requested):
I like this gem a lot, @maxkadel ; thanks for implementing it.  I have a question about this PR's approach, and one small grammar fix.

lib/tasks/admins.rake
maxkadel: Yeah, I was trying to think about how to make sure they don't get out of sync as well. What if we added it as a monthly cron job? Part of what I like about adding it in the database, is that it makes it easy to set / unset your own admin status for testing.

sandbergja: Can we accomplish the same with, say, `ORANGELIGHT_ADMIN_NETIDS='netid1 netid2' rails s` locally?

Or on a staging box, changing the app_config and reloading passenger?

sandbergja (approved):
Thanks, @maxkadel

PR #3135, opened 2022-08-16
RES_SHARE location always unavailable

### LINKED ISSUE #3003: Review availability for item in RES_SHARE$IN_RS_REQ. It should be unavailable
https://catalog.princeton.edu/catalog/99118399983506421

https://bibdata.princeton.edu/bibliographic/99118399983506421/holdings/22555936970006421/availability.json

### LINKED ISSUE #3043: In search results change the availability label to 'Unavailable' for resource sharing location
In the search results change the availability label to red 'Unavailable' for 'RES_SHARE$IN_RS_REQ'

rladdusaw:
Resolves #3003 
Resolves #3043

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -229,9 +229,9 @@ module ApplicationHelper
       end
 
       block << if holding['temp_location_code'].present?
-                 content_tag(:li, info.html_safe, data: { availability_record: check_availability, record_id: document['id'], holding_id: id, temp_location_code: holding['temp_location_code'], aeon: aeon_location?(location), bound_with: document.bound_with? })
+                 content_tag(:li, info.html_safe, class: 'holding-status', data: { availability_record: check_availability, record_id: document['id'], holding_id: id, temp_location_code: holding['temp_location_code'], aeon: aeon_location?(location), bound_with: document.bound_with? })
                else
-                 content_tag(:li, info.html_safe, data: { availability_record: check_availability, record_id: document['id'], holding_id: id, aeon: aeon_location?(location), bound_with: document.bound_with? })
+                 content_tag(:li, info.html_safe, class: 'holding-status', data: { availability_record: check_availability, record_id: document['id'], holding_id: id, aeon: aeon_location?(location), bound_with: document.bound_with? })
                end
 
       cdl_placeholder = content_tag(:span, '', class: 'badge badge-primary', 'data-availability-cdl' => true)
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -228,11 +228,7 @@ module ApplicationHelper
         info << content_tag(:div, search_location_display(holding, document), class: 'library-location', data: { location: true, record_id: document['id'], holding_id: id })
       end
 
-      block << if holding['temp_location_code'].present?
-                 content_tag(:li, info.html_safe, class: 'holding-status', data: { availability_record: check_availability, record_id: document['id'], holding_id: id, temp_location_code: holding['temp_location_code'], aeon: aeon_location?(location), bound_with: document.bound_with? })
-               else
-                 content_tag(:li, info.html_safe, class: 'holding-status', data: { availability_record: check_availability, record_id: document['id'], holding_id: id, aeon: aeon_location?(location), bound_with: document.bound_with? })
-               end
+      block << content_tag(:li, info.html_safe, class: 'holding-status', data: { availability_record: check_availability, record_id: document['id'], holding_id: id, temp_location_code: holding['temp_location_code'], aeon: aeon_location?(location), bound_with: document.bound_with? }.compact)
 
       cdl_placeholder = content_tag(:span, '', class: 'badge badge-primary', 'data-availability-cdl' => true)
       block << content_tag(:li, cdl_placeholder.html_safe)
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -228,13 +228,11 @@ module ApplicationHelper
         info << content_tag(:div, search_location_display(holding, document), class: 'library-location', data: { location: true, record_id: document['id'], holding_id: id })
       end
 
-      temp_location_code = holding['temp_location_code']
-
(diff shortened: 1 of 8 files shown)


christinach (changes requested):
I will add some specific comments in the pr but for now I will block it because it breaks the availability status for other locations  and shows as unavailable items that are available. 
We also want a test for this location. Maybe update the existing one https://github.com/pulibrary/orangelight/blob/70f77abb34bfe0401d7b04ba375daca8b8c35df3/spec/javascript/orangelight/availability.spec.js#L548

as an example check record 99101035463506421

app/javascript/orangelight/availability.es6
```diff
@@ -194,7 +201,14 @@ export default class AvailabilityUpdater {
         const { label, cdl, temp_location } = holding_records[id][holding_id];
         // case :constituent with host ids.
         // data-record-id has a different this.id when there are host ids.
-        let availability_element = $(`*[data-availability-record='true'][data-record-id='${id}'][data-holding-id='${holding_id}'] .availability-icon`);
+        let availability_element;
+
+        // If we are not getting holding info select the availavility element by record id only.
```
christinach, line 206: This comment can be misleading. We are getting holding info. We dont get an actual holding_id in the availability response. We get the temporary location code as a holding_id. But the rest is holding information. 
```
  "99125473625906421": {
    "RES_SHARE$IN_RS_REQ": {
      "on_reserve": "N",
      "location": "RES_SHARE$IN_RS_REQ",
      "label": "Resource Sharing Library - Lending Resource Sharing Requests",
      "status_label": "Unavailable",
      "copy_number": null,
      "cdl": false,
      "temp_location": true,
      "id": "RES_SHARE$IN_RS_REQ"
    }
  }
}
```
the `data-holding-id='${holding_id}' in the DOM is coming from https://catalog-qa.princeton.edu/catalog/99125473625906421/raw specifically the holdings_1display field. Then we do the match with the availability response. 
The specific location `RES_SHARE$IN_RS_REQ` even though it is a temporary location (has the location code as a holding id in the availability response), it is indexed as a permanent location (check the holdings_1display). This is why there is no direct much. 
To update the DOM we use the https://github.com/pulibrary/orangelight/blob/main/app/services/physical_holdings_markup_builder.rb . There is some code to handle the temporary location and set the holding_id for the request button so that the user can load the request form. https://github.com/pulibrary/orangelight/blob/70f77abb34bfe0401d7b04ba375daca8b8c35df3/app/services/physical_holdings_markup_builder.rb#L285

Maybe try to update the `holding_id` in the element with class holding-status , with the value of the temporary location code only when it is a `RES_SHARE$IN_RS_REQ`. I haven't tested this locally but let me know if you want me to take a look. https://github.com/pulibrary/orangelight/blob/70f77abb34bfe0401d7b04ba375daca8b8c35df3/app/services/physical_holdings_markup_builder.rb#L49

christinach: I meant to paste this line with class: holding-status. https://github.com/pulibrary/orangelight/blob/70f77abb34bfe0401d7b04ba375daca8b8c35df3/app/services/physical_holdings_markup_builder.rb#L61

spec/features/availability_spec.rb
```diff
@@ -27,6 +27,21 @@
     end
   end
 
+  describe 'Holdings for items with temp location', js: true do
```
christinach, line 30: @rladdusaw there is a test in jest that needs to be updated. I would update the jest one and remove the availability_spec.rb.

in https://github.com/pulibrary/orangelight/blob/main/spec/javascript/orangelight/availability.spec.js#L548-L579
update the inner_html with:
```
<td class="holding-status" data-availability-record="true" data-record-id="99125535710106421" data-holding-id="22936525030006421" data-aeon="false" data-temp-location-code="RES_SHARE$IN_RS_REQ">
<span class="availability-icon badge badge-danger" title="Unavailable">Unavailable</span>
</td>
```
and the rest of the values accordingly.

christinach (changes requested):

app/helpers/application_helper.rb
```diff
@@ -227,7 +227,12 @@ def holding_block_search(document)
         end
         info << content_tag(:div, search_location_display(holding, document), class: 'library-location', data: { location: true, record_id: document['id'], holding_id: id })
       end
-      block << content_tag(:li, info.html_safe, data: { availability_record: check_availability, record_id: document['id'], holding_id: id, aeon: aeon_location?(location), bound_with: document.bound_with? })
+
+      block << if holding['temp_location_code'].present?
+                 content_tag(:li, info.html_safe, class: 'holding-status', data: { availability_record: check_availability, record_id: document['id'], holding_id: id, temp_location_code: holding['temp_location_code'], aeon: aeon_location?(location), bound_with: document.bound_with? })
+               else
+                 content_tag(:li, info.html_safe, class: 'holding-status', data: { availability_record: check_availability, record_id: document['id'], holding_id: id, aeon: aeon_location?(location), bound_with: document.bound_with? })
+               end
```
sandbergja, line 235: Can we handle the potential for a blank `holding['temp_location_code']` without adding this repetitive conditional?  Either by teaching the js to handle a nil value in data-temp-location-code or something like:

```
block << content_tag(:li, info.html_safe, class: 'holding-status', data: { availability_record: check_availability, record_id: document['id'], holding_id: id, temp_location_code: holding['temp_location_code'], aeon: aeon_location?(location), bound_with: document.bound_with? }.compact)
```

(Adding `.compact` should remove `temp_location_code` from the data hash if it is nil.

sandbergja (changes requested):
Looks good; thanks @rladdusaw.  Just one style request.

app/helpers/application_helper.rb
rladdusaw: Done.

sandbergja (approved):
Looks great, @rladdusaw !  Thank you for your work on this.

christinach (approved):
@rladdusaw thanks for the pr. I've tested various cases and it looks good to me.

PR #3147, opened 2022-08-18
i1063: unAPI <abbr> element should not be the child of a <ul> element

### LINKED ISSUE #1063: Move <abbr> from unapi implementation in catalog/_index_default out of ul
Violation: `ul` and `ol` must only directly contain `<li> <script> or <template>` elements

Can we move this [abbr tag](https://github.com/pulibrary/orangelight/blob/1e937344e629592d7122dc2933cb70d4bace1d94/app/views/catalog/_index_default.html.erb#L4) elsewhere in the record-wrapper? It cannot be a direct child of `ul`

sandbergja:
closes #1063

### DIFF
--- a/spec/views/catalog/index_spec.rb
+++ b/spec/views/catalog/index_spec.rb
@@ -27,6 +27,16 @@ RSpec.describe 'catalog/index' do
     end
   end
 
+  describe 'search results' do
+    it '<ul> elements contain only <li>, <script>, or <template>' do
+      visit '/catalog?q=korea'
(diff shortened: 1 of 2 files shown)


spec/views/catalog/index_spec.rb
```diff
@@ -27,6 +27,16 @@
     end
   end
 
+  describe 'search results' do
```
maxkadel, line 30: Very cool test!

maxkadel (approved):

PR #3152, opened 2022-08-22
Deprecate BorrowDirect api

### LINKED ISSUE #3093: Deprecate Integration with Borrow Direct API for "Resource Sharing" requests
This API will be deprecated when the current Relais Borrow Direct platform is sunset. Our current Resource Sharing integration works attempts to place a request in Borrow Direct via API when an item is in a state that makes it eligible for resource sharing. If that attempt does not succeed we place an illiad resource sharing request for the patron for the item. When we transition to ReShare will need to deprecate that first attempt to request by API since no current equivalent exists in ReShare's suite of APIs. All resource sharing requests will be place in ILLiad by default. Resource Sharing staff will need to update their workflows to check ReShare for items that are lendable by Borrow Direct.

maxkadel:
Closes #3093


maxkadel:
With a branch deployed to Staging, with some headers included for testing (since removed):

main - flipflop feature on
requests_on_main_no_flipflop

this branch - flipflop feature on
requests_on_branch_flipflop_on

this branch - flipflop feature off
requests_on_branch_flipflop_off

kevinreiss (approved):
Looks @maxkadel. Thanks for re-organizing the specs as well. I was able to submit a number of test requests successfully in staging.

PR #3156, opened 2022-08-25
Manage unauthorized requests to SCSB

### LINKED ISSUE #3142: Route Requests that have a SCSB API Error to ReCAP Problems
Currently when the availability call for SCSB does not provide any data we assume the items are "in-process". In-process means the items are filed in a ReCAP holding location but are not yet accessioned into the facility. We need to add a case that accounts for a scenario when the SCSB availability API produces an error. These items should not be treated as "in-process", but should instead be sent into our triage procedure for SCSB errors. For items in this case we accept the user's request for the item but route it to the recapproblems email address. This points in the code are useful reference points to track current behavior and see where we can make adjustments.

https://github.com/pulibrary/orangelight/blob/e9554b8bac6f51d08e4e4741aefb8e6704301b3b/app/models/requests/request.rb#L243

https://github.com/pulibrary/orangelight/blob/e9554b8bac6f51d08e4e4741aefb8e6704301b3b/app/models/requests/request.rb#L221

https://github.com/pulibrary/orangelight/blob/ddb644404b5471b5c7d263efc9f62be017ef0b1e/app/models/concerns/requests/scsb.rb#L54

maxkadel:
Closes #3142 

If the application can't connect to SCSB, it should say that the item is 'Not Available", and send the user to resource sharing


maxkadel:
Bad key:
image

Good key:
image

spec/cassettes/good_ol_good_bibdata_scsb_key.yml
```diff
@@ -0,0 +1,288 @@
...
+      string: '{"bibliographicId":".b106574619","institutionId":"NYPL"}'
+    headers:
+      User-Agent:
+      - Faraday v0.17.5
+      Content-Type:
+      - application/json
+      Accept:
+      - application/json
+      Api-Key:
+      - TESTME
```
carolyncole, line 263: ❤️

maxkadel: VCR has caught me out in the past, so now I know to check :D

spec/models/concerns/requests/scsb_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+      source: 'pulsearch',
+      mfhd: nil,
+      patron: patron
+    }
+  end
+  let(:request_scsb) { Requests::Request.new(params) }
+  let(:first_item) { request_scsb.holdings["8076325"]["items"][0] }
+  let(:second_item) { request_scsb.holdings["8076325"]["items"][1] }
+
+  context 'with a good scsb key', vcr: { cassette_name: 'good_ol_good_bibdata_scsb_key', record: :none } do
```
christinach, line 28: @maxkadel thank you.
Instead of bad and good can we use authorized and unauthorized?

maxkadel: Thanks for looking it over - I've changed to authorized / unauthorized.

kevinreiss (changes requested):
@maxkadel this looks good but we should trace through one more scenario to see if we need to make a change somewhere. https://catalog-staging.princeton.edu/requests/99122304923506421?mfhd=22511126440006421 is an item that lists as "item in place" in Alma but is not accessioned into the SCSB UAT environment. On the request form we are producing a green badge that says Available - Not Available. This may be just another weird edge case or turn into another ticket.

kevinreiss (approved):

PR #3167, opened 2022-08-31
Removed barcode login button

### LINKED ISSUE #3107: Remove Barcode Login Option from the options to access "My Account"
"Barcode" login users cannot access the Alma Library Account available behind the "Library Account" link in the user menu on the catalog. We should remove that one as a choice when the user selects that link. 

Screen Shot 2022-08-03 at 1 01 05 PM

Barcode users can still access bookmarks and search history so that option should still be available on those menu choices until #2902 is completed.

rladdusaw:
Resolves #3107

### DIFF
--- a/app/views/shared/_login.html.erb
+++ b/app/views/shared/_login.html.erb
@@ -1,4 +1,3 @@
- <% do_not_show_barcode ||= false %>
  <div class='card border-0'>
     <div class='card-body text-center'>
       <%= link_to t('blacklight.login.netid_login_msg'), main_app.user_cas_omniauth_authorize_path, :class => 'btn btn-primary' %>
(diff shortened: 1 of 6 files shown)


app/views/shared/_login.html.erb
```diff
@@ -1,19 +1,10 @@
- <% do_not_show_barcode ||= false %>
  <div class='card border-0'>
     <div class='card-body text-center'>
       <%= link_to t('blacklight.login.netid_login_msg'), main_app.user_cas_omniauth_authorize_path, :class => 'btn btn-primary' %>
       <hr>
       <p class="or-divider">
       or
       </p>
-      <% unless do_not_show_barcode %>        
-      <p><a role="button" data-toggle="collapse" href="#collapseBarcode" aria-expanded="false" aria-controls="collapseBarcode" class="btn btn-outline-secondary">
```
sandbergja, line 10: We should also remove the collapseBarcode section later in this partial.

sandbergja (changes requested):
Thanks @rladdusaw !  Apologies if this is scope creep, but should we also take this opportunity to remove all the back-end code related to barcode authentication?  Things like:
- [barcode-related code in the User model](https://github.com/pulibrary/orangelight/blob/main/app/models/user.rb#L55-L67)
- the [barcode omniauth strategy](https://github.com/pulibrary/orangelight/blob/main/lib/omniauth/strategies/barcode.rb)
- possibly requests-related code?  I'm not sure if there is functionality specific to barcode users that doesn't also apply to alma users

Whether or not the back-end code removal fits into the current work, I also added a comment for some more view-layer stuff that can go.

sandbergja (approved):
Thanks @rladdusaw!

Issue #3175, opened 2022-09-07
JSON-LD not displaying "type" from "genre_facet" field


tpendragon:
### Expected behavior
When I go to https://catalog.princeton.edu/catalog/9968663853506421.jsonld I should see a "type" key that says "Maps", similar to https://bibdata.princeton.edu/bibliographic/9968663853506421/jsonld

### Actual behavior
There's no "type" key.

### Impact of this bug
I can't migrate Figgy's record fetching to pull from Orangelight (and thus not use up Alma API limits) until this is done.

### Implementation notes, if any

I suspect this is because `genre_facet` isn't a stored field, so you don't have access to it. I'm not sure if the best bet is to index it in a different field or if there's something you can do that won't require a full reindex.

maxkadel:
It looks like for the different vocabularies there is a different solr field for the facets - https://github.com/pulibrary/bibdata/blob/7c3e63c479ac2f9c8a59b53cfdbe878763ecf5bf/marc_to_solr/lib/traject_config.rb#L869-L887

Should we only map LC genre terms? What if there are terms from multiple vocabularies (it looks like it's not a repeatable field)?

maxkadel:
Closed by PR #3177

PR #3177, opened 2022-09-07
Map genre from LC for jsonld

maxkadel:
Connected to #3175


tpendragon:
This should work!

sandbergja (approved):
Thanks @maxkadel

PR #3185, opened 2022-09-08
Remove constituent part display

### LINKED ISSUE #3082: Constituent part(s) and Related record(s) fields largely duplicate one another
**Expected behavior**
Each constituent record is listed once on a host record's show page

**Actual behavior**
The first ten constituent records are listed twice on a host record's show page: first as a link in the "Related record(s)" field, then as plain text in the "Constituent part(s)" field.

**Steps to replicate**

Go to https://catalog.princeton.edu/catalog/99124945733506421 and look at the two fields.

**Impact of this bug**

More cluttered look on the show page.  For records with a viewer, this pushes the viewer even farther down below the fold.

**Additional info**

Both are taken from the 774 field:

- related record field uses the MMS ID from the 774$w in a solr call to fetch the title from the constituent record
- Constituent parts field uses the metadata directly from the 774 in the host record

maxkadel:
- Duplicative of Related record(s)

I'm not sure this is what we want, but it's a try.

Closes #3082 

image

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -374,7 +374,6 @@ class CatalogController < ApplicationController
     config.add_show_field 'other_title_display', label: 'Other title(s)'
     config.add_show_field 'other_title_1display', hash: true
     config.add_show_field 'in_display', label: 'In'
-    config.add_show_field 'constituent_part_display', label: 'Constituent part(s)'
     config.add_show_field 'other_editions_display', label: 'Other editions'
     config.add_show_field 'isbn_display', label: 'ISBN'
     config.add_show_field 'issn_display', label: 'ISSN'


maxkadel:
This currently does not work for SCSB records, since they do not have their local id in the `related_record_s` field. For example:

* https://catalog.princeton.edu/catalog/SCSB-9588344
```
related_record_s: [
  "(OCoLC)1057895563",
  "(OCoLC)1080520347  "
],
...
constituent_part_display: [
  "Wadsworth, Alexander, 1806-1898. New plan of the city of Boston",
  "City of Charlestown, 1866. 1 map"
]
```

catalog-staging with PR at this point:
image

christinach:
in production there are 9 records.
SCSB-9588344,
SCSB-12563166,
SCSB-5715787,
SCSB-11225801,
SCSB-11537300,
SCSB-6740416,
SCSB-12798531,
SCSB-9401911,
SCSB-8915880

christinach:
Is there a way to create a scsb partial and render the field only if the document id is a scsb id?

christinach (approved):
In our stand up we decided that we can remove this display considering the small number of SCSB records.

Issue #3210, opened 2022-09-22
Add expand/contract option for lengthy content notes. 


kevinreiss:
Now that https://github.com/pulibrary/bibdata/pull/1998 has been merged. We should introduce an expand/collapse option with similar behavior to what was add for related records in.

### Current display
https://catalog.princeton.edu/catalog/99116299583506421 is a record with a very long contents notes. 

### Desired display
https://github.com/pulibrary/orangelight/commit/145b9ef32dd78a8312fdc5518d8da1b3a5ba11d9 is the PR where the related records expand/contract was introduced. We display three items by default.

PR #3215, opened 2022-09-28
Refactor show page to use Blacklight Rendering Pipeline

sandbergja:
Using the [Blacklight Rendering Pipeline](https://github.com/projectblacklight/blacklight/wiki/value-rendering) to clean up the show and show_top_fields views.

In preparation for some work on #3210


maxkadel (approved):
Looks great! Thank you Jane!

PR #3221, opened 2022-10-04
Send aeon within request

maxkadel:
- I'm concerned this may not be as secure as we need

### DIFF
--- a/app/controllers/requests/request_controller.rb
+++ b/app/controllers/requests/request_controller.rb
@@ -5,7 +5,7 @@ include Requests::ApplicationHelper
 
 module Requests
   class RequestController < ApplicationController
-    before_action :authenticate_user!, except: [:index]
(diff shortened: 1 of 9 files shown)


kevinreiss (approved):

kevinreiss:
Hi, I have reviewed the major scenarios here in qa

1. Aeon Holding Single Item
2. Aeon Holding Multiple items
3. ReCAP Requests
4. On campus pick-up requests
5. Numismatics Record

We don't have senior theses or SCSB records in QA so I can't fully vet those, but looking over the code I can't think of any side effects that we might encounter in those scenarios. I think this PR addresses the major issue we have in production and if one calls the form with the "wrong" aeon parameter, i.e. sending aeon=true to a non-aeon item the form fails gracefully and provides the user with no actionable form actions.

kevinreiss:
I tested all these scenarios as both an unauthenticated and an authenticated user.

PR #3260, opened 2022-11-16
Add Firestone to Stackmap using feature flipper

### LINKED ISSUE #3254: Point Firestone Shelf Locations at Stackmap
In order to test new floorplans in Firestone we need to have the Stackmap Service class return links to stackmap rather than the Firestone locator. When you click a "Where to Find It" link in the catalog you will be sent to stackmap. This is a change we likely want to put behind the feature flipper, since we will need to allow for testing of the data in staging/qa. We should not have to adjust the parameters passed over to the stackamp. 

# implementation note
I think we can eliminate the else statement here:
https://github.com/pulibrary/orangelight/blob/e63bc13205475da3bcfa967b706d3557f7955e9f/app/services/stackmap_service.rb#L24

maxkadel:
Closes #3254

### DIFF
--- a/app/services/stackmap_service.rb
+++ b/app/services/stackmap_service.rb
@@ -74,7 +74,11 @@ class StackmapService
       end
 
       def stackmap_libs
-        %w[arch eastasian engineer lewis mendel plasma stokes]
+        if Flipflop.firestone_locator?
+          %w[arch eastasian engineer lewis mendel plasma stokes]
+        else
+          %w[arch eastasian engineer lewis mendel plasma stokes firestone]
+        end
       end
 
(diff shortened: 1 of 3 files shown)


kevinreiss (approved):

kevinreiss:
@maxkadel when we deploy this to a branch does it default to "on" or "off"?

maxkadel:
> @maxkadel when we deploy this to a branch does it default to "on" or "off"?

@kevinreiss it defaults to "on", meaning it defaults to using the old Firestone locator when it's deployed, and has to be intentionally flipped to make it use Stackmap for Firestone.

PR #3261, opened 2022-11-16
Added mms id and barcode to clancey errors

### LINKED ISSUE #3248: Report Record Details when a Clancy Error is encountered.
Right now when one gets an error from a Clancy request the error template that goes to recapproblems reports only the patron details of the failed requests not the MMS ID or barcode of the item that was requested. Add this data to the email template.

Current display (with patron info cut off):

Screen Shot 2022-11-07 at 1 54 04 PM

rladdusaw:
Resolves #3248


sandbergja (approved):
Thanks, @rladdusaw !

PR #3266, opened 2022-11-21
Disabling submit button until pickup selected

### LINKED ISSUE #3229: "Item does not have a delivery mode" - being logged as error, but not displayed to users
### Expected behavior
When a user tries to request an item without selecting required delivery options, a flash message should display that they need to choose one, and a request should not be sent out until they do so. The request should be intercepted before it raises a server error. 

### Actual behavior
When you click the "Request this Item" button without selecting a delivery mode, it reloads the page, but does not show any flash message to show why your request did not go through. It raises a server error that is logged in Honeybadger.

### Steps to replicate
* Log in
* Go to https://catalog.princeton.edu/requests/9923208643506421?mfhd=22670348320006421
* Click "Request this Item"
* Note that it keeps you on the same page, without a flash error or success message
* Go to Honeybadger, you should be able to see a newly raised error there

### Impact of this bug
* Patrons are not guided to fix what is missing in the form, potentially causing frustration and confusion
* Excessive errors are raised

https://app.honeybadger.io/projects/54399/faults/87665915

```
Exception: Requests::Submission: Item does not have a delivery mode
  [PROJECT_ROOT]/app/models/requests/submission.rb:152 :in `categorize_by_delivery_and_location`
          item["type"] = "annex"
        elsif off_site?(library_code)
          raise Exception, "#{self.class}: Item does not have a delivery mode" unless delivery_mode(item)
          item["type"] = library_code
          item["type"] += "_edd" if edd?(item)
```

rladdusaw:
Resolved #3229


maxkadel (changes requested):
Test please!

christinach:
Looks ok on catalog-qa.

PR #3286, opened 2022-12-06
Don't filter out viewers if it's a MVW

### LINKED ISSUE #3285: Only volumes have a Figgy viewer when a resource has a MVW and a Volume with the same Metadata ID
### Expected behavior
Two viewers should show for examples like https://catalog.princeton.edu/catalog/9950403683506421 

### Actual behavior
One viewer shows.

Note: We might revert the MVW to a single volume to get this to display for a class. Check the status of https://figgy.princeton.edu/catalog/54b399c7-f28c-46f6-a7b4-c835a60516c4 - if it has two volumes and is displaying, then this issue is fixed.

### Impact of this bug

We can't use MSI & Source Record in the catalog.

The code causing this seems to be https://github.com/pulibrary/orangelight/blob/f197ef4dbb9755136d6f4522ad14b7705856c92f/app/javascript/orangelight/figgy_manifest_manager.js#L113-L133

I think the code exists so that if there's a MVW and a set of volumes all hooked to the same bib-id it doesn't show both? I don't know of a use case like that, though.

tpendragon:
Closes #3285

This should also SIGNIFICANTLY improve the render time for large MVWs.

The priority for this is pretty high - it's breaking https://catalog.princeton.edu/catalog/9950403683506421 which is going to be used for a class on Thursday.

### DIFF
--- a/app/javascript/orangelight/figgy_manifest_manager.js
+++ b/app/javascript/orangelight/figgy_manifest_manager.js
@@ -114,21 +114,7 @@ class FiggyViewerSet {
       return []
     }
 
-    // Filter only for resources with child resources (i. e. non-FileSets) as members
-    // This is not performant, and should require a separate GraphQL query
-    const filteredResources = resources.filter((resource) => {
-      if (!resource['members'])
-        return true
-      return resource.members.filter((member) => {
-        return member.__typename != "FileSet"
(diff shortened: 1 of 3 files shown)


christinach (approved):
Thanks @tpendragon

PR #3298, opened 2022-12-16
Update README.md

christinach:
Update ruby, rails and blacklight versions. Remove blacklight_advanced_search from readme.


maxkadel (approved):
Thanks!

Issue #3302, opened 2022-12-19
Main branch is failing circleci


christinach:
see https://app.circleci.com/pipelines/github/pulibrary/orangelight/3081/workflows/5a8ad2d6-171b-4aed-a382-6f92117a1645/jobs/7772

christinach:
Reopening as circleci is still red.

christinach:
closed by https://github.com/pulibrary/orangelight/pull/3343

PR #3324, opened 2022-12-28
Increase cpu and ram in docker container - update capybara selenium configuration

christinach:
part of #3302 

Use resource_class to increase cpu and ram in docker container.
Update capybara/selenium configuration. 
Use default capybara driver.
Include puma in Gemfile


sandbergja (approved):
Thanks, @christinach ! 🐙

PR #3348, opened 2023-01-05
Add beginning harmful content feedback bar

### LINKED ISSUE #3328: Add the new feedback bar to the "record page"
### User story
Users should be able to see at the bottom of the record page the three different form options to ask a question, suggest a correction and report harmful language content. Align with the feedback/reporting mechanisms in Pulfalight to make a more consistent experience for users across our discovery systems.

### Concrete example
- https://findingaids.princeton.edu/catalog/C0101
- wireframe : https://github.com/pulibrary/orangelight/issues/3274#issuecomment-1354801937

At any time when the user scrolls up or down keep the feedback form bar at the bottom of the browser window so that it is always visible

maxkadel:
- Put behind feature flipper so we can have control of rollout and incrementally introduce feature - go to /features to turn on and test
- Right now links just go to top and are placeholders for future tickets

Closes #3328

At bottom of page:
image

Sticky!:
image

Mobile:
image

### DIFF
--- a/app/views/catalog/_show_harmful_content_feedback.html.erb
+++ b/app/views/catalog/_show_harmful_content_feedback.html.erb
@@ -1,0 +1,5 @@
+<nav class="navbar navbar-expand-lg harmful-content-feedback">
+  <a class="navbar-brand" href="#">Ask A Question</a>
+  <a class="navbar-brand" href="#">Suggest A Correction</a>
+  <a class="navbar-brand" href="#">Report Harmful Language</a>
+</nav>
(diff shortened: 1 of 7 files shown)


maxkadel:
Deployed to staging & feature turned on in that environment for QA.

app/views/catalog/_show_harmful_content_feedback.html.erb
```diff
@@ -0,0 +1,5 @@
+<nav class="navbar navbar-expand-lg harmful-content-feedback">
```
sandbergja, line 1:
```suggestion
<nav class="navbar navbar-expand-lg harmful-content-feedback" aria-label="feedback">
```

Since the `<nav>` element is used multiple times in the same page, it's best to include an aria-label or aria-labelledby to distinguish between them.

sandbergja (changes requested):
Thanks!  Just one small accessibility improvement suggestion.

app/views/catalog/_show_harmful_content_feedback.html.erb
maxkadel: Thanks so much, @sandbergja - I've committed your suggestion. I always appreciate your feedback, especially around accessibility!

sandbergja (approved):
Thanks, @maxkadel ! 🐡

PR #3350, opened 2023-01-05
Add the solr artifacts to the .gitignore

maxkadel:
These files are re-added each time I re-create my orangelight application locally, and it just adds mental overhead to remember not to commit them.


christinach (approved):
😄

sandbergja (approved):
Yay!

PR #3355, opened 2023-01-06
Even more deprecations

### LINKED ISSUE #2865: deprecation warning for should_show_spellcheck_suggestions?
Output from our tests:

DEPRECATION WARNING: should_show_spellcheck_suggestions? is deprecated and will be removed from a future release (moving into a private method of Blacklight::Response::SpellcheckComponent). (called from _app_views_catalog__did_you_mean_html_erb__260515581459729074_136740 at /home/circleci/orangelight/app/views/catalog/_did_you_mean.html.erb:1)

sandbergja:
Closes #2865

### DIFF
--- a/app/views/catalog/_did_you_mean.html.erb
+++ b/app/views/catalog/_did_you_mean.html.erb
@@ -1,5 +1,0 @@
-<% if should_show_spellcheck_suggestions? @response and  !@response.spelling.collation.nil? %>
-  <div id="spell">
-    <h4 class="suggest"><em><%= t('blacklight.did_you_mean', :options => link_to_query(@response.spelling.collation)).html_safe %></em></h4>
-  </div>
-<% end %>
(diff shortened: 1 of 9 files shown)


spec/features/advanced_searching_spec.rb
```diff
@@ -42,6 +42,14 @@
     expect(page).not_to have_content("Огонек : роман")
   end
 
+  it 'allows searching by publication date', js: true do
```
sandbergja, line 45: This test was to catch a regression I made in the advanced search form while working on the `render_range_input` deprecation.

app/views/catalog/_did_you_mean.html.erb
```diff
@@ -1,5 +0,0 @@
-<% if should_show_spellcheck_suggestions? @response and  !@response.spelling.collation.nil? %>
```
sandbergja, line 1: This partial is no longer needed, since we don't use the "Did you Mean?" feature

christinach (approved):
🍄

PR #3363, opened 2023-01-18
I2889 bookmark limit v2

sandbergja:


app/assets/javascripts/blacklight/blacklight.js
```diff
@@ -0,0 +1,12 @@
+// Customize which blacklight javascripts are imported
+// See: https://github.com/projectblacklight/blacklight/wiki/Using-Sprockets-to-compile-javascript-assets#customization-options
```
sandbergja, line 2: Using sprockets rather than vite here, since the blacklight_range_limit js assumes that the blacklight js is in sprockets

kevinreiss:
PR is compatible with BL7 Javascript. Going to eventually deploy this with BL8 compatibility after #3400 is completed.

Issue #3380, opened 2023-01-24
Run manually browse lists one at a time 


christinach:
If they are successful resume the cron job

christinach:
I generated the name_titles csv file and uploaded it in catalog-qa. Assuming that call numbers uploads successfully as well, since it was the same record that created the error in name-titles and call-numbers I will resume the cron job.

PR #3385, opened 2023-01-26
resume cron job to generate-load call-numbers and name-titles

christinach:
close https://github.com/pulibrary/orangelight/issues/3380


sandbergja (approved):
🐬

PR #3389, opened 2023-01-30
Remove facets helper

### LINKED ISSUE #3388: Remove facets helper
Only one of the methods in this helper file is still used.  It's used in a partial override that doesn't seem to be needed any more.  Let's remove both files!

sandbergja:
Closes #3388


app/components/blacklight/facet_item_component.rb
```diff
@@ -0,0 +1,124 @@
+# frozen_string_literal: true
```
sandbergja, line 1: Moving this component into the same path as the upstream one, so that we can override its behavior globally.

spec/requests/request_spec.rb
```diff
@@ -373,9 +373,10 @@
     it 'are configured with a tooltip for removing the book format facet parameter' do
       stub_holding_locations
       get '/?f%5Bformat%5D%5B%5D=Book&q=&search_field=all_fields'
-      expect(response.body.include?('<i class="fa fa-times" aria-hidden="true" data-toggle="tooltip" data-original-title="Remove"></i>')).to eq true
```
sandbergja, line 376: Not changing the expectations here, just trying to make this test a little clearer.

spec/features/facets_spec.rb
```diff
@@ -24,24 +24,24 @@
     before do
       visit '/catalog'
     end
-    it 'card-header in facet-panel-collapse has link with aria-expanded' do
-      a_tag_first = find(:xpath, "//*[@id='facet-panel-collapse']/div[1]/div[1]/a")
+    it 'card-header in facet-panel-collapse has button with aria-expanded' do
```
sandbergja, line 27: The selector has changed, but the behavior has not.  Updating the spec to use the new selector

spec/views/catalog/index_spec.rb
```diff
@@ -50,7 +50,7 @@
   describe 'advanced search' do
     it 'inclusive facets display when applied' do
       visit '/catalog?f_inclusive%5Bformat%5D%5B%5D=Audio&search_field=advanced'
-      expect(page).to have_selector('.blacklight-format.facet_limit-active')
+      expect(page).to have_selector('.blacklight-format.facet-limit-active')
```
sandbergja, line 53: One more selector change

app/components/blacklight/facet_item_component.rb
maxkadel: Ok, after much git traversing, I have found the original ticket! See https://github.com/pulibrary/orangelight/issues/210. If I'm reading it correctly (and it's 7:15 am my time, so I might not be), it sounds like it was originally added for accessibility, and I suspect we have better ways to do that now, that are hopefully in the Blacklight facet? And if not maybe we can add them there?

sandbergja: Thanks for tracking that down, @maxkadel !  I've removed our customization accordingly.

maxkadel (approved):
Looks great! 🦄 🍄

PR #3398, opened 2023-02-07
Added suggest a correction form

### LINKED ISSUE #3330: Create Suggest a Correction Form - emails to cataloging queue
### User story
part of https://github.com/pulibrary/orangelight/issues/3272

### Concrete example
example from findingaids: Screen Shot 2023-01-03 at 12 15 51 PM

Send email to cataloging queue
No need to add the field 'Box/Container Number' in the catalog form.

rladdusaw:
Resolves #3330


app/controllers/contact_controller.rb
```diff
@@ -11,9 +11,24 @@ def question
     end
   end
 
+  def suggestion
+    @form = SuggestCorrectionForm.new(suggestion_params)
+    if @form.valid? && @form.submit
+      flash[:success] = 'Your suggestion has been submitted'
```
maxkadel, line 17: I think we may want to change this to flash.now for all of these, but we can wait and do that in a separate commit on all 3

app/views/contact/suggestion_success.html.erb
```diff
@@ -0,0 +1,6 @@
+<%= render Blacklight::System::ModalComponent.new do |component| %>
+  <% component.title { t('blacklight.suggestion.form.title') } %>
```
maxkadel, line 2: Need a corresponding addition to `config/locales/blacklight.en.yml` - see e.g. https://github.com/pulibrary/orangelight/blob/0528c1d2ac991fc7ed87bf82e72e7af802d7ccc6/config/locales/blacklight.en.yml#L82-L84

spec/forms/suggest_correction_form_spec.rb
```diff
@@ -0,0 +1,79 @@
+# frozen_string_literal: true
```
maxkadel, line 1: Maybe also add a system test - e.g. https://github.com/pulibrary/orangelight/blob/0528c1d2ac991fc7ed87bf82e72e7af802d7ccc6/spec/system/catalog_show_spec.rb#L98-L111

maxkadel (changes requested):
Looking good! 🎉 

The changes I'd like to see are just adding the newlines and a system spec for the modal.

maxkadel (approved):
🎉

PR #3404, opened 2023-02-08
Include eslint in circleci

christinach:
closes https://github.com/pulibrary/orangelight/issues/3403


sandbergja (changes requested):
Thanks @christinach !  Does `eslint` also need to be included in the `build_accept` workflow?  Looks like eslint didn't run.

sandbergja (approved):
Thanks, @christinach !

🐬🐬🐬🐬🐬🐬

PR #3407, opened 2023-02-09
Move jest tests in a separate job in circleci config.yml

### LINKED ISSUE #3405: Jest tests should run in their own circleci job
The jest tests currently run as part of the `test` job.  Since that job is run in parallel, the jest tests end up being run 3 times, once by each machine!

We should separate the jest tests into their own job.  We can use the `eslint` job as an example.

christinach:
closes #3405


sandbergja (approved):
Thanks, @christinach ! 🐬

PR #3424, opened 2023-02-16
Add pa11y package. Include it in circleci

### LINKED ISSUE #3409: Add pa11y to CI for orangelight
[pa11y](https://pa11y.org/) is a nice npm package to test for accessibility.  It can use the axe-core rules, but it can also use additional rules.  It provides flexibility!

christinach:
closes #3409


christinach:
@sandbergja and I think that even though pa11y is a very good tool, it's hard to integrate with Rails. We will continue using for automated accessibility testing the 'axe-core' and the erblint-github gems.

PR #3427, opened 2023-02-21
Add diagrams of the requests architecture

sandbergja:


kevinreiss (commented):
These look great. Some of the business logic related to the services requestable items are elibiglbe for is currently in the Requests::Router class that is class by the request object in this method: https://github.com/pulibrary/orangelight/blob/main/app/models/requests/request.rb#L88. This is the only place the router is called, probably with some refactoring we could get rid of this class. It's currently an awkward because this class basically exists to alter the state of the requestable objects themselves. It's probably something we could handle a different way with re-factoring.

kevinreiss (approved):

Issue #3428, opened 2023-02-21
Stackmap specs are failing


sandbergja:
See https://app.circleci.com/pipelines/github/pulibrary/orangelight/3408/workflows/ac3be612-3726-4777-af3b-c71ec95932ba/jobs/9039 for an example failure.

sandbergja:
I was able to reproduce this test failure locally, and can also recreate it in the UI on catalog-staging when the Firestone Locator FlipFlop is turned off.  Is this an issue with the vendor's JS?  If so, should we mark the specs as pending?

maxkadel:
@sandbergja I think marking the specs as pending makes sense. Maybe we can also let them know their JS is broken?

sandbergja:
We are waiting on the vendor to let us know when these javascripts are ready.

Once their js works, our tests will inform us that we need to remove the pending line.

rladdusaw:
Take pending off of tests. Update language to match new button.

PR #3429, opened 2023-02-22
Mark stackmap vendor dependent tests pending

maxkadel:
Connected to #3428


kevinreiss (approved):

kevinreiss:
The vendor is supposed to be making adjustments on the scripts, I suspect they are changing the live scripts they sent to us.

PR #3434, opened 2023-02-27
One vendor-reliant test no longer pending

maxkadel:
- the fa-map-marker is now aria-hidden
- The two tests in spec/system/stackmap_spec.rb are still pending


sandbergja (approved):
🦄

PR #3448, opened 2023-03-08
Disable Metrics/AbcSize on individual methods

maxkadel:
- Scope the exclusion to a single method, rather than an entire file. This reduces the likelihood of accidentally introducing even more complexity.


maxkadel:
Hrm, actually, I think this is going to make it harder to find the worst offenders in the future, even though I think this is generally a good idea.

PR #3486, opened 2023-04-03
Render component/partial names as HTML comments in development environment

sandbergja:
closes https://github.com/pulibrary/orangelight/pull/3486

See https://blog.saeloun.com/2020/05/11/rails-support-annotates-html-output-with-template-file-names


christinach (approved):
Thanks @sandbergja 🍄

PR #3489, opened 2023-04-03
Allow components to be specified in index field config

### LINKED ISSUE #3488: Specifying a component for an index field in the catalog controller has no effect
### Expected behavior
As a developer, if I do the following in the Catalog Controller, the search results page should use the `ExampleComponent` to render the Published/Created field.

```
config.add_index_field 'pub_created_display', label: 'Published/Created', component: ExampleComponent
```

### Actual behavior

Adding the `component` to an index field's configuration has no effect.

### Steps to replicate

1. Create a file at app/components/example_component.rb with the following content:
    ```
    class ExampleComponent < Blacklight::MetadataFieldComponent
      def call
        'RENDERING A COMPONENT'
      end
    end
    ```
1. In the catalog controller, find the line that begins `config.add_index_field 'pub_created_display'`
1. Append `, component: ExampleComponent` to that line.
1. In your browser, do a search.
1. Note that the display has not changed at all.

### Impact of this bug

Blocking work on #3458

sandbergja:
closes #3488 

Thank you to @maxkadel for suggesting this approach!

### DIFF
--- a/app/components/index_metadata_component.html.erb
+++ b/app/components/index_metadata_component.html.erb
@@ -1,0 +1,9 @@
+<% @presenter.field_presenters.reject { |field| field.field_config.show == false }.each do |field_presenter| -%>
+  <li>
+    <% if field_presenter.component %>
+      <%= render field_presenter.component.new field: field_presenter, layout: IndexMetadataFieldLayoutComponent %>
+    <% else %>
+      <%= field_presenter.render %>
+    <% end %>
+  </li>
+<% end -%>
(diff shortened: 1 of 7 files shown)


christinach (commented):
@sandbergja thank you. I think you should update the title of the PR. I don't see in this pr the component to be specified in the catalog_controller.

christinach (approved):
ok, I see you why you named it this way. It's ok thank you!

PR #3497, opened 2023-04-06
Refactor render_top_field?

### LINKED ISSUE #3495: Standard catalog controller configuration for showing / hiding fields does not work
Per the [Blacklight documentation](https://github.com/projectblacklight/blacklight/wiki/Configuration---Results-View#conditional-rendering), you should be able to add a conditional to a configuration line (e.g. `if: true` or `if: display_my_show_field?(field_config)`); however, we have re-used some of this feature to show or hide fields in the prioritized / top section of the show page - see [render_top_field? method](https://github.com/pulibrary/orangelight/blob/0b488f547dc1b27738309bf47ce34dbeb310a034/app/helpers/catalog_helper.rb#L55-L58). If you add `if: false` to an `add_show_field` line, it will not display in the bottom section, but it will display in the top section.

We should re-instate the ability to show / hide a field on the show page, without making it automatically render in the top section.

maxkadel:
- Previous implementation meant that we could not easily use the "if:" configuration option without surprising behavior.
- Added class to `hash_show` partial to make testing easier

Closes #3495

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -251,16 +251,18 @@ class CatalogController < ApplicationController
     # config.add_show_field 'subtitle_vern_display', :label => 'Subtitle'
 
     # Top fields in show page / prioritized information
-    config.add_show_field 'author_display', label: "Author/\u200BArtist", browse_link: :name, if: false
+    # Top fields must include `if: false`, or they will be duplicated in the Details section
+    # They should be designated with either `default_top_field: true`, `coin_top_field: true` or both
+    config.add_show_field 'author_display', label: "Author/\u200BArtist", browse_link: :name, if: false, default_top_field: true
     # For uniform titles from 240 field - when there is both an author and uniform title
-    config.add_show_field 'name_uniform_title_1display', label: 'Uniform title', helper_method: :name_title_hierarchy, if: false
+    config.add_show_field 'name_uniform_title_1display', label: 'Uniform title', helper_method: :name_title_hierarchy, if: false, default_top_field: true
     # For uniform titles from 130 field - when there is only a uniform title, and no author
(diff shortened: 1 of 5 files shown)


maxkadel:
Deployed to staging 4/6/2023, 5:24 pm EST

app/controllers/catalog_controller.rb
```diff
@@ -251,16 +251,18 @@ class CatalogController < ApplicationController
     # config.add_show_field 'subtitle_vern_display', :label => 'Subtitle'
 
     # Top fields in show page / prioritized information
-    config.add_show_field 'author_display', label: "Author/\u200BArtist", browse_link: :name, if: false
+    # Top fields must include `if: false`, or they will be duplicated in the Details section
```
sandbergja, line 254: Not a blocker, but it does seem awkward to have to remember to use the `if: false` as well as the `*_top_field: true`.  It would be more ergonomic to just need one.  Maybe we could open another ticket for that as future work?

sandbergja (approved):
Thanks, @maxkadel  🦄

Just one comment that I think we should do eventually, but it doesn't have to block this getting merged.

app/controllers/catalog_controller.rb
sandbergja: Oooh, also noting that if we get rid of the `if: false` requirement in the CatalogController, we could scrap the `AlwaysShowMetadataFieldComponent` too.  Literally its only job is to make those top_fields show despite being set to `if: false`.

christinach (approved):
Thanks @maxkadel !

app/controllers/catalog_controller.rb
maxkadel: Yes, I was wanting to get rid of the `if: false`, but then was worrying I was getting into perfectionism, and thought for now I'd get it to where we can get in the 583 changes, then keep refactoring. I forgot about the AlwaysShowMetadataFieldComponent class, that would be nice to be able to remove.

PR #3518, opened 2023-04-17
Address show_fields_for deprecation

### LINKED ISSUE #3490: Deprecation warning: Calling show_fields_for with a scalar value is deprecated
Seen when running the test suite:

```
DEPRECATION WARNING: Calling show_fields_for with a scalar value
is deprecated. It must receive an Enumerable. (called from
_app_views_catalog__show_top_fields_html_erb__4288495172432519664_78100
at /home/circleci/orangelight/app/views/catalog/_show_top_fields.html.erb:5)
```

maxkadel:
- When a document does not have a format field, it will not raise a deprecation warning

Closes #3490


sandbergja (approved):
Thanks, @maxkadel ! 🦄

PR #3525, opened 2023-04-19
Use zeitwerk as autoloader

### LINKED ISSUE #3524: Migrate to zeitwerk
Before upgrading to Rails 7 (#3321), we will need to address a few errors that appear when we use the zeitwerk autoloader.  Zeitwerk is the only autoloader available in Rails 7. 

### Technical implementation

- [Configure zeitwerk in config/application.rb](https://guides.rubyonrails.org/classic_to_zeitwerk_howto.html#applications-running-rails-6-x)
- Run `bin/rails zeitwerk:check`
- Fix any errors
- Repeat steps 2&3 as needed

sandbergja:
closes #3524


christinach (approved):
Thanks @sandbergja 🍄

PR #3528, opened 2023-04-20
Refactor to reduce method complexity

### LINKED ISSUE #3446: Lower maximum ABC size (complexity) allowed by Rubocop
In the rubocop_todo.yml the allowed `Metrics/AbcSize` is [set to 52](https://github.com/pulibrary/orangelight/blob/main/.rubocop_todo.yml#L18); however, this is a "dangerous" number according to the [Rubocop documentation](https://docs.rubocop.org/rubocop/cops_metrics.html#metricsabcsize):

> Interpreting ABC size:
> - ⇐ 17 satisfactory
> - 18..30 unsatisfactory
> - `> 30` dangerous

There are also some files that are explicitly excluded from even this low bar - see https://github.com/pulibrary/orangelight/blob/main/.rubocop.yml#L23-L38.

We should:
- [ ] Find the method that has the highest AbcSize (commenting out the stanzas in the `.rubocop_todo.yml` and `.rubocop.yml` should allow you to find it)
- [ ] Ensure it has complete test coverage using data in Coveralls
- [ ] Refactor the method so that it has lower complexity, probably by breaking into smaller methods
- [ ] Update the `.rubocop_todo.yml` and `.rubocop.yml` to ensure there is not regression
- [ ] Repeat for the method with the next highest AbcSize

rladdusaw:
Resolves #3446

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -22,16 +22,6 @@ AllCops:
 
 Metrics/AbcSize:
   Exclude:
-    - 'app/models/concerns/blacklight/solr/document/marc.rb'
-    - 'app/controllers/account_controller.rb'
-    - 'app/controllers/orangelight/browsables_controller.rb'
-    - 'app/controllers/users/omniauth_callbacks_controller.rb'
-    - 'app/controllers/account_controller.rb:'
-    - 'app/controllers/concerns/orangelight/catalog.rb'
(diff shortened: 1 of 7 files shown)


sandbergja (approved):
Thanks, @rladdusaw !

PR #3533, opened 2023-04-21
i3090: Remove bookmarks not tied to a current user

### LINKED ISSUE #3090: Bookmarks table has bookmarks from user/s that do not exist anymore in the users table
**Observation**
Bookmarks table has bookmarks from user/s that do not exist anymore in the users table.

Search the bookmarks table for bookmarks that belong to users that no longer exist in the users table.

sandbergja:
* Remove bookmarks not tied to a current user.
* Add a database constraint that deletes any remaining bookmarks when the user is deleted.
* Adds a test to confirm that the guest user delete rake task also removes the guest users' bookmarks.

closes #3090


spec/db/migrate/20230419231330_add_user_fk_to_bookmarks_spec.rb
```diff
@@ -0,0 +1,62 @@
+# frozen_string_literal: true
```
sandbergja, line 1: Not sure if this test will be helpful long term, except maybe as an example?  Thoughts on if we should keep it in this PR?

christinach: @sandbergja we can remove this test. Thanks.

christinach:
@sandbergja I didn't merge so that you remove the test you pointed out.

PR #3537, opened 2023-04-27
3523 purge bookmarks solr

### LINKED ISSUE #3523: Purge Bookmarks Pointing at Solr Documents that No Longer exist.
### Implementation notes

- Before deleting any bookmarks, ensure that the solr id is not a voyager ID that could be mapped to a valid alma id.
- Be mindful of the number of solr calls you make.

maxkadel:
Closes #3523

### DIFF
--- a/app/models/bookmark.rb
+++ b/app/models/bookmark.rb
@@ -14,7 +14,8 @@ class Bookmark < ApplicationRecord
   end
 
   def self.destroy_without_solr_documents
-    Bookmark.find_in_batches do |bookmarks|
+    Bookmark.find_in_batches.with_index do |bookmarks, batch|
+      Rails.logger.info { "Processing destroy_without_solr_documents group ##{batch}" }
       bookmark_doc_ids = bookmark_doc_ids(bookmarks)
       doc_ids_without_solr_doc = bookmark_doc_ids - doc_ids_in_solr(bookmark_doc_ids)
 
@@ -26,9 +27,18 @@ class Bookmark < ApplicationRecord
     end
   end
 
+  def self.update_to_alma_ids
+    Bookmark.where("LENGTH(document_id) <= 7").find_in_batches.with_index do |bookmarks, batch|
+      Rails.logger.info { "Processing update_to_alma_ids group ##{batch}" }
+      bookmarks.each do |bookmark|
+        bookmark.document_id = bookmark.voyager_to_alma_id
+        bookmark.save if bookmark.changed?
+      end
+    end
+  end
+
   def self.bookmark_doc_ids(bookmarks)
-    ids = bookmarks.map(&:document_id).to_set
-    ids.map { |id| ensure_voyager_to_alma_id(id) }.to_set
+    bookmarks.map(&:document_id).to_set
   end
 
   def self.doc_ids_in_solr(bookmark_doc_ids)
@@ -39,8 +49,8 @@ class Bookmark < ApplicationRecord
 
   # This method duplicates logic in the SolrDocument, but it did not seem well-suited for
   # re-use directly here (needing to instantiate a SolrDocument for each ID before being able to call it)
-  def self.ensure_voyager_to_alma_id(id)
-    return id if id.length > 7 && id.start_with?("99")
-    "99#{id}3506421"
+  def voyager_to_alma_id
+    return document_id if document_id.length > 7 && document_id.start_with?("99")
+    "99#{document_id}3506421"
   end
 end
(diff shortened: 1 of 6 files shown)


spec/factories/bookmarks.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
```
sandbergja, line 3: Nice, thanks for adding this factory!

sandbergja (changes requested):
Thanks, @maxkadel !  It's looking good, but I got an error when I ran `bundle exec rake orangelight:clean:bookmarks:without_solr_documents` on staging:

```
RSolr::Error::Http: RSolr::Error::Http - 414 Request-URI Too Long
Error: <html>
<head><title>414 Request-URI Too Large</title></head>
<body>
<center><h1>414 Request-URI Too Large</h1></center>
<hr><center>nginx/1.23.2</center>
</body>
</html>
```

maxkadel:
> Thanks, @maxkadel ! It's looking good, but I got an error when I ran `bundle exec rake orangelight:clean:bookmarks:without_solr_documents` on staging:
> 
> ```
> RSolr::Error::Http: RSolr::Error::Http - 414 Request-URI Too Long
> Error: <html>
> <head><title>414 Request-URI Too Large</title></head>
> <body>
> <center><h1>414 Request-URI Too Large</h1></center>
> <hr><center>nginx/1.23.2</center>
> </body>
> </html>
> ```

Ok, thanks for testing it (I didn't want to without having had another set of eyes on it). 

It's probably that the default for finding in batches is 1,000, and that's probably too many for building the Solr url. I think we can lower the size of the batches - do you have a guess what size would be reasonable?

spec/models/bookmark_spec.rb
```diff
@@ -7,4 +7,65 @@
...
+    it 'does not change Bookmarks with an alma document_id' do
+      expect do
+        Bookmark.update_to_alma_ids
+      end.not_to change { bookmark_three.reload.document_id }
+    end
+
+    it 'updates bookmarks with a voyager document_id' do
+      expect do
+        Bookmark.update_to_alma_ids
+      end.to change { bookmark_one.reload.document_id }
```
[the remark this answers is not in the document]
maxkadel: Done!

maxkadel:
Hrm, when I ran it on staging, the number of bookmarks went from 2,193 to 110, which makes me a little nervous - that seems like an awfully big reduction. I wonder if it's not waiting long enough for the Solr response / requesting too often?

maxkadel:
> Hrm, when I ran it on staging, the number of bookmarks went from 2,193 to 110, which makes me a little nervous - that seems like an awfully big reduction. I wonder if it's not waiting long enough for the Solr response / requesting too often?

The Solr query was defaulting to only 10 responses, no matter how many ids we requested, so it was calculating way more missing document_ids than were correct. I added a configuration to match the number of bookmarks we look up to the number of rows in the Solr response.

spec/models/bookmark_spec.rb
```diff
@@ -7,4 +7,65 @@
...
+    it 'does not change Bookmarks with an alma document_id' do
+      expect do
+        Bookmark.update_to_alma_ids
+      end.not_to change { bookmark_three.reload.document_id }
+    end
+
+    it 'updates bookmarks with a voyager document_id' do
+      expect do
+        Bookmark.update_to_alma_ids
+      end.to change { bookmark_one.reload.document_id }.from(voyager_id_one).to(converted_voyager_id)
```
sandbergja, line 68: Cool!  I didn't know about this syntax.

app/models/bookmark.rb
```diff
@@ -12,4 +12,47 @@ class Bookmark < ApplicationRecord
...
+      if doc_ids_without_solr_doc.present?
+        doc_ids_without_solr_doc.each do |doc_id|
+          Bookmark.where(document_id: doc_id)&.destroy_all
+        end
+      end
+    end
+  end
+
+  def self.update_to_alma_ids
+    Bookmark.where("LENGTH(document_id) <= 7").find_in_batches.with_index do |bookmarks, batch|
```
sandbergja, line 32: For peace of mind, could we also confirm that this won't attempt to convert SCSB-*, coin-*, and dsp* ids to alma ids?  I don't think we have one, but theoretically, a document with id:coin-1 would also appear in these results.

sandbergja (changes requested):
Thanks, @maxkadel !  I really like the configurable batch approach.  Just one more thing to check. 🦄

app/models/bookmark.rb
maxkadel: Thanks, good catch, updated.

kevinreiss (approved):

PR #3538, opened 2023-04-27
Changed call number font to Roboto Mono

### LINKED ISSUE #3275: Use a serif font for call numbers
### Current display
When a call number has the characters L and I, it's hard to tell the difference because of sans-serif font. Example: https://catalog.princeton.edu/catalog/9947668823506421

### Desired display
Change to a serif font for call numbers to make it easier to tell the difference.

### Notes
Approved by DSSG in Feb. 2020

rladdusaw:
Resolves #3275


sandbergja (changes requested):
Thanks, @rladdusaw ! I noticed one issue: the "Browse related items" link underneath the call number is also appearing in the Roboto Mono font.

PR #3541, opened 2023-05-02
Show fill-in form when there are too many items to retrieve

### LINKED ISSUE #3536: Display Basic Request form for items that have more than 500? number of holdings
### User story
As a user I would like to have the opportunity be able to request items from holdings that have many item records attached to them. 

### Acceptance criteria

    * axe dev tools
    * Lighthouse
    * Wave

### Concrete example
Visit https://catalog.princeton.edu/catalog/995639693506421 and try to load the request form at https://catalog.princeton.edu/requests/995639693506421?aeon=false&mfhd=22515072090006421. It times out. 

### Implementation notes
Barring a larger re-factor an interim solution could be to look at the indexed item data in the solr_document for each record and count how many individual items are stored in the holdings_display1 key (note each holding in this key has it's own items array if items are present on that holding), if the count of items is above a threshold of 500 the request form could deliver the same fallback form we show for records that have no item data at all. This would at least allow the user to self-specify which year/volumen of an enumerated work they want.

maxkadel:
Closes #3536

See [example request page](https://catalog-staging.princeton.edu/requests/9933643713506421?aeon=false&mfhd=22727480400006421) 

Request page with fill-in enumeration option


kevinreiss (changes requested):
An option for dealing with with locations that are circulates: false. Could we add a case to this statement https://github.com/pulibrary/orangelight/blob/a60910573f2ce6cfa435d424aea481b2ae337d77/app/models/requests/request_decorator.rb#L72-L76 that calls the too_many_items? method and sets fill_in to true. I think that would allow the fill-in option to appear for the circulates: false holdings for this workaround.

kevinreiss (approved):

kevinreiss:
This is handling both the circulating and non-circulating locations now.

PR #3553, opened 2023-05-09
Use the same OpenURL ctx generation routine for all of orangelight

sandbergja:


spec/models/solr_document_spec.rb
```diff
@@ -185,10 +185,6 @@
       it 'returns a ctx with a format book' do
         expect(solr_document.to_ctx(solr_document['format']).to_hash['rft.genre']).to eq('book')
       end
-
```
sandbergja, line 188: We excluded titles because it caused trouble in umlaut.  We no longer use umlaut, so we can include titles in OpenURL again.

christinach (approved):
thanks @sandbergja 🍄

app/models/requests/solr_open_url_context.rb
```diff
@@ -1,85 +1,20 @@
...
-      end
-
-      def copy_metadata(format:, metadata:)
-        ctx = ContextObject.new
-        ctx.referent.set_format(format.to_s)
-        copy_generic_metadata(ctx:, metadata:, format:)
-        # canonical identifier for the citation?
-        ctx.referent.add_identifier("https://bibdata.princeton.edu/bibliographic/#{metadata[:id]}")
-        copy_referrer_info(ctx:, metadata:)
-        TYPE_MAPPING[format].each { |identifier, metadata_key| ctx.referent.set_metadata(identifier.to_s, metadata[metadata_key]) }
```
kevinreiss, line 65: How does the default routine handle the type mapping for formats?

Issue #3273, opened 2022-12-02
Submit Alma constituent record metadata with Aeon Requests


doncat:
TN 354603 provide an example.  A user requests a specific constituent title.  Only host data get passed to Aeon.  Staff need a way to identify the individual item within the host that is being requested.  Pulling the _Item..._ descriptive fields from the constituent, and putting its MMS ID in ReferenceNumber would solve the problem.  It would also parallel EAD component IDs in that field.  Location, CallNumber and ItemNumber (barcode) should continue to come from the host holdings and item, of course.
As a by-product, this change would help with use statistics.

christinach:
Example records that we could look at: 
Host record: https://catalog.princeton.edu/catalog/99125086293506421/staff_view

Constituent record: https://catalog.princeton.edu/catalog/9935069203506421/staff_view

christinach:
@doncat which are the descriptive fields `Pulling the Item... descriptive fields from the constituent,`?

doncat:
These are the Aeon fields:
ItemAuthor, ItemTitle, ItemSubTitle, ItemDate, ItemEdition, ItemVolume, ItemIssue, ItemPages, ItemISxN, ItemCitation, ItemNumber.
Some are kind of implausible!
ReferenceNumber (MMS ID) would be highly useful.

doncat:
Any news on this?  I've had renewed enquiries from the curatorial staff.  Apparently it's a sore point for them.

kevinreiss:
This is a fairly challenging update through I understand why it would help to have the constituent metadata passed over rather than the host record. We will still need to allow the request code to pass over any item level information that resides with the parent record. In our current backlog I do not think we will have time to take substantial look at this until March given other commitments for the upcoming month.

doncat:
April now.  Any prospect of looking more closely?

doncat:
Hello again.  Almost not April any more.  Is there any chance of progress on this matter?
Attaching note from a faculty member, a heavy user of the Jacques Derrida library as further illustration of the problem.
[Re_ Visitor_ the former Ambassador of France to the US.pdf](https://github.com/pulibrary/orangelight/files/11334019/Re_.Visitor_.the.former.Ambassador.of.France.to.the.US.pdf)

kevinreiss:
Hello we discussed this at our weekly check-in for OL issues today. There is a prospect to bring bibliographic level data for the constituent records into the Aeon request.  

[...]

PR #3559, opened 2023-05-12
Show page links directly to aeon, rather than a redirect

sandbergja:
This passes information from both the host and constituent record, closing #3273

Still to-do:

- [x] Manual qa
- [x] Add an integration test

### DIFF
--- a/spec/models/requests/aeon_url_spec.rb
+++ b/spec/models/requests/aeon_url_spec.rb
@@ -1,0 +1,69 @@
+# frozen_string_literal: true
+require 'rails_helper'
+
+RSpec.describe Requests::AeonUrl do
+  let(:holdings) do
+    { "12345" => {
+      "location" => "Special Collections - Numismatics Collection",
+      "library" => "Special Collections",
+      "location_code" => "rare$num",
+      "call_number" => "Coin 3750",
+      "items" => [{ "holding_id" => "22740186070006421", "id" => "23740186060006421", "barcode" => "24680" }]
+    } }
+  end
+  let(:document) do
+    SolrDocument.new({
+                       id: '9999999',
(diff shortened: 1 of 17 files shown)


spec/models/requests/aeon_url_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+  context 'when a specific item is passed in' do
+    subject do
+      item = Requests::Requestable::Item.new({ "barcode" => "32101071302192", "id" => "23667098960006421", "holding_id" => "22667098990006421", "copy_number" => "1", "status" => "Available", "status_label" => "Item in place", "status_source" => "base_status", "process_type" => nil, "on_reserve" => "N", "item_type" => "Closed", "pickup_location_id" => "rare", "pickup_location_code" => "rare", "location" => "rare$ctsn", "label" => "Special Collections - Cotsen Children's Library", "description" => "Vol 1: no. 1 - 4 Jan 2013 - Oct 2013", "enum_display" => "Vol 1: no. 1 - 4", "chron_display" => "Jan 2013 - Oct 2013", "in_temp_library" => false })
+      described_class.new(document:, item:).to_s
+    end
+    it('takes the barcode from the item') do
+      expect(subject).to include('ItemNumber=32101071302192')
+    end
+    it('takes enumeration from the item') do
+      expect(subject).to include('rft.volume=Vol+1%3A+no.+1+-+4')
```
kevinreiss, line 89: Per Aeon's OpenURL mapping table it should accept "rft.volume" as a valid param for volume info, however by manual test it does not. It did accept just plain "volume". The Aeon form param is literally "ItemVolume", maybe for now we should pass enumeration as that param.

kevinreiss (approved):
I will make a separate ticket to deal with my comment on the volume.

PR #3563, opened 2023-05-16
Remove unused requests tests

sandbergja:
Removes `pending` tests, `xit` tests, commented-out tests, and tests with no assertions.

### DIFF
--- a/spec/features/requests/request_spec.rb
+++ b/spec/features/requests/request_spec.rb
@@ -51,140 +51,6 @@ describe 'request', vcr: { cassette_name: 'request_features', record: :none }, t
       end
     end
 
-    context 'Temporary Shelf Locations' do
-      describe 'Holding headings', js: true do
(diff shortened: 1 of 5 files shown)


spec/features/requests/request_spec.rb
kevinreiss, line 1: This batch of tests in request_spec.rb  are likely the only ones in this clean-up PR that we might have to test in a future state. We had a workflow where anyone could supply an email and have an item from ReCAP paged to Firestone, hence all the "fill_in" lines in each test for email. These were called "Guests" at the time. They were walk-in patrons who had no long-term affliation with the Library. We now use "Guest" as a defined patron group in Alma so it's very different than before. I don't think any of the code here is particularly relevant though. If we did extend support again I think we'd write the tests differently now.

kevinreiss (approved):
Thanks for cleaning this up @sandbergja.

PR #3568, opened 2023-05-18
i3554 missing illiad elements

### LINKED ISSUE #3554: ILLiad request is not passing the right data
### Expected behavior
ILLiad request passes the information that is the most helpful for resource sharing staff.

### Actual behavior
There are some problems with the data passed to the form

### Steps to replicate

1. Go to https://catalog.princeton.edu/catalog/9953212593506421
2. Press the Request button
3. Go to your account > Digitization requests
4. Open your ILLiad account
5. Note that the date field is empty.  It should have the date from the solr document
6. Note that the ISBN is "9780521860765 978052".  It should either be just the first ISBN ("9780521860765") or all of the ISBNs (we're not sure -- we should ask Liza and Nick)
7. Note that the form includes "COVID-19 Campus Closure ".  It should have some other text that we can get from Liza and Nick.

### Impact of this bug

It makes it harder for staff to identify a lender for materials, especially for materials without an ISBN.

We probably are missing a date due to #3553

sandbergja:
closes #3554


maxkadel (approved):
Looks good! 🍄

Issue #3599, opened 2023-05-30
Mulit-Item Special Collections Aeon Request form Fails to load. 


kevinreiss:
### Expected behavior
When I visit an special collections holding with many items attached to it I expect to see a list of items I can request via the Reading Room Request Button (aka Aeon). 

### Actual behavior
For some items the form produces the generic error screen. 

### Steps to replicate
Try to visit https://catalog.princeton.edu/requests/9930682063506421?mfhd=22707844130006421

### Impact of this bug
Users are unable to request individual items on Aeon holdings. 

### Honeybadger link and code snippet, if applicable
https://app.honeybadger.io/projects/54399/faults/96388200

sandbergja:
@kevinreiss and I looked at this today.  We found:

- This example record has more than 500 items, so the `too_many_items?` method returns true
- The `too_many_items?` behavior is set up to show the fill_in form for non-SC items, but it throws an error for aeon items.
- We're not sure yet what the desired behavior is.  The ideas we discussed were:
    - Ignoring `too_many_items?` for special collection materials
    - Showing the fill_in form for SC materials.  This is not a good option, because the circulation department handles the fill_in form, not special collections.
    - Linking directly to aeon in this case.  We are currently pre-populating the form with the first item from the holdings_1display, which is not great, since it is probably not be the item that somebody wants.  We could leave it blank (but the show page doesn't show which items are available).

This issue affects 30 records in the catalog.

maxkadel:
Code merged and deployed to production.

PR #3613, opened 2023-06-20
Ensure NullItem inherits from Item and SimpleDelegator

maxkadel:
When NullItem does not have the correct inheritance, it does not behave like a hash, as Item does, raising errors. 

Connected to #3599


sandbergja (approved):
Thanks, @maxkadel ! 🦄

PR #3621, opened 2023-06-28
When there is a failure render correct form

maxkadel:
See https://app.honeybadger.io/projects/54399/faults/97177468#notice-trace

```
NoMethodError: undefined method `permit!' for nil:NilClass params[:ask_a_question_form].permit!
```


sandbergja (approved):
Thanks, @maxkadel! 🦄

Issue #3610, opened 2023-06-19
The `blacklight:delete_old_searches` rake task caused a deadlock in the database


sandbergja:
### Expected behavior
Our application doesn't create deadlocks.

### Actual behavior
It totally did!

### Steps to replicate

Not sure

### Impact of this bug

One database process wasn't able to complete its work, leading to a less predictable system.

### Honeybadger link and code snippet, if applicable

```
where(['created_at < ? AND user_id IS NULL', Time.zone.today - days_old]).delete_all
```

https://app.honeybadger.io/projects/54399/faults/96951231/

### Root cause

@carolyncole identified the issue: multiple servers are trying to delete the same rows at the same time!  Ansible sets up this cron job on all the orangelight servers:

https://github.com/pulibrary/princeton_ansible/blob/f0a58b969dd3f3a273adc8efc2ab67f95d42f11f/roles/blacklight_app/tasks/main.yml#L6

### Implementation notes

A possible approach would be:

- [x] remove the cron job from ansible
- [x] manually delete the cron job from the orangelight servers
- [x] set up a new cron job in whenever that only runs on a single orangelight server

sandbergja:
```
root@lib-postgres-prod3:/var/lib/postgresql/13/main/pg_log# grep -ni 337972254 postgresql-2023-06-19*
postgresql-2023-06-19_000000.csv:798216:2023-06-19 22:00:05.601 UTC,"orangelight","orangelight_production",6136,"128.112.201.229:50630",6490cfe4.17f8,3,"DELETE waiting",2023-06-19 22:00:04 UTC,97/3629833,337972254,LOG,00000,"process 6136 detected deadlock while waiting for ShareLock on transaction 337972255 after 1000.052 ms","Process holding the lock: 6137. Wait queue: .",,,,"while deleting tuple (1129,27) in relation ""searches""","DELETE FROM ""searches"" WHERE (created_at < '2023-06-12' AND user_id IS NULL)",,,"/opt/orangelight/shared/bundle/ruby/3.1.0/bin/rake","client backend"
postgresql-2023-06-19_000000.csv:798217:2023-06-19 22:00:05.601 UTC,"orangelight","orangelight_production",6136,"128.112.201.229:50630",6490cfe4.17f8,4,"DELETE",2023-06-19 22:00:04 UTC,97/3629833,337972254,ERROR,40P01,"deadlock detected","Process 6136 waits for ShareLock on transaction 337972255; blocked by process 6137.
[...]

PR #3627, opened 2023-06-30
Move task from cron to whenever

maxkadel:
- This allows us to run the task on only one server, avoiding database locks
- We will also need to find a way *not* to create the cron job via ansible
  - See https://github.com/pulibrary/princeton_ansible/pull/4040 which will remove the creation of the cron job - we will still need to manually delete them on other servers.

Connected to #3610


sandbergja (approved):
🦄

PR #3631, opened 2023-07-07
Removing feature flipper

### LINKED ISSUE #3618: Remove the harmful_content_feedback feature flipper
This feature has been enabled (i.e.  `Flipflop.harmful_content_feedback? == true` for some time.  We're keeping the feature, but it no longer needs a feature flipper.

### Success criteria

- [x] This flipflop feature is removed
- [x] Double check that there is no code that is called when `Flipflop.harmful_content_feedback?` is false

rladdusaw:
Resolves #3618

### DIFF
--- a/app/views/catalog/show.html.erb
+++ b/app/views/catalog/show.html.erb
@@ -21,9 +21,7 @@
         </dd>
       </dl>
     <% end %>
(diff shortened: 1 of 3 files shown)


lizgarciao (approved):
Thanks!

PR #3658, opened 2023-07-28
Upgrade selenium-webdrivers gem, remove webdrivers gem

sandbergja:
Based on this guidance: https://github.com/titusfortner/webdrivers/issues/247#issuecomment-1648154088

I also had to `xattr -d com.apple.quarantine /opt/homebrew/bin/chromedriver`, possibly because I had inadvertently upgraded chromedriver via brew recently?


maxkadel (approved):
Thank you! 🍄

Issue #3615, opened 2023-06-23
Audit the docs directory for accessibility


sandbergja:
Use the [accessibility audit procedure](https://github.com/pulibrary/dacs_handbook/blob/main/Accessibility/audit_procedure.md) to audit all pages in the /docs directory of this repository.  Address any accessibility issues.  Create github issues for any issues that would take longer than 1 day to address.

sandbergja:
This work is done!  Thanks, @lizgarciao , @ishasinha1 , and @christinach !

PR #3665, opened 2023-08-02
Fix header level order

lizgarciao:
connected #3615


sandbergja (approved):
Thanks, @lizgarciao ☺️

PR #3669, opened 2023-08-03
Make sure that boolean operators in json queries are actually intended

### LINKED ISSUE #3668: cleaned_query method should be used on JSON queries as well
This is a follow-up to #3666 , which only partially resolved the issue.

### Expected behavior

When I do a search for "DEMETER DOES NOT REMEMBER" in a field that has been configured to use the JSON Query DSL, it returns the book "Demeter does not remember".

### Actual behavior

It thinks that "NOT" is a boolean operator, and instead tries searching for titles that include "Demeter" and "Does", but not "remember".

1. Change the luceneMatchVersion to 7.1.0 in solr/conf/solrconfig.xml
1. Comment out the line `Rake::Task["pulsearch:solr:update"].invoke` in lib/tasks/server.rake
1. Change the `add_search_field` for 'all_fields' in app/controllers/catalog_controller.rb to:

    ```
    config.add_search_field 'all_fields', label: 'Keyword' do |field|
      field.clause_params = {edismax: {}}
    end
	```

1. `bundle exec rspec ./spec/requests/request_spec.rb:40`

### Impact of this bug
Blocks work on #3645

sandbergja:
Extends the fix from #817 to JSON DSL queries as well

closes #3668

### DIFF
--- a/app/models/search_builder.rb
+++ b/app/models/search_builder.rb
@@ -12,9 +12,17 @@ class SearchBuilder < Blacklight::SearchBuilder
                                      series_title_results pul_holdings html_facets
                                      numismatics_facets numismatics_advanced]
 
+  # mutate the solr_parameters to remove words that are
+  # boolean operators, but not intended as such.
   def cleanup_boolean_operators(solr_parameters)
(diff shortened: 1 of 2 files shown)


maxkadel (approved):
Thanks @sandbergja! 🍄

Issue #3645, opened 2023-07-25
Refactor numismatics advanced search using the built-in advanced search BL form


christinach:
The built-in advanced search form is included in BL 7 and BL 8.
Things to consider:
1. The queries will turn from q1, q2, q3 to [clause0], [clause1], [clause2] 
2. It uses view components rather the view partials. 
3. It will need a lot of work to make the form look as it does now.
4. It uses the json query DSL. This is the new way BL communicates with SOLR. This will require changes to the catalog controller.

### Acceptance criteria
- [ ] There is a new form
- [x] It is behind a feature flipper
- [x] It is ok if this form does not actually successfully perform a search - the search is blocked by changes in Solr / Blacklight

related to #3412

maxkadel:
See also [https://github.com/pulibrary/pul_solr/issues/370](pul_solr issue to add new handler)

PR #3670, opened 2023-08-03
Flipflop to opt in to the JSON query DSL

sandbergja:
To make it easier to test the JSON query DSL, let's make it easy to turn on and off in the flipflop UI.  The steps to do this in a local dev environment:

1. Change the luceneMatchVersion to 7.1.0 in solr/conf/solrconfig.xml
2. Comment out the line `Rake::Task["pulsearch:solr:update"].invoke`  in lib/tasks/server.rake
3. `ORANGELIGHT_ADMIN_NETIDS=my-netid bundle exec rails s`
4. Turn the feature on/off as desired at localhost:3000/features

part of #3645

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -478,6 +478,7 @@ class CatalogController < ApplicationController
     #   field.include_in_simple_select = false
 
     config.add_search_field 'all_fields', label: 'Keyword' do |field|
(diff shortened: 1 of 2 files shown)


maxkadel (approved):
Awesome! 🍄 

It'd be cool if we could figure out a way we could switch back and forth on staging / QA without extra steps, but I'm not sure there's a good way to change the lucene match version other than the one you propose.

PR #3694, opened 2023-08-21
Added flipper feature for new aeon url

### LINKED ISSUE #3676: Toggle Aeon Base URL via Feature Flipper
The current aeon_base value should be set to this temporary test URL (for use only in staging and QA). 

```https://princeton.aeon.atlas-sys.com/logon```

When the new system goes live it will be 

```https://lib-aeon.princeton.edu/logon```

Th existing production base URL (this needs to remain in place until the morning of the migration). 

```https://lib-aeon.princeton.edu/aeon/aeon.dll```

The querystring also needs to be updated in the fashion you see in this commit: https://github.com/pulibrary/orangelight/pull/3675/commits/2aa7ab8ad5619fc6e07a4732167f45d1962cf6dc. 

The following two params need to be added to be the test and the current "OpenURL?" needs to be removed from the querystring. 
```
Action=10&Form=30&
```
The full base with the new required query string params looks like: https://lib-aeon.princeton.edu/logon?Action=10&Form30.
```

This method likely needs to be re-factored: https://github.com/pulibrary/orangelight/blob/ffcff4bbad555d8a7a2d862bc43f947071cb11d4/app/models/requests/aeon_url.rb#L14

rladdusaw:
Resolves #3676

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -505,7 +505,7 @@ GEM
       sprockets-rails
       tilt
     scrub_rb (1.0.1)
-    selenium-webdriver (4.10.0)
+    selenium-webdriver (4.11.0)
       rexml (~> 3.2, >= 3.2.5)
       rubyzip (>= 1.2.2, < 3.0)
       websocket (~> 1.0)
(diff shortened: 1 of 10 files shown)


maxkadel (changes requested):
Looks good, but I'd like to avoid Flipflop in controllers if at all possible.

maxkadel (approved):
Thanks @rladdusaw ! 🌮 🎉

Issue #3651, opened 2023-07-27
Remove unused geohash field from solr schema.xml


sandbergja:
This field is not used in bibdata or orangelight.  Furthermore, it uses a [fieldtype called GeoHash that is deprecated](https://solr.apache.org/docs/8_6_0/solr-core/org/apache/solr/schema/GeoHashField.html).  We should remove it from pul_solr.

sandbergja:
Closed by https://github.com/pulibrary/orangelight/pull/3697

PR #3697, opened 2023-08-22
Updated with latest solrconfig and schema

### LINKED ISSUE #3648: Remove unused title_addl_lr copyField from solr configuration
Our solr configuration copies the contents of the `title_addl_la` field to a field called `title_addl_lr`:

https://github.com/pulibrary/pul_solr/blob/9945ef4167f44acb277f02b0059fa9df992aee4d/solr_configs/catalog-production-v2/conf/schema.xml#L763

As far as I can tell, this field is not used anywhere.  Nor is it copied to any other field.  We should confirm this, and if it is truly not being used, remove it from the config in the pul_solr repository.

rladdusaw:
Resolves #3648 #3651


sandbergja (approved):
🐈‍⬛ Thanks, @rladdusaw !

Issue #3695, opened 2023-08-22
Move Flipflop out of controller


maxkadel:
Right now the JSON query DSL flipflop is within the CatalogController, which means that it won't actually switch the feature unless the application server is re-started. 

We should move this into a helper or service, so that Flipflop works as expected, to ease testing and have more consistent behavior.

rladdusaw:
Example code (Just don't start the method name with "get" to make Rubocop happy) https://github.com/pulibrary/orangelight/pull/3694/commits/13b1d340e6649f7c56c826d7890a66d70044a497

Issue #3412, opened 2023-02-10
Refactor advanced search using the built-in advanced search BL form


christinach:
The built-in advanced search form is included in BL 7 and BL 8.
Things to consider:
1. The queries will turn from q1, q2, q3 to [clause0], [clause1], [clause2] 
2. It uses view components rather the view partials. 
3. It will need a lot of work to make the form look as it does now.
4. It uses the json query DSL. This is the new way BL communicates with SOLR. This will require changes to the catalog controller.

### Acceptance criteria
- [x] There is a new form
- [x] It is behind a feature flipper
- [X] It is ok if this form does not actually successfully perform a search - the search is blocked by changes in Solr / Blacklight

maxkadel:
Still to do:
- Advanced search facets
- Actually searching when you press search
- Numismatics form still uses gem partials rather than Blacklight components

maxkadel:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/DACS-Maintenance-5da0db919c4ddf0001b2eb92/issues/pulibrary/orangelight/3412?planning-poker) with Zenhub @regineheberlein @kevinreiss @christinach @rladdusaw @sandbergja

sandbergja:
I was able to follow these procedures to get advanced search working with the blacklight .internal_test_app: https://code4lib.slack.com/archives/C54TB5WDQ/p1687452502967029

These don't work for orangelight -- I wonder if it's because we also have a complex default query that we also send.

kevinreiss:
Form is partially working, but need to work through these scenarios. 

1. Add support for Boolean operators (they are there now but don't work)
2. Pivot facets are failing when you edit an active search.

PR #3700, opened 2023-08-24
I3412 built in advanced search form

maxkadel:
Connected to #3412 

Current status:
- Styling parallels current advanced search
- Builds initial advanced searches correctly, including selecting field for search and passing along booleans
- Tests in `spec/features/advanced_searching_spec.rb` pass when run individually; however, not in the context of the whole suite, because the controller is not re-loaded, so the json dsl is not used when it should be. This should resolve if #3695 is resolved.
- From a search, clicking "Edit search" raises `Error: Pivot Facet needs at least one field name:`
- From advanced search, does not create a "constraint-value" on the search results page (also true for Numismatics search)
- When trying to remove a search constraint generated by regular search, raises `RSolr::Error::Http - 400 Bad Request
Error: Error when parsing json query, expect a json object here, but found : null`

### DIFF
--- a/app/components/orangelight/advanced_search_form_component.rb
+++ b/app/components/orangelight/advanced_search_form_component.rb
@@ -1,0 +1,43 @@
+# frozen_string_literal: true
+
+class Orangelight::AdvancedSearchFormComponent < Blacklight::AdvancedSearchFormComponent
+  def sort_fields_select
+    options = sort_fields.values.map { |field_config| [helpers.sort_field_label(field_config.key), field_config.key] }
+    return unless options.any?
+
+    select_tag(:sort, options_for_select(options, params[:sort]), class: "sort-select", aria: { labelledby: 'advanced-search-sort-label' })
+  end
+
+  def initialize_search_filter_controls
+    fields = blacklight_config.facet_fields.select { |_k, v| v.include_in_advanced_search || v.include_in_advanced_search.nil? }
+
+    fields.each do |_k, config|
+      config.advanced_search_component = Orangelight::FacetFieldCheckboxesComponent
+      display_facet = @response.aggregations[config.field]
+      with_search_filter_control(config:, display_facet:)
+    end
+  end
+
+  def pub_date_field
+    blacklight_config.facet_fields['pub_date_start_sort']
+  end
+
+  def pub_date_presenter
+    view_context.facet_field_presenter(pub_date_field, {})
+  end
+  def initialize_search_field_controls
+    search_fields.values.each.with_index do |field, i|
+      with_search_field_control do
+        fields_for('clause[]', i, include_id: false) do |f|
+          content_tag(:div, class: 'form-group advanced-search-field row mb-3') do
+            f.label(:query, field.display_label('search'), class: "col-sm-3 col-form-label text-md-right") +
+              content_tag(:div, class: 'col-sm-9') do
+                f.hidden_field(:field, value: field.key) +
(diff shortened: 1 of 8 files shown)


app/components/orangelight/advanced_search_form_component.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  def pub_date_presenter
+    view_context.facet_field_presenter(pub_date_field, {})
+  end
+
+  def initialize_search_field_controls
+    search_fields.values.each.with_index do |field, index|
+      with_search_field_control do
+        fields_for_etc(index:, field:)
```
sandbergja, line 32: Could the method get a more descriptive name, e.g.

```suggestion
        generate_fields(index:, field:)
```

spec/components/orangelight/advanced_search_form_component_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+
+RSpec.describe Orangelight::AdvancedSearchFormComponent, type: :component, advanced_search: true do
+  subject(:render) do
+    component.render_in(view_context)
+  end
+  let(:component) { described_class.new(url: '/whatever', response:, params:) }
+
+  let(:response) do
+    Blacklight::Solr::Response.new({}.with_indifferent_access, {})
+    # Blacklight::Solr::Response.new({ facet_counts: { facet_fields: {
```
sandbergja, line 13: Could you please remove the commented-out response?

spec/features/advanced_searching_spec.rb
```diff
@@ -2,7 +2,7 @@
 
 require 'rails_helper'
 
-describe 'advanced searching' do
+describe 'advanced searching', advanced_search: true do
```
sandbergja, line 5: What does `advanced_search: true` do?

spec/features/advanced_searching_spec.rb
```diff
@@ -50,6 +50,66 @@
...
+      # defaults to keyword
+      fill_in(id: 'clause_0_query', with: 'gay')
+      choose(id: 'clause_2_op_must_not')
+      # defaults to title
+      fill_in(id: 'clause_2_query', with: 'RenoOut')
+      click_button('advanced-search-submit')
+      expect(page.find(".page_entries").text).to eq('1 entry found')
+      expect(page).to have_content('Seeking sanctuary')
+      expect(page).not_to have_content('Reno Gay Press and Promotions')
+      # expect(page).to have_content('Title NOT RenoOut')
```
sandbergja, line 109: Could we remove this commented-out expectation?

sandbergja (changes requested):
Thanks, @maxkadel !  Just a few small cleanup suggestions.

spec/features/advanced_searching_spec.rb
maxkadel: Moved to separate failing-for-a-different-reason test.

app/components/orangelight/advanced_search_form_component.rb
maxkadel: Committed

spec/components/orangelight/advanced_search_form_component_spec.rb
maxkadel: Removed

spec/features/advanced_searching_spec.rb
sandbergja: So cool!  Thanks!

PR #3701, opened 2023-08-25
New approach for JSON Query DSL flipflop

### LINKED ISSUE #3695: Move Flipflop out of controller

sandbergja:
Rather than consulting Flipflop during the initial field configuration (in CatalogController), let's consult Flipflop where those configuration values are actually used (in the SearchBuilder)

closes #3695 

Partially reverts commit 656190e84785f4f29890eff24411f3e131e83488.

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -477,9 +477,7 @@ class CatalogController < ApplicationController
     #   :include_in_advanced_search => true
     #   field.include_in_simple_select = false
 
-    config.add_search_field 'all_fields', label: 'Keyword' do |field|
-      field.clause_params = { edismax: {} } if Flipflop.json_query_dsl?
-    end
+    config.add_search_field 'all_fields', label: 'Keyword'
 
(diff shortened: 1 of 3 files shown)


app/controllers/catalog_controller.rb
```diff
@@ -478,7 +478,6 @@ class CatalogController < ApplicationController
     #   field.include_in_simple_select = false
 
     config.add_search_field 'all_fields', label: 'Keyword' do |field|
```
maxkadel, line 480: Does this still need to take a block, without the conditional? Or could it just be:
```
    config.add_search_field 'all_fields', label: 'Keyword'
```

(can't seem to do this as a "suggestion" since it has deleted lines or something)

maxkadel (approved):
I really like this approach!

Made one small comment about the catalog_controller but it doesn't need to be a blocker.

PR #3709, opened 2023-09-05
Remove advanced search controller

maxkadel:
- Per @kevinreiss, this behavior is no longer needed, either with or without the BlacklightAdvancedSearch gem


rladdusaw (approved):

sandbergja (approved):
Thanks, @maxkadel !  🦄

PR #3754, opened 2023-09-20
i3476 prod update

kayiwa:
- comment out bionic vms
- upgrade all vms to jammy

related to https://github.com/pulibrary/princeton_ansible/issues/3476


christinach:
Thanks @kayiwa

PR #3764, opened 2023-09-26
Add aria-hidden true in <span> tag so that the screenreader will not announce encoded string

### LINKED ISSUE #3042: Developer code comments are printing to the live source.
Developer code comments are printing to the live source.

Example record: https://catalog.princeton.edu/catalog/99123192613506421

christinach:
closes #3042


sandbergja (approved):
🐬

PR #3766, opened 2023-09-27
Dont announce unapi-id when using screenreader

### LINKED ISSUE #3765: [Accessibility] unapi-id should not be announced by the screenreader
Announcing the unapi-id brings confusion and no value to a screen reader user.

christinach:
closes #3765


sandbergja (approved):
🐬

PR #3819, opened 2023-11-07
Add a search builder that allows use of a new request handler

### LINKED ISSUE #3808: Implement algorithm switching backend
We'll have chosen a method in pulibrary/figgy#3745. Implement that method here in such a way that it is implemented via our choice in pulibrary/figgy#3749 (feature flipper? something else?).

Ensure that the code is isolated enough that it can be taken and implemented in other projects.

## Acceptance Criteria

- [ ] There are tests that show that we can switch between different request handlers based on the query parameters
- [ ] If we adjusted our approach in deviation from https://github.com/pulibrary/orangelight/pull/3812, we update that documentation as part of closing this issue.

hackartisan:
Based on a querystring parameter, and also behind a feature flipper.

closes #3808 

requires pulibrary/pul_solr#375

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -787,4 +787,27 @@ class CatalogController < ApplicationController
       render plain: 'OK'
       nil
     end
+
+    def search_service_context
+      return {} unless Flipflop.multi_algorithm?
+      return {} unless alternate_search_builder_class # use default if none specified
+      { search_builder_class: alternate_search_builder_class }
+    end
+
+    def alternate_search_builder_class
+      return unless search_algorithm_param && allowed_search_algorithms.include?(search_algorithm_param)
+
+      "#{search_algorithm_param}_search_builder".camelize.constantize
+    end
+
+    # When adding a new Ranking Algorithm the name will need to be added to this list before it can be utilized
+    # For example if you added a cats ranking algorithm, with a CatsSearchBuilder you would add "cats" to this list ["engineering","cats"]
+    # This is to make sure the user can not just execute any SearchBuilder in the system
+    def allowed_search_algorithms
+      ["engineering"]
+    end
+
+    def search_algorithm_param
+      params[:search_algorithm]
+    end
 end
--- a/app/models/engineering_search_builder.rb
+++ b/app/models/engineering_search_builder.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+# This is a demonstration search builder, not intended for prodcution use.
+class EngineeringSearchBuilder < SearchBuilder
+  self.default_processor_chain += [:switch_request_handler]
+
+  def switch_request_handler(solr_parameters)
+    solr_parameters[:qt] = "engineering_search"
+  end
+end
--- a/spec/controllers/catalog_controller_spec.rb
+++ b/spec/controllers/catalog_controller_spec.rb
@@ -117,4 +117,75 @@ RSpec.describe CatalogController do
       end
     end
   end
+
+  describe "switching search algorithms" do
+    let(:repository) { CatalogController.blacklight_config.repository }
+
+    before do
+      allow(repository).to receive(:search).and_call_original
+      allow(controller.blacklight_config).to receive(:repository).and_return(repository)
+    end
+
+    context "when the multi-algorithm feature is on" do
+      before do
+        allow(Flipflop).to receive(:multi_algorithm?).and_return(true)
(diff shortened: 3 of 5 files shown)


app/controllers/catalog_controller.rb
```diff
@@ -743,48 +743,68 @@ def citation
...
+    { search_builder_class: alternate_search_builder_class }
+  end
+
+  def alternate_search_builder_class
+    return unless search_algorithm_param && allowed_search_algorithms.include?(search_algorithm_param)
+
+    "#{search_algorithm_param}_search_builder".camelize.constantize
+  end
+
+  def allowed_search_algorithms
```
carolyncole, line 803: I think we should add some comment here about needing to add any new searchbuilder to here for algorithm switching
```suggestion
  # When adding a new Ranking Algorithm the name will need to be added to this list before it can be utilized
  # For example if you added a cats ranking algorithm, with a CatsSearchBuilder you would add "cats" to this list ["engineering","cats"]
  # This is to make sure the user can not just execute any SearchBuilder in the system
  def allowed_search_algorithms
```

app/models/engineering_search_builder.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+# This is a demonstration search builder, not intended for prodcution use.
```
carolyncole, line 3: ❤️

lib/tasks/pulsearch.rake
```diff
@@ -40,6 +40,6 @@ namespace :pulsearch do
   private
 
     def url_for_file(file)
-      "https://raw.githubusercontent.com/pulibrary/pul_solr/main/solr_configs/catalog-production-v2/#{file}"
+      "https://raw.githubusercontent.com/pulibrary/pul_solr/main/solr_configs/catalog-production-v3/#{file}"
```
carolyncole, line 43: I had to change the core name to get the tests passing in pul_solr.  This may cause issues for the tests

spec/controllers/catalog_controller_spec.rb
```diff
@@ -117,4 +117,75 @@
       end
     end
   end
+
+  describe "switching search algorithms" do
+    let(:repository) { CatalogController.blacklight_config.repository }
+
+    before do
+      allow(repository).to receive(:search).and_call_original
+      allow_any_instance_of(Blacklight::Configuration).to receive(:repository).and_return(repository)
```
carolyncole, line 126: I wonder if we can utilize the `controller` variable here.  I did not think of it when I was playing with the tests
```suggestion
      allow(controller.blacklight_config).to receive(:repository).and_return(repository)
```

carolyncole (commented):
Just some thoughts...

app/models/engineering_search_builder.rb
```diff
@@ -0,0 +1,10 @@
...
+class EngineeringSearchBuilder < SearchBuilder
+  self.default_processor_chain += [:switch_request_handler]
+
+  def switch_request_handler(solr_parameters)
+    solr_parameters[:qt] = "engineering_search"
```
sandbergja, line 8: I like this `qt` approach!  I wonder if still works in the following specific case:

1. The JSON Query DSL feature is enabled
2. The View Components Advanced Search feature is enabled
3. The user does a search in the advanced search screen, and switches to the "engineering" requesthandler

If I remember correctly, those circumstances send queries to the `advanced` requesthandler, rather than the default one, where I don't know that `qt` would have an effect.

app/controllers/catalog_controller.rb
```diff
@@ -787,4 +787,27 @@ def basic_response
...
+    def alternate_search_builder_class
+      return unless search_algorithm_param && allowed_search_algorithms.include?(search_algorithm_param)
+
+      "#{search_algorithm_param}_search_builder".camelize.constantize
+    end
+
+    # When adding a new Ranking Algorithm the name will need to be added to this list before it can be utilized
+    # For example if you added a cats ranking algorithm, with a CatsSearchBuilder you would add "cats" to this list ["engineering","cats"]
+    # This is to make sure the user can not just execute any SearchBuilder in the system
+    def allowed_search_algorithms
```
sandbergja, line 806: Thanks for adding this check!

sandbergja (commented):
Thanks, team 1!  Just a question that came to mind while I was reading up on what the `qt` parameter means 🤣

sandbergja (approved):
Thanks, team 1!

Issue #3789, opened 2023-10-20
Build the form


kelynch:

kelynch:
Look to this for an example:
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBL2EyQVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--ed051b82d8688f9b99c4bd1666621394743a8441/Screenshot%202023-11-08%20at%2011.06.28%E2%80%AFAM.png)

sdellis:
[This is the most recent mockup.](https://github.com/pulibrary/orangelight/issues/3777#issuecomment-1802490424)

sdellis:
[This is the most recent mockup.](https://github.com/pulibrary/orangelight/issues/3777#issuecomment-1802490424)

kelynch:
Per the comment in #3777:

`
Based on a discussion the team had during today's check-in, we are going to simplify the initial deliverable by creating a link to the search results the user is reporting on instead of re-creating/i-framing the search results in the form modal. This link should open a new tab to prevent the user from losing any content they may have entered in the form.
`

Issue #3777, opened 2023-10-09
Design the biased search results reporting form


kelynch:
Part of #3761 

Utilize feedback page mockups from [this slide deck](https://docs.google.com/presentation/d/1ygrKEbRC3t5ppfWRdLpcOjv7LIrGwp8zaHeGIwukIEw/edit#slide=id.g20f5e26bc90_0_15).

* Review and decide on a wireframe to use (see https://github.com/pulibrary/orangelight/issues/3790)
* Define the preliminary fields of the form
  * Name (optional)
  * Email (optional)
  * Message (required)
  *  Search parameters (required, prepopulated with search parameters from the user)
* Define the text on the form
  * Use the text from the wireframe in the slide deck, minus the second sentence in the first paragraph (starts with "For example...") 
* Define the post-submission message of the form
  * Review the post-submission text for the harmful content form in Orangelight, and start with something like this. 
* Determine and design how the form looks and behave on the search results page
  * Starting with the orange banner approach

kelynch:
Consider the pattern created by the Report Harmful Language form and its call to action at the record level, and what we could do to be appropriately in sync with this.

The form should explicitly include the information they are submitting, including contextual search results, search terms, parameters, facets, etc in a meaningful way.

kelynch:
We decided not to provide preemptive examples of biased results to users, so as not to give them stereotype threat.  While this is in the mockups linked above, we do not intend to use this language.

sdellis:
One thing that's been expressed by the implementation team is the desire to be consistent and use UI patterns for reporting harmful language that we already have in place. The team has pointed to the lower tray of "buttons" on the Finding Aids and the Catalog record/show pages as something users may be familiar with. 

I wonder if this is the best approach though? I find these buttons to be a bit hard to see. I wonder if we would include the "Ask a Question" and "Suggest a Correction" links, or if we would simply have one gray button for "Report Biased Results"? We can contrast the original mockup that has a "banner/announcement" pattern with the mockup that uses the existing UI "button tray" pattern for interacting with PUL Staff. My question is "Would one large Report Biased Results button at the bottom essentially serve as a less visible and descriptive banner?"

Thoughts?

New Mockup:
<img width="1049" alt="Screenshot 2023-10-25 at 9 37 52 AM" src="https://github.com/pulibrary/orangelight/assets/609368/406a1ee9-0835-4fd7-8426-58f9d905b6d8">

Original Mockup:
[...]

PR #3822, opened 2023-11-08
Submit biased results feedback

hectorcorrea:
First pass at creating the form to allow the users to submit feedback related to biased search results.

Works towards https://github.com/pulibrary/orangelight/issues/3789


sdellis:
An update to the mockup that simplifies our initial approach to providing context for the user can be found [in the comments on the original ticket](https://github.com/pulibrary/orangelight/issues/3777#issuecomment-1814773585).

sandbergja (approved):
Thanks, group 3!

PR #3828, opened 2023-11-13
Remove highlight-query class from closing em tag

christinach:
Remove highlight-query class from closing em tag


rladdusaw (approved):
Thanks!

PR #3841, opened 2023-11-16
Added subjects and notes to displayed fields

rladdusaw:


kevinreiss (approved):
Should we also have tests for the scenarios where we have a notes field but not a subject field with higlighted or when neither field has highlighting? We don't need that to block this PR for the purposes of today's demo.

rladdusaw:
> Should we also have tests for the scenarios where we have a notes field but not a subject field

I'm going to work on that today we wanted to get this in before the round-up

PR #3634, opened 2023-07-10
Indicate to screen readers when a link opens in a new tab

ishasinha1:
Closes #3507

PR #3842, opened 2023-11-17
Make search results link open in new tab

kelynch:


app/views/catalog/_report_biased_results_form.html.erb
```diff
@@ -7,7 +7,7 @@
     You can help us address this issue by reporting any biased or harmful results you encounter on Library search tools by using the form below. If you would prefer to report this anonomously, you may leave the name end email fields blank.
     <br/>
     <br/>
-    You are reporting bias in the following <%= link_to 'search results', form.context %>
+    You are reporting bias in the following <%= link_to 'search results', form.context, target: '_blank' %>
```
sandbergja, line 10: Could we please also include an indication to screen reader users that this opens in a new tab?  Either in the link text itself, or we could add a little icon like in #3634?

sandbergja (changes requested):
Thanks, @kelynch !  Just one accessibility suggestion

app/views/catalog/_report_biased_results_form.html.erb
kelynch: Great idea!  I'll do this after this morning's demo.

sandbergja (approved):
Confirmed that VoiceOver now specifies that this link opens in a new tab.  Thanks, @kelynch !

PR #3867, opened 2023-12-01
i3232: Remove basic search bar and sort results from numismatics page

### LINKED ISSUE #3232: [Numismatics form] Two Keyword Search Boxes are confusing to users on Numismatics Search Page
### Current display
https://catalog.princeton.edu/numismatics 
Numismatics users go directly to the keyword search box at the top of the page (the Library Search Box), versus the keyword search box within the Numismatics search. Users are confused as to why coins are not the most relevant search results within that search and cannot find what they are looking for (example: a specific coin number).  The keyword search within the numismatics search options, however, works great for this purpose.  

### Desired display
I believe discussion with UX folks would be helpful here. User testing and a better idea why/if the catalog search is used at all with initial coin searches would probably be a question of mine (i.e "is the catalog search box even needed on this page, if the intent is to search coins only?"). 

### Notes

Demo of UX issues from 10/07/2022
https://user-images.githubusercontent.com/45043014/194570677-7121eb40-ef73-4dc8-bc40-828a5ee9d738.mp4

### Acceptance criteria:
- [x] Remove the 'sort results by' dropdown.
- [x] Remove the catalog search box from the numismatics search form view.
- [x] I can successfully use this feature with Voice Over in Safari.
- [x] I have checked this feature using one of the following browser extensions and confirmed that it does not add any level A or level AA issues:
    * axe dev tools
    * Lighthouse
    * Wave

sandbergja:
closes #3232

### DIFF
--- a/app/components/numismatics_search_form_component.html.erb
+++ b/app/components/numismatics_search_form_component.html.erb
@@ -38,12 +38,6 @@
   </div>
 
   <div class="search-submit-buttons clearfix col-sm-7">
-    <% if sort_fields_select %>
-      <div class="form-group sort-buttons pull-left">
-        <%= content_tag :label, t('blacklight.advanced_search.form.sort_label'), id: 'advanced-search-sort-label', for: 'sort', class: 'control-label' %>
-        <%= sort_fields_select %>
-      </div>
-    <% end %>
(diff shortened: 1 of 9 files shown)


christinach (approved):
Thanks @sandbergja !

PR #3869, opened 2023-12-04
Remove main search bar from advanced search form

### LINKED ISSUE #3760: Remove the search bar in the advanced search screen
### Current display

Search term is in both the basic keyword search box and the advanced keyword search box

### Steps to recreate

1. Do a keyword search in the catalog
2. Press the "Edit search" button

### Research and/or mockup needed

The current display is potentially confusing, since it presents users with two separate search forms both with their query, but which behave differently.  Perhaps we should not populate the keyword into the basic search form when editing a search.  Or perhaps we shouldn't show the basic search form at all when editing a search.

### Related ticket

#3232

christinach:
closes #3760

### DIFF
--- a/app/components/orangelight/advanced_search_form_component.html.erb
+++ b/app/components/orangelight/advanced_search_form_component.html.erb
@@ -49,12 +49,6 @@
   </div>
 
   <div class="search-submit-buttons clearfix col-sm-7">
-    <% if sort_fields_select %>
-      <div class="form-group sort-buttons pull-left">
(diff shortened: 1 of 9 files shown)


christinach:
Deployed on catalog-qa

kevinreiss (approved):

kevinreiss:
This branch seems to work well with the regular advanced search screen and the numismatics screen. Tested on catalog-qa.

PR #3871, opened 2023-12-04
Delete unused partial _advanced_navbar.html.erb

### LINKED ISSUE #3757: Check if we still need the shared/_advanced_navbar partial
@christinach and I noticed this comment, saying that it is possibly not being used:
https://github.com/pulibrary/orangelight/blob/main/app/views/shared/_advanced_navbar.html.erb#L19

christinach:
closes #3757


kevinreiss (approved):
Noting this bug is also present in this branch: https://github.com/pulibrary/orangelight/issues/3874. It's unrelated to the change here.

PR #3891, opened 2023-12-06
bundle update

sandbergja:
When deploying this to staging, I get the following in the nginx log:

```
Error: The application encountered the following error: You have already activated
base64 0.1.1, but your Gemfile requires base64 0.2.0. Since base64 is a default gem,
you can either remove your dependency on it or try updating to a newer version of
bundler that supports base64 as a default gem. (Gem::LoadError)
```

I think https://github.com/pulibrary/princeton_ansible/pull/4498 would resolve this issue, so we should push that over the finish line first.


christinach:
Ran bundle update with no issues in https://github.com/pulibrary/orangelight/pull/4596

PR #3892, opened 2023-12-07
Remove whitelist term

sandbergja:
Per our inclusive language norms (https://github.com/pulibrary/pul-it-handbook/blob/main/norms/inclusive_language.md)


christinach (approved):
🍄

PR #3931, opened 2024-01-10
Added config value with clearer screen reader text

### LINKED ISSUE #3375: Results-per-page dropdown on the search results page should have a better accessible name
@rladdusaw, @maxkadel and I found this while testing a different issue in VoiceOver.

### Expected behavior
When you open the "20 results per page" dropdown, VoiceOver on Firefox does not give a confusing name to the dropdown.

### Actual behavior

When you open the dropdown in firefox, voiceover announces "menu 20 per page 4 items".

### Steps to replicate

1. Open the catalog in firefox
2. Perform a search
3. Turn on voiceover
4. Tab until you get to the "20 per page" button
5. Press enter to open the dropdown
6. Press control + option + Right arrow to access the dropdown
7. Notice that voiceover announces "menu 20 per page 4 items", rather than something more expected like "menu Results per page 4 items"

### Impact of this bug

Adds confusion to using this with a screen reader.

This issue is also present in Blacklight itself, so we may want to open a PR there.

rladdusaw:
Resolves #3375 

There was a configuration that was not set for this element. We had discussed a possible upstream PR, but since it's able to be configured I don't think it's necessary.

### DIFF
--- a/config/locales/blacklight.en.yml
+++ b/config/locales/blacklight.en.yml
@@ -58,6 +58,7 @@ en:
       sort:
         label: 'Sort<span class="d-none d-xl-inline"> by %{field}</span>'
       per_page:
+        label: '%{count}<span class="sr-only visually-hidden"> results per page</span>'
         button_label: '%{count}<span class="d-none d-xl-inline"> per page</span>'
       start_over: 'Start over'
       cite: 'Cite'


christinach (approved):
Thanks @rladdusaw 🐈‍⬛

PR #3932, opened 2024-01-10
Use SameSite=strict for orangelight cookie

sandbergja:
This reduces the number of scenarios in which the cookie will be sent, providing better CSRF protection and reducing the size of the Cookies header in some cases.

Part of https://github.com/pulibrary/dacs_handbook/issues/154


christinach (approved):
Thanks @sandbergja 🍄

Issue #3975, opened 2024-03-13
Repeated Timeout Errors Sending Mail. 


kevinreiss:
### Expected behavior
When a user places a request the system sends out one or more emails notifying staff of the request and confirming with the user that the request was placed. 

### Actual behavior
User submits a request for an item. The system displays an error message to the user because the send_mail action for one of the notification emails has generated a Timeout error. However, the API calls to place the request actually succeed so the request is successfully accepted by the desired systems (i.e. Alma, ILLiad, SCSB). The displayed failure notice in the UI then leads the user to attempt to place the request again which results in another error displayed because the processing systems think it's a duplicate request. 

### Steps to replicate
This an intermittent issue that is difficult to reliably replicate. This issue has increased in frequency in the last ten days, it's happening roughly 3-5 times a day. 

### Impact of this bug
This provides a degraded user experience.  When the timeout error happens the UI tells the user the request has encountered an error and they try to request the item again. Since the API calls are actually successfully, the second attempt fails and generates a notification to the recapproblems email queue saying the user attempted to request a barcode that is not available. 

### Honeybadger link and code snippet, if applicable
See most recent honeybadger errors that fit this pattern, there are three distinct errors that appear all in a send_mail method. 

https://app.honeybadger.io/projects/54399/faults/99737318 - Net::ReadTimeout
https://app.honeybadger.io/projects/54399/faults/101572128 - Net::OpenTimeout
https://app.honeybadger.io/projects/54399/faults/105534722 - Net::ReadTimeout

kevinreiss:
This also seems to be happening in approvals:

https://app.honeybadger.io/projects/67941/faults/105117183
https://app.honeybadger.io/projects/67941/faults/105650358

kevinreiss:
There are multiple permutations of this issue that can be seen in honeybadger in recent weeks. All of the errors starting with Net::ReadTimeout seem to be a variation of the same thing. The system reaches a point where it is trying to send mail and the action logs and error. Looking at the pattern they have occurred occasionally over a long period of time, but switched to happening multiple times per day around March 7th.

christinach:
One approach:
Check for the Net::ReadTimeout error and when it happens , retry sending the email.

kevinreiss:
John K. Provided this sample from the 3/18 mail log

```
[root@lib-ponyexpr systems]# tail -4000 /var/log/maillog | grep "did not issue MAIL/EXPN/VRFY/ETRN"
[...]

PR #3981, opened 2024-03-20
Provide 10 second defaults on open or read calls for SMTP.

kevinreiss:
Attempt to mitigate #3975.


christinach (approved):
Thanks @kevinreiss

PR #3999, opened 2024-04-08
i3998: Display selected facets when editing a facet search

### LINKED ISSUE #3998: Edit Search: facets that you selected in advanced search do not remain
### Expected behavior
If I select an facet on the advanced search form, then search, then edit my search, the facet value should appear selected on the combobox.

### Actual behavior

The combobox is empty when I edit the search.

### Steps to replicate

1. Go to https://catalog.princeton.edu/advanced
2. Select a format: Audio
3. Press Search
4. Press Edit Search
5. Notice that the format combobox no longer says Audio
6. Notice the console error

### Impact of this bug

Small, since it does keep the Audio facet in subsequent searches, and the advanced form indicates elsewhere that the Audio facet is selected.

This happens because we are comparing the `.text` and `.textcontent` of an element, which may not be the same if the element's text contains multiple spaces next to each other.

sandbergja:
closes #3998

### DIFF
--- a/app/javascript/orangelight/multiselect_combobox.es6
+++ b/app/javascript/orangelight/multiselect_combobox.es6
@@ -67,7 +67,7 @@ export default class MultiselectCombobox {
 
     #applySelections() {
         this.hiddenSelect.querySelectorAll('option:checked').forEach(selectedOption => {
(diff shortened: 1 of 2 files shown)


christinach (approved):
Thanks @sandbergja

PR #4000, opened 2024-04-08
Minor performance refactor for advanced search forms

sandbergja:
When running rubyprof on the numismatics screen locally, we noticed that
  * Object#blank is called 39863 times
  * Array#each is called 10480 times
  * BasicObject#! is called 97717 times

...and this is using the local development data set, which has very few coins.

When looking at the stack traces for some of these calls, they all came from the `facet_value_checked?` method in Blacklight Advanced Search gem, many of them from it parsing filters from the URL params, which is memoized so re-using the BlacklightAdvancedSearch::QueryParser saves us re-doing that work each time.

Re-using the `BlacklightAdvancedSearch::QueryParser` object reduces the number of calls we have to make to these methods to 22164, 7863, and 62083 respectively.

### DIFF
--- a/app/views/advanced/_facet_as_select.html.erb
+++ b/app/views/advanced/_facet_as_select.html.erb
@@ -3,6 +3,7 @@
       <% if display_facet.name == 'advanced_location_s' %>
         <%= render 'location_code_facet', display_facet: %>
       <% else %>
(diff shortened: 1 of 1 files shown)


christinach (approved):
Thanks @sandbergja

PR #4026, opened 2024-05-30
Explicitly turn off enable_starttls for SMTP

sandbergja:
In testing with the rails console, @kayiwa found that we need `enable_starttls: false` in order to send mail via the new lib-ponyexpr-prod.  See https://github.com/pulibrary/princeton_ansible/pull/4980#issuecomment-2140181019.  We may decide to use TLS for this connection in the future.

Using the rails console, I confirmed that I can still send mail through the old lib-ponyexpr with an explicit `enable_starttls: false`.


christinach (approved):
Thanks @sandbergja and @kayiwa

PR #4038, opened 2024-06-18
Add check to show deaccessioned items as being not in clancy

rladdusaw:


kevinreiss (approved):
In addition to the "green" test suite Ryan and I also manually tested in staging the various marquand recap and stacks request scenarios and things were working as far as we can tell.

PR #4051, opened 2024-07-19
Following changes in pul_solr, update the sol-conf

christinach:


sandbergja (approved):
Thanks, @christinach 🐬

PR #4063, opened 2024-07-25
Update commit hash for health monitor dependency

sandbergja:
`bundle install` was failing for me locally on main.  `bundle update health-monitor-rails` fixed it, by finding the most up-to-date commit hash for our fork.


christinach (approved):
Thanks @sandbergja !

PR #4064, opened 2024-07-25
Refactor new forms to not inherit from MailForm::Base

sandbergja:
We were only using `MailForm::Base` for its `spam?` method, which is simple enough to implement ourselves.


christinach (approved):
Thanks @sandbergja

Issue #3909, opened 2023-12-15
Use Libanswers API for Catalog Feedback Form 


kevinreiss:
### What maintenance needs to be done?
Convert catalog support webform to submit tickets via the libanswers api rather than email them to libanswers queue directly. 

### Level of urgency
- [ ] High
- [ ] Moderate
- [x] Low

### Why is this maintenance needed?
- This can potentially make libanswers less vulnerable to spam from the catalog support webforms. 

### Acceptance criteria
 - [ ] User submissions go into the libanswers catalog feedback queue
 - [ ] User submissions should record an email address for the submitter so Libanswers can send replies to the person who sent in the inquiry.
 - [ ] Review the form's spam protection. We want to make sure we don't spam the queue via the API.
 - [ ] Remove the email address from the queue in libanswers so that spam cannot be sent there anymore.

### Implementation notes, if any
If this approach looks like it will work we should replicate it for the other feedback options on the site's record page (i.e. Harmful Content Report). 

 

### Documentation
https://ask.springshare.com/libanswers/faq/1669

sandbergja:
A few questions on this:

-  The API gives the opportunity to pass along the user's user agent and IP address.  Do we want either of those in libanswers for these tickets?
-  Do we want to indicate in some way on the ticket that it was created via the form and/or the API?
-  Currently, we use the subject "Princeton University Library Catalog Feedback Form" for any ticket that comes in via the feedback form.  Do we want to keep this behavior?

maxkadel:
I'm hesitant about passing along the user's IP address based on privacy concerns, especially for the harmful content feedback forms. But I'm also not sure how collecting IP addresses for this specific purpose compares with other ways we pass along or log IP addresses.

I wonder if we could add a denylist that responded to IP addresses without connecting them with specific user feedback? I recognize that might be harder to populate without collecting the IP addresses in the first place.

sandbergja:
@maxkadel - thanks!  Digging a little deeper, I found that we are already passing along the IP address, user agent, and referer url.  So maybe it's time to reevaluate if that has been useful, or if we can move away from including IP addresses.

A denylist definitely seems possible -- the IP address is available in the `FeedbackController` as `request.remote_ip`.

sandbergja:
In today's standup, we decided:

- We should remove the IP address
- We should indicate that it came via the API
- We can keep the subject "Princeton University Library Catalog Feedback Form"

maxkadel:
Done! 🎉

[...]

PR #4066, opened 2024-07-25
Refactor FeedbackForm to not use MailForm

sandbergja:
This refactor makes the feedback form more similar to the other forms.  It also will enable us to more easily swap out the email functionality for Libanswers API ticket creation

Helps with #3909


christinach (approved):
Thanks @sandbergja

PR #4067, opened 2024-07-26
Remove mail_form gem

sandbergja:
It is no longer needed (after #4066 and #4064).

This includes a refactor of the Report Biased Results form, which is not in production and is behind a feature flipper.  That form is not complete (for example, the form template does not have a honeypot field to catch spam bots), I did not attempt to fix it, just kept the incomplete behavior the same.


maxkadel (approved):
Thanks @sandbergja! 🍄  - always love a PR that takes out something we're not using!

PR #4070, opened 2024-07-26
Update to main version of health-monitor-rails gem

maxkadel:
Already merged to the gem! (not released yet)


maxkadel:
aaaand, it's released in the gem. Sheesh this guy is fast.

PR #4073, opened 2024-07-26
Use libanswers API, rather than email, to create tickets in libanswers for the FeedbackForm

sandbergja:
Does most of the work of #3909 (there is a bit of cleanup to be done after this is merged and deployed to all environments)


christinach (approved):
thanks @sandbergja 🍄

PR #4077, opened 2024-07-29
Add status monitors for vendor APIs

### LINKED ISSUE #4053: Add Monitors for Vendor APIs the Catalog Integrates with
Add monitor for external websites the catalog depends on. 

### Level of urgency
- [ ] High
- [x] Moderate
- [ ] Low

- Better monitoring of services catalog features depends on. 

### Acceptance criteria
 - [ ] Monitor for ILLiad API
 - [ ] Monitor for Aeon Open URL request endpoint
 - [ ] Monitor for Stackmap endpoint. 
 - [ ] Monitor for SCSB API

rladdusaw:
Resolves #4053

### DIFF
--- a/config/initializers/health_monitor.rb
+++ b/config/initializers/health_monitor.rb
@@ -7,7 +7,11 @@ Rails.application.config.after_initialize do
       c.url = Blacklight.default_index.connection.uri.to_s
     end
 
+    config.add_custom_provider(AeonStatus)
     config.add_custom_provider(BibdataStatus)
+    config.add_custom_provider(IlliadStatus)
+    config.add_custom_provider(ScsbStatus)
+    config.add_custom_provider(StackmapStatus)
 
     # Make this health check available at /health
     config.path = :health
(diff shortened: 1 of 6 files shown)


config/initializers/health_monitor.rb
```diff
@@ -7,7 +7,11 @@
       c.url = Blacklight.default_index.connection.uri.to_s
     end
 
+    config.add_custom_provider(AeonStatus)
```
maxkadel, line 10: Should all of these new providers be marked as critical? (i.e., should the load balancer not serve the app if any of them are down?) I would argue that all of the newly added ones should report, but not be considered critical.

rladdusaw: Made the change. Couldn't get it to work in the config, but was able to make them non-critical in the custom providers.

sandbergja (approved):
Thanks, @rladdusaw ! 🐈‍⬛

Issue #4054, opened 2024-07-24
Mobile Usability Issue - Largest Contentful Paint takes more than 4 seconds on Mobile


maxkadel:
### What maintenance needs to be done?
Ensure Largest Contentful Paint (LCP) takes less than 4 seconds for record show pages in mobile view.

### Level of urgency
- [ ] High
- [x] Moderate
- [ ] Low

### Why is this maintenance needed?
- From the Google Search Console (GSC) - [LCP issue: longer than 4s (mobile)](https://search.google.com/search-console/core-web-vitals/drilldown?resource_id=sc-domain%3Acatalog.princeton.edu&item_key=CAMQAhgD).

> URLs with this issue provide a poor page experience

Example URLs:
- https://catalog.princeton.edu/catalog/997169603506421
- https://catalog.princeton.edu/catalog/SCSB-9333941
- https://catalog.princeton.edu/catalog?q1=%D8%B3%D9%84%D8%B3%D9%84%D8%A9+%D8%A7%D9%84%D8%AA%D8%B1%D8%A7%D8%AB+%D8%A7%D9%84%D8%B9%D9%84%D9%88%D9%8A+%D8%9B&f1=in_series&search_field=advanced

### Acceptance criteria
 - [ ] For mobile, the LCP consistently takes less than 4 seconds on links above
 - [ ] Once the fix is deployed to production, click the "validate fix" button on the [Google Search Console page](https://search.google.com/search-console/core-web-vitals/drilldown?resource_id=sc-domain%3Acatalog.princeton.edu&item_key=CAMQAhgD)

### Implementation notes, if any
- [Article linked from GSC about LCP](https://web.dev/articles/lcp)

sandbergja:
We have sped up the mobile LCP (as measured by lighthouse) by 4 seconds!  We also pressed the Validate Fix button.  We'll find out in 28 days if the search console agrees that the show page is good now.

PR #4082, opened 2024-07-30
[#4054] defer loading for some javascript

sandbergja:
Adding 'defer' tells the browser not to wait for this javascript before it works on rendering the page.

For me running the catalog locally, it took Lighthouse's mobile First Contentful Paint  on a show page down from 20.3 seconds to 16 seconds

Helps with #4054


christinach (approved):
thanks 🍄

PR #4084, opened 2024-07-31
[#4054] Add lighthouse to CI

sandbergja:
Includes an assertion that the Largest Contentful Paint is under 21 seconds, which is super high! This is partly because:
* we are not running in the production Rails environment (which would be better)
* partly because of lighthouse's throttling, which emulate a device with a slow CPU and a slow 4G connection
* partly because of performance problems in the catalog (which we actually want to measure)

I think it would be nice to gradually reduce the number of seconds in our assertion as we make performance improvements in the application.

Helps with #4054.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -131,6 +131,33 @@ jobs:
         - store_test_results:
             path: ~/rspec
 
+  lighthouse:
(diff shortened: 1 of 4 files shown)


christinach (approved):
thanks @sandbergja

PR #4086, opened 2024-07-31
[#4054] Don't serve jquery v1.12.4 outside of requests

sandbergja:
Previously, we were asking clients to download two different versions of jquery, but they only need one.  They may need both for requests, due to the datatables dependency, that could use some further investigation.  This commit only removes the old jquery for non-requests pages.

Helps with #4054

### DIFF
--- a/app/assets/javascripts/orangelight.js
+++ b/app/assets/javascripts/orangelight.js
@@ -1,5 +1,3 @@
-//= require 'jquery'
-
 $(function () {
   //link highlighting of hierarchy
   $('.search-subject, .search-name-title, .search-title').hover(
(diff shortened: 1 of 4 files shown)


christinach (approved):
Thanks @sandbergja

christinach (approved):
Thanks @sandbergja 🍄

app/assets/javascripts/orangelight.js
```diff
@@ -1,5 +1,3 @@
-//= require 'jquery'
```
christinach, line 1: 👍🏻

lighthouserc.js
```diff
@@ -2,7 +2,7 @@ module.exports = {
   ci: {
     assert: {
       assertions: {
-        'largest-contentful-paint': ['error', { maxNumericValue: 21000 }],
+        'largest-contentful-paint': ['error', { maxNumericValue: 20000 }],
       },
```
christinach, line 6: 🎉

PR #4089, opened 2024-07-31
Improve lighthouse instructions

sandbergja:
Thanks to @christinach for trying out the current instructions


christinach (approved):
Thanks @sandbergja !!

PR #4090, opened 2024-07-31
Remove modernizr

sandbergja:
We don't appear to be using any of its browser feature-detection features at the moment.

This reduces the unminified application.js by 30 KB.

Helps with #4054


christinach (approved):
thanks @sandbergja

PR #4091, opened 2024-08-01
[#4054] Only load font-awesome CSS once

sandbergja:
Prior to this commit, we loaded font-awesome CSS twice: once via sprockets, and once from a CDN.

Helps with #4054


christinach (approved):
Thanks @sandbergja

PR #4101, opened 2024-08-05
Remove unnecessary holding_request_block calls from views

sandbergja:
Two views call `holding_request_block`, then don't actually use the returned values for anything.  `_show_other_versions` is the slowest partial in the show page, according to the datadog APM, so this is a low-hanging fruit to make it faster.

Helps with #4054


christinach (approved):
Thanks @sandbergja

PR #4103, opened 2024-08-05
[#4054] Only ship the bootstrap js that we actually use

sandbergja:
Helps with #4054.

This reduces the unminified, uncompressed application.js from 883 kB to 811 kB.

Lighthouse says that the application.js file takes 739 ms to execute in production (when emulating a slow phone CPU). This should reduce it a little.


christinach (approved):
Thanks @sandbergja

PR #4109, opened 2024-08-06
Namespace Illiad classes

maxkadel:
supports upgrading to rails 7

### DIFF
--- a/lib/orangelight/illiad_account.rb
+++ b/lib/orangelight/illiad_account.rb
@@ -2,38 +2,39 @@
 
 require 'faraday'
 require 'faraday-cookie_jar'
-
-class IlliadAccount
-  def initialize(patron)
-    @barcode = patron['barcode']
-    @last_name = patron['last_name']
-    @patron_id = patron['patron_id']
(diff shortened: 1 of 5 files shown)


lib/orangelight/illiad_account.rb
```diff
@@ -2,38 +2,39 @@
...
+          req.headers['ApiKey'] = @illiad_api_key
+        end
+      rescue Faraday::ConnectionFailed
+        Rails.logger.info("Unable to Connect to #{@illiad_api_base}")
+        return false
       end
-    rescue Faraday::ConnectionFailed
-      Rails.logger.info("Unable to Connect to #{@illiad_api_base}")
-      return false
+      response.success?
```
christinach, line 27: @maxkadel did you mean to move this?

maxkadel: Everything is just indented one so the git compare is funky. It hasn't actually moved as far as I can tell.

christinach (approved):
Thanks @maxkadel !

PR #4112, opened 2024-08-06
Match black colors in the footer

sandbergja:
Prior to this commit, the lux footer had a background color of #121212, while the surrounding container had a background color of #040404, which clashed a little bit.  This commit uses #121212 for both blocks.

Before:
The catalog footer, with two different shades of black in the background

After:
The catalog footer, but with only one shade of black in the background


christinach (approved):
Thanks @sandbergja

kevinreiss (approved):

Issue #4088, opened 2024-07-31
Use LUX to Provide Site Header


kevinreiss:
### Desired display
Unchanged except we now use LUX to render the site's header. 

### Acceptance criteria

- [ ] LUX provides the site's header
- [ ] We provide the same "user menu" menu options (i.e. Help, Feedback, Your Account) we do now. 
- [ ] Test "Your Account" Menu Function for "Library Account"
- [ ] Test "Your Account" Menu Function for "Bookmarks"
- [ ] Test "Your Account" Menu Function for "Search History"
- [ ] I can successfully use this feature with Voice Over in Safari.
- [ ] I have checked this feature using one of the following browser extensions and confirmed that it does not add any level A or level AA issues:
    * axe dev tools
    * Lighthouse
    * Wave

PR #4116, opened 2024-08-07
Use prettier for the spec directory

sandbergja:
This helps us maintain consistent formatting for javascript tests and JSON fixtures

Also, configure prettier to require semicolons

Part of #4088 -- I'm trying to break a large PR into some smaller changes

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,6 +1,5 @@
 node_modules
 .gitignore
-spec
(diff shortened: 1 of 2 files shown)


christinach (approved):
Thanks @sandbergja @maxkadel @christinach 
🍄 🦄 🐬

PR #4117, opened 2024-08-07
[#4088] Use Lux to provide header

### LINKED ISSUE #4088: Use LUX to Provide Site Header

sandbergja:
* Lux's header is wrapped in a vue component that provides the header logic that we previously had in view partials and jquery.
* We add a route /sign_out so that we can link to it as a GET request.  This follows how approvals handles this.  In theory, we could have kept the DELETE /users/sign_out endpoint only and used Lux's method property to specify that this was a DELETE request.  In practice, this had a bug when the user signed in, then signed out without having navigated to another page, Rails was expecting a different CSRF token, so the signout failed.
* Also updates some tests to have js: true, so they can render the new header
* Adds vue/test-utils

closes #4088


app/assets/javascripts/orangelight.js
```diff
@@ -122,19 +122,4 @@ $(function () {
   $('.clickable-row').on('click', function () {
     window.location = $(this).data('href')
   })
-
```
sandbergja, line 125: This is re-implemented in the vue component below

config/routes.rb
```diff
@@ -66,6 +66,7 @@
   devise_scope :user do
     get '/users/signup' => 'devise/registrations#new', :as => :new_user_registration
     post '/users' => 'devise/registrations#create', :as => :user_registration
+    get "sign_out", to: "devise/sessions#destroy"
```
sandbergja, line 69: Adding a route to sign out with an HTTP GET request (previously, you had to send an HTTP DELETE request)

PR #4119, opened 2024-08-09
Eager load assets in CI

maxkadel:
- We have seen issues with this on other apps, good practice


sandbergja (approved):
Thanks, @maxkadel 🦄

PR #4120, opened 2024-08-09
[OrangelightHeader]Add LuxHeader in a container to get the Bootstrap width.

christinach:
[OrangelightHeader]Add header in a container to get the Bootstrap width.
[OrangelightHeader]Add container in a parent class with background-color var(--color-gray-100)

# Notes
Lux is using a different max-width to match with the max-width of the new Library Website. 
Bootstrap's container width is smaller. 
We move the header the same way we moved the footer in a container so that they are aligned with the main content of the app which is part of a different container.

## Lux Header not in a  container
Lux Header not in a  container

## Lux Header in a container:
Lux Header in a container


maxkadel (approved):
Thanks @christinach!

PR #4125, opened 2024-08-19
[#4123] Avoid layout shift and quickly changing colors in the header

### LINKED ISSUE #4123: Lux header flashes
### Expected behavior

The header does not flash and suddenly push down the rest of the content of the page while the page is loading.

### Actual behavior

It does flash, every single time you load a new page in the catalog

### Steps to replicate

1. Open your browser's dev tools
2. Go to the network tab
3. To make the flash and layout shift easier to see (it can happen very fast!), choose a throttling option to emulate a slower internet connection
4. Go to a catalog page, and watch for the layout shift and sudden color change.

### Impact of this bug

From a libanswers report:

> I know it's still in development, but the new(?) black header atop the Library Catalog appears to "flash" once as it appears, as any page loads; which I predict will be disturbing to users. Every click to view a search result, or go back a page, etc., causes this distraction which was not there before.

* We should reserve some space for the Lux Header at the top of the screen, and have its background color be the same as the Lux Header's background color, so there is no sudden color change or layout shift on page load.

sandbergja:
This commit moves the .pul_header container out of the Vue component (which is rendered after the rest of the page), and into erb and the application CSS, which will block rendering.  It also gives the container an explicit height.

This prevents the situation where the page renders with light colored content at first, then changes to very dark gray and shifts the layout once the Vue component loads, which looks like a distracting flash.

Also, add a threshold for layout shift to our lighthouse check, to hopefully guard against regressions.

closes #4123

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -43,6 +43,7 @@
 @import 'components/font-face';
 @import 'components/front-page';
 @import 'components/footer';
+@import 'components/header';
 @import 'components/header--secondary';
 @import 'components/lists';
 @import 'components/nav--accounts';
--- a/app/assets/stylesheets/components/header.scss
+++ b/app/assets/stylesheets/components/header.scss
@@ -1,0 +1,8 @@
+/* The container that the lux header will be in.
+   It is the same background color and height as
+   the lux header, so that when the lux header loads,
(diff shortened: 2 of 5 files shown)


christinach (approved):
Thanks @sandbergja 🍄

kevinreiss (approved):

PR #4143, opened 2024-08-26
Remove stringex. serves 5.8MiB of memory

christinach:
Saves 5.8 MiB of memory when app starts
checked it with derialed_benchmark gem and `bundle exec derailed bundle:mem`

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -80,7 +80,6 @@ gem 'sneakers'
 gem 'sprockets-es6'
 # For call-stack profiling flamegraphs
 gem 'stackprof'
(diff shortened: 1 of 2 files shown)


christinach:
we're using https://github.com/pulibrary/stringex/blob/2eeb842fafad1f4ea14cf8a8f7be8d9d3bd1b874/lib/stringex/string_extensions.rb#L145

christinach:
@sandbergja @rladdusaw We should find a different solution. The last time this gem was updated was 10 years ago.

sandbergja:
closing in favor of #4613

PR #4144, opened 2024-08-27
Link to series advanced search based on query dsl

### LINKED ISSUE #4140: `more_in_this_series_link` does not work without Blacklight Advanced Search gem
### Expected behavior
The `more_in_this_series_link` finds results whether or not the advanced search gem is used

### Actual behavior
The search only works when the syntax from the Blacklight Advanced Search gem is used.

### Steps to replicate
Hard to do without removing the gem, but if you look at the syntax, you will note that it uses the `q1=foo` syntax, rather than the newer `clause[1]` syntax

### Impact of this bug
Cannot remove the Blacklight Advanced Search gem and upgrade to Blacklight 8 without fixing

Currently in `app/processors/orangelight/series_link_processor.rb`

Tests in both `spec/processors/orangelight/series_link_processor_spec.rb` and `spec/features/more_in_this_series_spec.rb`

Have one [option in a branch](https://github.com/pulibrary/orangelight/compare/i3679_remove_advanced_search_gem?expand=1), but this works only with the json query dsl turned on, so would need to be put with the current code behind a flipper

maxkadel:
Closes #4140

### DIFF
--- a/app/processors/orangelight/series_link_processor.rb
+++ b/app/processors/orangelight/series_link_processor.rb
@@ -20,11 +20,25 @@ module Orangelight
       # @param title [String] the title of the series
       # @return [String] the link markup
       def more_in_this_series_link(title)
-        no_parens = authorized_form_of_title(title).gsub(/[()]/, '')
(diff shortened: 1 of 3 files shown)


christinach (approved):
Thanks @maxkadel ! 🦄

PR #4150, opened 2024-08-28
Upgrade to latest Blacklight 7.x release

maxkadel:
This includes a bug fix for a bug that was making it so we made all our solr requests twice!


christinach (approved):
Thanks @maxkadel !

PR #4161, opened 2024-08-29
Update left_anchor test with two exampls of AND and NOT

christinach:


sandbergja (approved):
Thanks, @christinach 🐬 !  It's much clearer when this test is looking for an actual title.

Issue #3387, opened 2023-01-30
Upgrade Blacklight Gem to 8.0


christinach:

kevinreiss:
Current Draft PR: https://github.com/pulibrary/orangelight/pull/4516/files. Team will do some testing of this in QA on 11/4.

christinach:
DONE!

PR #4162, opened 2024-08-29
[#3387] Refactor format facet into a View Component

sandbergja:
When updating the blacklight gem to 8, our view partial mysteriously did not get the correct local variables passed to it.  This commit avoids that issue by refactoring the view partial into a View Component, which gets all the data it needs in both blacklight 7 and 8.

Helps with #3387


christinach (approved):
thanks @sandbergja

PR #4164, opened 2024-08-30
[#3387] Use if: false to hide fields from the search results screen

sandbergja:
Blacklight 8 seems to no longer support show: false for index fields.
Fortunately, we can use if: false to achieve the same effect.

helps with #3387


sandbergja:
closing in favor of #4166

PR #4165, opened 2024-08-30
Address fixture_path deprecation

sandbergja:
Addresses the following deprecation, which is filling up the terminal output when running tests:

```
DEPRECATION WARNING: TestFixtures#fixture_path is deprecated and will be removed in Rails 7.2. Use #fixture_paths instead.
```


christinach (approved):
thanks 🍄

PR #4166, opened 2024-08-30
[#3387] Refactor "show" option for index fields

sandbergja:
Prior to this commit, we had a custom configuration option for index fields called `show`.  When `show` is true, the field displays in both the HTML and JSON search results.  When it is false, the field only displays in the JSON search results.

Previously, we implemented this behavior in `_index_default.html.erb`. This was later slightly refactored into a view component that was called by `_index_default.html.erb`.  However, `_index_default.html.erb` is no longer used in Blacklight 8, so our logic gets ignored in Blacklight 8.

This commit moves the behavior to the catalog controller, using a new private method named `json_request?`.

Helps with #3387


christinach (approved):
thanks @sandbergja

PR #4169, opened 2024-08-30
Wrap our current index page partials in a subclass of Blacklight::DocumentComponent

sandbergja:
Blacklight offers two ways to customize how your search results display:
* `config.index.partials`, where you can specify specific partials you want to show in the view directory
* `config.index.document_component`, where you can specify a subclass of `Blacklight::DocumentComponent` that displays the item

In Blacklight 8, the default behavior is to use `Blacklight::DocumentComponent` directly, which renders all the relevant metadata for the result, and then display the partials you've configured in `config.index.partials`.

The catch is that the partials we've configured in `config.index.partials` _also_ display the metadata, so we end up displaying all the metadata twice.

This commit subclasses `Blacklight::DocumentComponent` to display only our partials.

Also, refactor `_index_header_default` into a subclass of `Blacklight::DocumentTitleComponent`, since it made it easier to test.


app/assets/stylesheets/components/search-results.scss
```diff
@@ -13,7 +13,8 @@
   margin-bottom: 0;
 }
 
-.documents-list h3 {
```
sandbergja, line 16: Making this selector more specific so that Blacklight 8's CSS doesn't override it

app/components/index_document_component.html.erb
```diff
@@ -0,0 +1,17 @@
...
+  id: @id,
+  data: {
+    'document-id': @document.id.to_s.parameterize,
+    'document-counter': @counter,
+  },
+  itemscope: true,
+  itemtype: @document.itemtype,
+  class: classes.flatten.join(' ') do %>
+    <div class="row">
+      <div class="record-wrapper">
```
sandbergja, line 11: A new view component that houses our search results partials

app/components/index_title_component.html.erb
```diff
@@ -0,0 +1,11 @@
+<header class="documentHeader">
```
sandbergja, line 1: A new view component to show document titles on the search result page (a replacement for app/views/catalog/_index_header_default.html.erb)

christinach (approved):
Thanks @sandbergja !

PR #4170, opened 2024-09-03
Default to Blacklight 8 features on in test and CI

### LINKED ISSUE #4141: Default to Blacklight 8 features being on for test and CI
Right now, the default for the Blacklight 8 features is off in all environments. We should default to having them on so that we do not have any surprise regressions. I have seen a few [articles similar to this one](https://medium.com/samsara-engineering/why-you-should-invert-your-feature-flags-in-tests-c629531f630), which advocates for defaulting to having your new features default to on for CI.

### Level of urgency
- [ ] High
- [x] Moderate
- [ ] Low

- In order to ensure we are prepared for the Blacklight 8 upgrade, we should have our tests default to having related features "on" for test and CI

### Acceptance criteria
 - [ ] The Blacklight 8 features are "on" in test and CI environments, unless explicitly turned "off" (right now we have the opposite behavior)
 - [ ] It does not slow production performance
 - [ ] It is easy to remove once these features are deprecated
 - [ ] All tests pass

I experimented with doing this with a proc, which seemed to work ok

```ruby
# config/features.rb
group :blacklight_8 do
  feature :json_query_dsl,
  description: "When on / true, use the JSON query DSL for search fields in the advanced search.  When off / false, use query params",
  default: proc { ENV["RAILS_ENV"] == "test" || ENV["RAILS_ENV"] == "CI" ? true : false }

  feature :view_components_numismatics,
  description: "When on / true, use the built-in advanced search form for numismatics.  When off / false, use the traditional one",
  default: proc { ENV["RAILS_ENV"] == "test" || ENV["RAILS_ENV"] == "CI" ? true : false }

  feature :view_components_advanced_search,
  description: "When on / true, use the built-in advanced search form.  When off / false, use the traditional one",
  default: proc { ENV["RAILS_ENV"] == "test" || ENV["RAILS_ENV"] == "CI" ? true : false }
end
```

maxkadel:
Closes #4141

### DIFF
--- a/app/models/engineering_search_builder.rb
+++ b/app/models/engineering_search_builder.rb
@@ -5,6 +5,7 @@ class EngineeringSearchBuilder < SearchBuilder
   self.default_processor_chain += [:switch_request_handler]
 
   def switch_request_handler(solr_parameters)
+    Rails.logger.warn("The engineering search does not work with the json_query_dsl") if Flipflop.json_query_dsl?
     solr_parameters[:qt] = "engineering_search"
   end
 end
--- a/spec/features/advanced_searching_spec.rb
+++ b/spec/features/advanced_searching_spec.rb
@@ -60,18 +60,24 @@ describe 'advanced searching', advanced_search: true do
     expect(page).to have_content('Aomen')
   end
 
-  it 'can exclude terms from the search' do
-    visit '/advanced'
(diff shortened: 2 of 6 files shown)


sandbergja (approved):
Thanks, @maxkadel 🦄

PR #4186, opened 2024-09-03
[#3387] Remove Blacklight::SearchFields

sandbergja:
This module is deprecated in Blacklight 7 and removed in Blacklight 8.

We only use it when setting up a test case, and the test still passes without it.  Also, remove a comment that refers to this module.

Helps with #3387


christinach (approved):
🍄

PR #4188, opened 2024-09-04
Capybara improvements

maxkadel:
- Make RUN_IN_BROWSER behavior consistent between feature and system specs
- When system test using javascript fails, save the screenshot where folks can see it (may help with debugging intermittent CI failures in future)
  - I dropped the commit, but created a purposefully failing system test using javascript, so that it could make a screenshot, which you can see in the [artifacts from this CI run](https://app.circleci.com/pipelines/github/pulibrary/orangelight/4907/workflows/1609144f-455f-43f3-a9f2-cf6686cffab1/jobs/18033/artifacts) (should only be available for 30 days)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -119,8 +119,6 @@ jobs:
           name: Index Test Data
           command: bundle exec rake pulsearch:solr:index
       - ruby/rspec-test
-      - store_test_results:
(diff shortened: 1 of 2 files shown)


christinach (approved):
Thanks @maxkadel

Issue #4175, opened 2024-09-03
Run Reek on the Requestable Class 


kevinreiss:
### What maintenance needs to be done?
 The reek score of this file is currently 20. 

### Level of urgency
- [ ] High
- [ ] Moderate
- [ ] Low

### Why is this maintenance needed?
- The file has a very high complexity score.

### Acceptance criteria
 - [ ] Warnings are lowered to 10 or less. 

### Implementation notes, if any
```reek app/models/requests/requestable.rb``` to see current report.

sandbergja:
The Requestable class is now down to 6 reek warnings, so I'm closing this ticket.

PR #4205, opened 2024-09-04
[#4175] Extract Requestable::Holding class

sandbergja:
Rather than passing around the holding as a hash throughout the application, this commit begins an incremental refactor to pass around an object that can answer questions about a holding from the `holdings_1display`. This will eventually mean that other classes won't need to know the exact internal structure of the holding hash, they can just ask `Requestable::Holding` for that information.

For now, though, this is just an incremental step, extracting a few methods from `Requestable` into `Requestable::Holding`, and using the hash elsewhere in the application.

This reduces the reek count in `Requestable` by 2.

Helps with #4175

### DIFF
--- a/app/models/requests/requestable.rb
+++ b/app/models/requests/requestable.rb
@@ -2,7 +2,6 @@
 module Requests
   class Requestable
     attr_reader :bib
-    attr_reader :holding
     attr_reader :item
     attr_reader :location
     attr_reader :call_number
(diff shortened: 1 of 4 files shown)


christinach (approved):
Thanks @sandbergja 🍄

christinach (approved):
Ευχαριστώ! @sandbergja !

christinach (approved):
Thanks @sandbergja !

PR #4206, opened 2024-09-04
[#4175] Simplify collection code checks

sandbergja:
helps with #4175

Reduces reek warnings by 3 by using `Item`'s existing `collection_code` method


christinach (approved):
thanks 🍄

Issue #4178, opened 2024-09-03
Run Reek on Request Class


kevinreiss:
### What maintenance needs to be done?
 The reek score of this file is currently 53. 

### Level of urgency
- [ ] High
- [ ] Moderate
- [ ] Low

### Why is this maintenance needed?
- The file has a very high complexity score.

### Acceptance criteria
 - [ ] Warnings are lowered to 40 or less. 

### Implementation notes, if any
```reek app/models/requests/request.rb``` to see current report.

sandbergja:
The reek score has been reduced to 40, so we can close this now.  Thanks, @christinach ! 🐬

PR #4213, opened 2024-09-04
[#4178] Don't call loc["library"] twice

christinach:
this change brings down reek count from 53 to 50


sandbergja (approved):
Thanks, @christinach 🐬

PR #4215, opened 2024-09-04
Uncomment and delete tests to reflect that campus is now open

sandbergja:
One of these tests said to uncomment it when campus is open post-COVID, another said to delete it when campus is open.  This commit does so.


christinach (approved):
🍄

PR #4216, opened 2024-09-04
[#4178] Remove commented code from app/models/requests/request.rb

christinach:
part of #4178


sandbergja (approved):
Thanks, @christinach ! 🐬

PR #4218, opened 2024-09-04
[#4175] Limit the mutability of Requestable's services array

sandbergja:
Prior to this commit, the `@services` variable of the `Requestable` class had an `attr_accessor`, so other classes could make any changes they wanted to the array.

This commit adds an explicit writer method, so that other classes don't need to be so familiar with the internal workings of `Requestable`, and can only mutate `@services` in one way (totally replacing it with a new array).

Resolves one reek warning, and helps with #4175


christinach (approved):
thanks @sandbergja

PR #4223, opened 2024-09-04
[#4175] Extract Item#location method

sandbergja:
This frees the Requestable class of the responsibility of checking the state of the location key in the Item hash; Item can just provide this information directly to Requestable.

Helps with #4175


christinach (approved):
thanks @sandbergja

PR #4226, opened 2024-09-05
[#4175] Start gradually replacing location hash with a Location object in Requestable and ApplicationHelper

sandbergja:
Reduces reek warnings by 4

Helps with #4175


christinach (approved):
Thanks @sandbergja !

PR #4232, opened 2024-09-05
Start using the Location#annex?, #aeon?, and #code methods more

sandbergja:
This continues to gradually replace use of the location hash with the new Location class (started in #4226).


christinach (approved):
Thanks @sandbergja !

PR #4236, opened 2024-09-05
[#3872] Remove Pageable class

### LINKED ISSUE #3872: Remove the Pageable Class from Requests
We no longer use the Pageable class in the requests module for any current business logic. 

### Level of urgency
- [ ] High
- [ ] Moderate
- [ ] Low

- This is unused code. 

### Acceptance criteria
 - [ ] Pageable class is removed
 - [ ] Test suite removes associated tests for this class. 
 - [ ] Current test suite continues to pass. 

This is the [only place](https://github.com/pulibrary/orangelight/blob/e2bd076f6617b7bff87e55d5fd32e0f1b2051e4a/app/helpers/requests/application_helper.rb#L254) a method from this class is used.

sandbergja:
closes #3872

### DIFF
--- a/app/helpers/requests/application_helper.rb
+++ b/app/helpers/requests/application_helper.rb
@@ -237,18 +237,11 @@ module Requests
       return multi_item unless requestable_list.size == 1
       if requestable_list.first.services.empty?
         no_item
-      elsif requestable_list.first.charged?
-        return multi_item if requestable_list.first.annex?
-        single_item # no_item
-      else
-        submit_message_for_requestable_items(requestable_list)
-      end
-    end
-
-    def submit_message_for_requestable_items(requestable_list)
-      single_item = "Request this Item"
-      multi_item = "Request Selected Items"
-      if requestable_list.first.annex?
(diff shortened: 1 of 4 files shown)


app/helpers/requests/application_helper.rb
```diff
@@ -248,7 +248,7 @@ def submit_message(requestable_list)
     def submit_message_for_requestable_items(requestable_list)
       single_item = "Request this Item"
       multi_item = "Request Selected Items"
-      if requestable_list.first.annex? || requestable_list.first.pageable_loc?
+      if requestable_list.first.annex?
```
kevinreiss, line 251: Does this mean on items in the annex will ever see the "multi_item" submit message on line 250?

sandbergja: Ah, good catch!  Should it be more like:
```suggestion
      if requestable_list.first.annex? || !requestable_list.first.charged?
```

kevinreiss (approved):

PR #4240, opened 2024-09-06
Remove comments that don't provide significant information

christinach:


sandbergja (approved):
Thanks, @christinach 🐬

PR #4243, opened 2024-09-06
Remove Requests::Aeon from the ApplicationHelper

sandbergja:
It seems to no longer be needed.


christinach (approved):
@sandbergja Thanks

PR #4260, opened 2024-09-06
[#4198] Refactor out the EmptyRequestDecorator

### LINKED ISSUE #4198: Investigate if we need the EmptyRequestDecorator
If it's not needed remove it. 
If it's needed add documentation explaining how it is used.

sandbergja:
This decorator was previously used to provide an error page when a user attempted to load the request form for an Alma item but did not include a mfhd parameter in their URL.

This commit uses a new .erb template to provide that error message instead, which I personally find a little easier to understand.

closes #4198


christinach (approved):
🍄

PR #4263, opened 2024-09-06
Use Location object instead of location hash in RequestableDecorator

sandbergja:
Continuing the gradual refactor started in #4226


christinach (approved):
thanks @sandbergja

PR #4264, opened 2024-09-07
Add recap? method in Location class.

christinach:
Add recap? method in Location class.
Use it in requestable object


sandbergja (approved):
Thanks, @christinach 🐬

PR #4266, opened 2024-09-09
[#4200] Remove map_url method and Mapable class

### LINKED ISSUE #4200: Remove method map_url
Remove method `def map_url`

sandbergja:
They are no longer used

Closes #4200


christinach (approved):
Thanks @sandbergja

PR #4268, opened 2024-09-09
[#4178] Remove thesis and numismatics. both of them go through AEON.

christinach:


sandbergja (approved):
Thanks, @christinach @sandbergja

app/controllers/requests/request_controller.rb
```diff
@@ -38,16 +37,6 @@ def aeon?
       false
     end
 
-    def redirect_single_aeon_thesis_numismatics
```
kevinreiss, line 41: After we remove this, do we still see the behavior where numismatics and thesis holdings redirect you directly to Aeon?

spec/models/requests/aeon_url_spec.rb
```diff
@@ -67,14 +67,6 @@
       expect(subject).to include('Site=MARQ')
     end
   end
-  context 'when a thesis holding' do
```
kevinreiss, line 70: I think we should retain these specs, we should make sure that thesis holdings are routed to MUDD.

app/controllers/requests/request_controller.rb
christinach: Yes they go through AEON

sandbergja: @kevinreiss do you mean if a user goes directly to https://catalog.princeton.edu/requests/coin-876?aeon=false ? If so, it doesn't seem to be redirecting in production currently.

sandbergja: whoops, scratch that.  https://catalog.princeton.edu/requests/coin-876?aeon=false&mfhd=numismatics does still redirect in prod, but not in our branch.  Is this redirect behavior needed?

sandbergja: @christinach and I looked at the http://localhost:3000/requests/coin-876?aeon=false&mfhd=numismatics together, and we don't think it presents a problem.  The show page does not link to that URL, and a user probably couldn't have bookmarked it, since it's a redirect.

PR #4271, opened 2024-09-09
Show aeon option for SCSB supervised use

sandbergja:
This fixes a bug that @christinach and I found:

In https://catalog.princeton.edu/catalog/SCSB-5777026/raw.json, supervised use is capitalized as "Supervised use", but in our code, we were only looking for "Supervised Use"


christinach (approved):
Thanks @sandbergja @christinach 🍄 🐬

PR #4275, opened 2024-09-09
Remove online? method and related code

### LINKED ISSUE #4246: Remove the online? method and requestable_online_items partial

Remove the `online?` method and `requestable_online_items` partial from requests

### Level of urgency
- [ ] High
- [x] Moderate
- [ ] Low

- These were used in voyager, which had separate `online` locations for electronic resources.  We no longer have these locations in Alma.  Even if we did, the requests form is limited to one MFHD at a time, so users would not see a mix of physical and online holdings.

### Acceptance criteria
 - [ ] We remove the `online?` method
 - [ ] We remove the `requestable_online_items` partial

rladdusaw:
Resolves #4246


sandbergja (approved):
Thanks, @rladdusaw 🐈‍⬛

PR #4276, opened 2024-09-09
[#4270] Remove the unused openurl_ctx_kev method from requests

### LINKED ISSUE #4270: Investigate openurl_ctx_kev
Determine whether we can remove `openurl_ctx_kev` method (included in debride report) - if it is safe to remove, do so, if not, add to `.debride_allowlist.txt`

If this can be removed, determine whether SolrOpenUrlContext ctx method can also be removed.
The method is used a lot in the view layer.

sandbergja:
Note that other parts of orangelight continue to use a similarly named method: export_as_openurl_ctx_kev

closes #4270

### DIFF
--- a/.debride_allowlist.txt
+++ b/.debride_allowlist.txt
@@ -54,4 +54,3 @@ sent
 # additional_validation!
 # recap_partner_location_codes
 # pageable_loc?
-# openurl_ctx_kev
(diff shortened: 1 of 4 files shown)


christinach (approved):
Thanks @sandbergja

PR #4277, opened 2024-09-09
[#4258] Rename clancy? method to clancy_available?

sandbergja:
This makes it clearer how it is different from the `item_at_clancy?` method

Helps with #4258


christinach (approved):
thanks 🍄

PR #4284, opened 2024-09-10
Remove user_barcode method from Requestable

sandbergja:
It is not used.  Note that `Requests::Submission` also has a `user_barcode` method; that one is actually used.


christinach (approved):
Thanks @sandbergja

PR #4285, opened 2024-09-10
Refactor two methods out of Requests::Router

sandbergja:
These methods both looked through an array of eligibility objects and selected the eligible? ones. This commit combines the two arrays.


christinach (approved):
Thanks @sandbergja !

PR #4286, opened 2024-09-10
[#4247] Refactor annex eligibility into its own class

### LINKED ISSUE #4247: Extract the Annex eligibility checks from the requests Router into its own class

Extract the Annex eligibility checks from the requests Router into its own class.

* The Router class will no longer need to know all the details about what makes a requestable and user eligible to request the Annex service.
* The eligibility logic for the Annex service will be easier to understand, test, and modify in a single location

### Acceptance criteria
 - [ ] There is a new class in the `ServiceEligibility` module that is responsible for determining if a user and a requestable are eligible for the Annex service
 - [ ] The new class has a similar API to other classes in the `ServiceEligibility` module 
 - [ ] The Router uses the new class to determine eligibility for the Annex service

See PR #4238 for examples

sandbergja:
Also, make sure that OnShelfDigitize still returns true for annex items.

closes #4247


maxkadel (approved):
Thanks @sandbergja! 🍄

PR #4288, opened 2024-09-10
Extract in_process eligibility from router

### LINKED ISSUE #4249: Extract the InProcess eligibility checks from the requests Router into its own class

Extract the InProcess eligibility checks from the requests Router into its own class.

* The Router class will no longer need to know all the details about what makes a requestable and user eligible to request the InProcess service.
* The eligibility logic for the InProcess service will be easier to understand, test, and modify in a single location

### Acceptance criteria
 - [ ] There is a new class in the `ServiceEligibility` module that is responsible for determining if a user and a requestable are eligible for the InProcess service
 - [ ] The Router uses the new class to determine eligibility for the InProcess service

See PR #4238 for examples

rladdusaw:
Resolves #4249

### DIFF
--- a/app/models/requests/router.rb
+++ b/app/models/requests/router.rb
@@ -52,10 +52,11 @@ module Requests
       # rubocop:disable Metrics/MethodLength
       def calculate_alma_or_scsb_services
         return [] unless auth_user?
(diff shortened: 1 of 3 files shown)


sandbergja (approved):
Thanks, @rladdusaw 🐈‍⬛

PR #4289, opened 2024-09-10
Extract OnOrder eligibility checks

### LINKED ISSUE #4248: Extract the OnOrder eligibility checks from the requests Router into its own class

Extract the OnOrder eligibility checks from the requests Router into its own class.

* The Router class will no longer need to know all the details about what makes a requestable and user eligible to request the OnOrder service.
* The eligibility logic for the OnOrder service will be easier to understand, test, and modify in a single location

### Acceptance criteria
 - [ ] There is a new class in the `ServiceEligibility` module that is responsible for determining if a user and a requestable are eligible for the OnOrder service
 - [ ] The new class has a similar API to other classes in the ServiceEligibility module
 - [ ] The Router uses the new class to determine eligibility for the OnOrder service

See PR #4238 for examples

maxkadel:
Closes #4248

### DIFF
--- a/app/models/requests/service_eligibility/on_order.rb
+++ b/app/models/requests/service_eligibility/on_order.rb
@@ -19,7 +19,7 @@ module Requests
       private
 
         def requestable_eligible?
-          requestable.on_order?
+          requestable.on_order? && !requestable.charged? && !requestable.in_process?
         end
 
         def user_eligible?
--- a/app/models/requests/service_eligibility/on_order.rb
+++ b/app/models/requests/service_eligibility/on_order.rb
@@ -1,0 +1,32 @@
+# frozen_string_literal: true
+module Requests
+  module ServiceEligibility
+    # on_order - material has a status in Alma that indicates it is ordered but has not yet arrived on campus
+    class OnOrder
+      def initialize(requestable:, user:)
+        @requestable = requestable
+        @user = user
+      end
+
+      def to_s
+        'on_order'
+      end
+
+      def eligible?
+        requestable_eligible? && user_eligible?
+      end
+
+      private
+
+        def requestable_eligible?
+          requestable.on_order?
+        end
+
+        def user_eligible?
+          user.cas_provider? || user.alma_provider?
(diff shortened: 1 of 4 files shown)


spec/models/requests/service_eligibility/on_order_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+  describe '#eligible?' do
+    it 'returns true if all criteria are met' do
+      requestable = instance_double(Requests::Requestable)
+      allow(requestable).to receive_messages(on_order?: true)
+
+      eligibility = described_class.new(requestable:, user:)
+
+      expect(eligibility.eligible?).to be(true)
+    end
+    it 'returns false if the item is at clancy' do
```
sandbergja, line 15:
```suggestion
    it 'returns false if the item is not on order' do
```

app/models/requests/service_eligibility/on_order.rb
```diff
@@ -0,0 +1,32 @@
...
+      end
+
+      def eligible?
+        requestable_eligible? && user_eligible?
+      end
+
+      private
+
+        def requestable_eligible?
+          requestable.on_order?
```
sandbergja, line 22: Do we need any of the other conditionals (e.g. `(requestable.alma_managed? || requestable.partner_holding?) && !requestable.aeon?`, `!requestable.charged?`, `!requestable.in_process?`), or are they all redundant?

sandbergja (commented):
Thanks, @maxkadel 🦄 !  Just two questions.

app/models/requests/service_eligibility/on_order.rb
maxkadel: I think they might be redundant, but just to be on the safe side I included them, since that will re-create the existing branching logic. I don't think it should be possible for on_order? to be true and those other conditions not be met, but I'm not positive, so might as well include it.

sandbergja (approved):
Thanks, @maxkadel 🦄

PR #4292, opened 2024-09-10
Remove request source - no longer used

### LINKED ISSUE #4173: Remove the source parameter and instance variable from Requests::Request

Remove the `source` initializer parameter and `@source` instance variable from `Requests::Request`

### Level of urgency
- [ ] High
- [x] Moderate
- [ ] Low

- `source` used to keep track of which system had sent the request.  At this point, only the catalog sends requests, so it is no longer needed.
- This will remove a little bit of complexity from requests

### Acceptance criteria
 - [ ] `Requests::Request` does not accept a `source` initializer parameter
 - [ ] `Requests::Request` does not have a `@source` instance variable

maxkadel:
Closes #4173

### DIFF
--- a/app/controllers/requests/request_controller.rb
+++ b/app/controllers/requests/request_controller.rb
@@ -13,7 +13,6 @@ module Requests
 
     def generate
       system_id = sanitize(params[:system_id])
-      source = sanitize(params[:source]) if params[:source].present?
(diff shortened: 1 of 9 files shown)


christinach (approved):
Thanks @maxkadel

PR #4304, opened 2024-09-11
Incrementally start using the Requests::Holding object in the physical holdings markup builder

sandbergja:


christinach (approved):
Thanks @sandbergja

PR #4311, opened 2024-09-11
Remove requests/application.html.

christinach:
Remove requests/application.html.
It is not used


sandbergja (approved):
Woo hoo!  Thanks, @christinach 🐬

PR #4316, opened 2024-09-12
Clarify between patron object and patron_hash

maxkadel:
- Update method names and add comments
  - Generally do not favor names that include verbs like "build" but name the thing that has been built


sandbergja (approved):
Thanks, @maxkadel 🦄

PR #4318, opened 2024-09-12
Re-create rubocop_todo - almost everything has improved!

maxkadel:


sandbergja (approved):
Exciting!  Thanks, @maxkadel 🦄

PR #4319, opened 2024-09-12
[#4313] Create a new import file for lux

### LINKED ISSUE #4313: Loading vite in requests causes error in dev console- Uncaught ReferenceError: Blacklight is not defined
### Expected behavior
We should not see any errors in the dev console.

### Actual behavior
After https://github.com/pulibrary/orangelight/pull/4310/files we're seeing the following error in the dev console when loading the requests form
```
application.js:21 Uncaught ReferenceError: Blacklight is not defined
```
which is caused in application.js entrypoint  because of:
```
Blacklight.onLoad(() => {
  const loader = new OrangelightUiLoader()
  loader.run()
})
```

christinach:
closes #4313 

Create a requests entrypoint
Import the new luxImport file in requests entrypoint and orangelight entrypoint

### DIFF
--- a/app/javascript/entrypoints/application.js
+++ b/app/javascript/entrypoints/application.js
@@ -1,24 +1,10 @@
-import OrangelightUiLoader from '../orangelight/orangelight_ui_loader.es6'
-import { createApp } from 'vue'
-import 'lux-design-system/dist/style.css'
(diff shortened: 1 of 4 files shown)


sandbergja (approved):
Thanks, @christinach 🐬

PR #4320, opened 2024-09-12
Pull out FullPatron class from Patron

maxkadel:


app/models/requests/full_patron.rb
```diff
@@ -0,0 +1,57 @@
...
+  class FullPatron
+    attr_reader :hash, :errors, :api_response, :uid
+    def initialize(uid:)
+      @errors = []
+      @uid = uid
+      @hash = patron_hash
+    end
+
+    private
+
```
christinach, line 13: @maxkadel we're reviewing the PR with @sandbergja .
Can we move the `attr_reader :api_response, :uid` in the private?

christinach (commented):
@maxkadel one question . It is not a blocker.

christinach (approved):
Thanks @maxkadel !

Issue #4303, opened 2024-09-11
Remove long-commented out code from Requests


maxkadel:
### What maintenance needs to be done?
Remove code that has been commented out for a long time from requests.

### Level of urgency
- [ ] High
- [x] Moderate
- [ ] Low

### Why is this maintenance needed?
- This code adds cognitive overhead every time we work on a section of code that has it, without (usually) adding value to the code. Often it is for long-deprecated features. 

### Acceptance criteria
 - [ ] Commented code is removed, unless it adds substantial value to still-used code
 - [ ] Tests still pass
 - [ ] Stretch goal - commented out or pending tests related to this code are also removed

PR #4327, opened 2024-09-12
[#4303] Remove commented code and comments that are not useful anymore

christinach:
helps with #4303


app/models/requests/submissions/generic.rb
```diff
@@ -4,9 +4,5 @@ class Generic < Service
     def initialize(submission, service_type: 'generic')
       super
     end
-
-    def handle
-      # something will go here eventually
```
christinach, line 9: 😆

sandbergja (approved):
Thanks, @christinach 🐬

Issue #4130, opened 2024-08-22
Use libanswers API, rather than email, for AskAQuestion, ReportHarmfulLanguage, and SuggestCorrection forms


sandbergja:
### What maintenance needs to be done?

Use libanswers API, rather than email, for the following forms (linked from the record show page):

- [x] AskAQuestion
- [x] ReportHarmfulLanguage
- [x] SuggestCorrection

### Level of urgency
- [x] High - we should prioritize this in the current work cycle, underneath the Blacklight 8 and Rails upgrades, but above other items
- [ ] Moderate
- [ ] Low

### Why is this maintenance needed?

Users with non-Princeton email addresses can currently not ask questions, report harmful language, or suggest corrections!

### Acceptance criteria
- [x] AskAQuestion form can be submitted with a non-Princeton email address and its responses appear in the correct libanswers queue
- [x] ReportHarmfulLanguage form can be submitted with a non-Princeton email address and its responses appear in the correct libanswers queue
- [x] SuggestCorrection form can be submitted with a non-Princeton email address and its responses appear in the correct libanswers queue

### Implementation notes, if any

See PR #4073, which switched the Feedback Form to use the libanswers API

sandbergja:
All three forms are now using the libanswers API in prod, so I am closing this ticket.

PR #4329, opened 2024-09-12
[#4130] Use Libanswers API for the ReportHarmfulLanguage form

sandbergja:
helps with #4130


christinach (approved):
Thank you @sandbergja !

PR #4331, opened 2024-09-13
Use libanswers API for Suggest a Correction form

sandbergja:
helps with #4130


christinach (approved):
Thank you @sandbergja !

PR #4333, opened 2024-09-13
[#4130] Remove unneeded code and configuration, now that all forms submit via Libanswers

sandbergja:
Helps with #4130


christinach (approved):
Thnaks @sandbergja 🍄

PR #4337, opened 2024-09-13
Remove reek warning for get_current_location

christinach:
reduce reek in route_requests


sandbergja (approved):
Thanks, @christinach 🐬

PR #4338, opened 2024-09-13
Remove not useful comments from form.rb

christinach:


sandbergja (approved):
Thanks, @christinach 🐬

PR #4342, opened 2024-09-16
Refactor FullPatron to use existing Bibdata Service

maxkadel:
Dries up code significantly

The only functional change is that it does not display a flash message to a user specifying the Alma API as the issue with their login when that occurs - I cannot think why this level of specificity would be helpful to the user, but I could be missing something!


christinach (approved):
Thanks @maxkadel !

PR #4343, opened 2024-09-16
Clean specs from campus_authorized and campus_authorized_category

christinach:
part of https://github.com/pulibrary/bibdata/issues/1972 #4343

Clean specs from campus_authorized and campus_authorized_category


maxkadel (approved):
Thanks @christinach! 🐬

PR #4346, opened 2024-09-16
Update CD config to narrow scope even more

acozine:
This PR removes all vars from the API call except the repo name, everything else is hardcoded in Tower.


sandbergja (approved):
Thanks, @acozine 🐴

christinach (approved):
Thanks @acozine

PR #4349, opened 2024-09-16
Dry up Illiad patron code some

maxkadel:
This could use some further work, but wanted to get to a stopping point that is working


sandbergja (approved):
Thanks, @maxkadel 🦄

PR #4350, opened 2024-09-16
Refactor AeonURL

sandbergja:
In preparation for work on pulibrary/bibdata#1827

- Stop nesting `Item` under `Requests::Requestable`, since it didn't autoload properly for use in `AeonUrl` while it was there.
- Consistently use the `Requests::Item` class within `AeonUrl`.  Before this PR, `item` could be either a `Hash` or a `Requests::Item`
- Use the existing `enum_value` and `barcode` methods from `Requests::Item`, rather than reimplementing that logic in `AeonUrl`

### DIFF
--- a/.reek.yml
+++ b/.reek.yml
@@ -191,7 +191,6 @@ detectors:
     - Requests::Form#off_site?
     - Requests::Form#serial?
     - Requests::Form#status_label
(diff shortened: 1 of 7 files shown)


christinach (approved):
Thanks @sandbergja 🍄

PR #4351, opened 2024-09-16
Requests::Form more refactoring

christinach:
Remove Form#off_site? method
Use item_object in Form#status_label
Add not_a_work_order? method in the Item class


sandbergja (approved):
Thanks, @christinach 🐬

PR #4355, opened 2024-09-17
Test alma affiliate users for in process and on order items

maxkadel:
Test to verify that users authenticated through Alma can request On Order and In Process materials


christinach (approved):
Thank you @maxkadel ! 🦄

PR #4366, opened 2024-09-18
Remove unused aeon_openurl method

sandbergja:
Aeon OpenURLs are now created using the AeonUrl class or `RequestableDecorator::aeon_url`


christinach (approved):
Thanks @sandbergja

PR #4367, opened 2024-09-18
Allow Item class to read enumeration+chronology info from either the description or enumeration field

sandbergja:
This is preparation for pulibrary/bibdata#1827, so that we can deploy that at any time and  maintain backwards compatibility.


maxkadel (approved):
Thanks @sandbergja! 🍄 🍄 🍄 (lots of mushrooms, given your last harvest!)

PR #4371, opened 2024-09-19
Add dastardly to CI

rladdusaw:


rladdusaw:
Decided not to adopt.

PR #4386, opened 2024-09-23
[#4372] Retrieving a temp location for the second time in Requests::Form should not return nil

### LINKED ISSUE #4372: Request form is failing with  NoMethodError in Requestable::recap_edd?
### Expected behavior

https://catalog.princeton.edu/requests/9938694433506421?aeon=false&mfhd=22575739710006421 works

### Actual behavior

It gives "Something went wrong".

### Impact of this bug
Users can't request certain materials

[honeybadger link](https://app.honeybadger.io/projects/54399/faults/112117828)

```
    def recap_edd?
      return location[:recap_electronic_delivery_location] == true unless partner_holding?
      in_scsb_edd_collection? && !scsb_in_library_use?
    end
```

```
NoMethodError: undefined method `[]' for nil:NilClass
```

### Notes

From @kevinreiss

> potentially this could be a data issue as well as their is some weirdness I notice in the record display.  This item is a standard recap$pa item and there doesn't appear to be a general problem for items in that location. I noticed this one has an item charged to the resource sharing location displayed on the holdings statement (which I actually think maybe should not be happy).

sandbergja:
Prior to this commit, `Requests::Form#get_current_location` would return the location data the first time it was called for a particular temporary location, and then cache the hash in memory.  However, the second time `#get_current_location` was called for a particular temp location, rather than returning the cached value, it would return `nil`, which causes errors later in the process.

This scenario happened when multiple items on a record are in the same temp location.

This commit extracts the caching logic into its own tiny class, and ensures that it returns the location hash whether or not it had been cached previously.

Closes #4372

### DIFF
--- a/app/models/requests/form.rb
+++ b/app/models/requests/form.rb
@@ -258,8 +258,8 @@ module Requests
 
       def get_current_location(item_location_code:)
         if item_location_code != location_code
-          @temp_locations ||= {}
-          @temp_locations[item_location_code] = get_location_data(item_location_code) if @temp_locations[item_location_code].blank?
+          @temp_locations ||= TempLocationCache.new
+          @temp_locations.retrieve(item_location_code)
         else
           location
         end
--- a/app/models/requests/temp_location_cache.rb
+++ b/app/models/requests/temp_location_cache.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+module Requests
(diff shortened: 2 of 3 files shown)


kevinreiss (approved):

christinach (approved):
Thanks @sandbergja

PR #4387, opened 2024-09-23
Add a new content-security-policy-report-only header

sandbergja:
Report any violations to the console and also to honeybadger, but don't actually block any violations.

The eventual goal will be to replace our current content-security-policy header (currently configured in princeton_ansible) with the contents of the content-security-policy-report-only introduced in this commit.  We can do that when we don't see many violations in Honeybadger.


christinach (approved):
Thanks @sandbergja ! 🍄

PR #4388, opened 2024-09-23
Address recent CSP reports from deploying to prod

sandbergja:
These were reported to Honeybadger


christinach (approved):
Thanks @sandbergja

PR #4398, opened 2024-09-26
Lint staged .es6 and vue files

christinach:


kevinreiss (approved):

regineheberlein (commented):
thanks @christinach !

PR #4401, opened 2024-09-30
CSP: allow any font source, as long as it is self, https, or data

sandbergja:
This allows MathJAX to fetch a font from cloudflare, and allows browser extensions to fetch their own fonts if needed


christinach (approved):
Thanks @sandbergja

Issue #4352, opened 2024-09-16
Evaluate and update our Content Security Policy


sandbergja:
### What maintenance needs to be done?

Evaluate our current content security policy, which [is provided at the nginx level](https://github.com/pulibrary/princeton_ansible/blob/2696c2bb32e71f86531e286ff798a78e86ca6f4d/roles/nginxplus/files/conf/http/catalog-prod.conf#L74-L75).  See what changes we need to make to keep it up-to-date and  provide our users with protections against XSS and clickjacking.

### Why is this maintenance needed?

To protect users from XSS and clickjacking.

### Acceptance criteria
 - [ ] Determine if the current CSP rule is still valid
 - [ ] Determine if we want to serve our CSP at the application (e.g. with the Rails CSP helper) or load balancer level
 - [ ] See if this PR offers a valid approach: https://github.com/pulibrary/orangelight/pull/3807
 - [ ] Turn on a [CSP in report-only mode](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP#testing_your_policy) and start refining it.
 - [ ] When it is good to go, turn on the CSP for real.

christinach:
We will move the stricter CSP out of the report only mode to blocking mode.

PR #4403, opened 2024-09-30
[CSP] Set object-src to none to remove vulnerability warning: "object-src" value is not safe

christinach:
related to #4352 

See MDN documentation: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/object-src


sandbergja (approved):
🍄 🐬

PR #4408, opened 2024-10-02
Remove unnecessary partial

sandbergja:
This partial is wrapped in an if statement with a condition that always returns `false`. It would return `true` if we had configured a custom `thumbnail_presenter`, but we don't have one of those, so it never displays anything.


christinach (approved):
Thanks @sandbergja ! 🍄

Issue #3570, opened 2023-05-19
Restrict automatic Google thumbnail import for Special Collections print materials records 


kelea99:
### User story
_As a researcher looking at online catalog records for print materials held in Special Collections, I may be confused by the thumbnail image automatically populated in the upper left corner of the record from Google. The thumbnail may be misrepresenting the year, edition, or item that is described in the record. It may influence whether I use, request for digitization, or discard the item in my research or may influence my decision to come to the library to view the item in person._ 

### An idea
What if we used the holding location codes for items as our map to define what "automatically populated by google" thumbnails should be enabled/disabled? This would be far easier than the idea of manually researching/selecting, ingesting, and adding thumbnails locally that represent the items in the catalog. If an item is important enough to have a thumbnail as a visual, perhaps it is important enough to be digitized in its entirety (then it would be automatically populated thanks to figgy ingest/completion). 

### Acceptance criteria

- [x] All "Aeon" locations in Bibdata data don't request a Google Book Cover Image.
- [x] Show the default cover image instead for these locations. 
- [x] Let SC Public Service staff know we are making this change. (Sara L. and Christa). 
- [x] I can successfully use this feature with Voice Over in Safari.
- [x] I have checked this feature using one of the following browser extensions and confirmed that it does not add any level A or level AA issues:
    * axe dev tools
    * Lighthouse
    * Wave

### Concrete example
Eric White received a question from a researcher about the editions of Candide present in Special Collections, who was led astray by the thumbnails of the title pages present in the catalog. See [slack comment](https://pulibrary.slack.com/archives/CA0UY0Q07/p1671710840477119) and [a title search for early editions of Candide](https://catalog.princeton.edu/?f%5Baccess_facet%5D%5B%5D=In+the+Library&q=Candide&range%5Bpub_date_start_sort%5D%5Bbegin%5D=1759&range%5Bpub_date_start_sort%5D%5Bend%5D=1853&search_field=left_anchor). 

A specific example of different editions that have the same image are:
* https://catalog.princeton.edu/catalog/9924628093506421 ([Paris, Lambert] 1759.)
* https://catalog.princeton.edu/catalog/9924628083506421 ([Genève, Cramer] 1759.)

kelea99:
[...]

PR #4411, opened 2024-10-02
[#3570] Don't display Google Books thumbnails for materials in aeon locations

sandbergja:
Helps with #3570

### DIFF
--- a/app/models/solr_document.rb
+++ b/app/models/solr_document.rb
@@ -174,6 +174,18 @@ class SolrDocument
     solr_document_id&.start_with? 'coin'
   end
 
+  def in_a_special_collection?
+    holdings_1display = self['holdings_1display']
+    return true unless holdings_1display
+
+    JSON.parse(holdings_1display)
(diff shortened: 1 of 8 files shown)


app/models/solr_document.rb
```diff
@@ -174,6 +174,18 @@ def numismatics_record?
...
 
+  def in_a_special_collection?
+    holdings_1display = self['holdings_1display']
+    return true unless holdings_1display
+
+    JSON.parse(holdings_1display)
+        &.values
+        &.any? do |holding|
+          location_code = holding['location_code']
+          Bibdata.holding_locations.dig(location_code, 'aeon_location') if location_code
```
kevinreiss, line 185: We could does check at the library level instead if it's desirable just leave out everything in MUDD, SC, Marquand, and EAL

christinach (approved):
Thanks @sandbergja . I'm approving, Ci is failing. Feel free to merge it when it's green.

Issue #3824, opened 2023-11-09
Update Search Results Online Display for Records with Multiple Online Portfolios in Alma. 


kevinreiss:
### Current display
When a record has multiple online links we display the "first" link on the search results page. [Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBLzIvQVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--6d4cf22250c35c2def317057953487c2bda0c51e/Screenshot%202023-11-09%20at%201.48.22%20PM.png), noting that this issue isn't present when it's a MMS ID that has two or more print holding locations. See search https://catalog.princeton.edu/catalog?q=science&search_field=all_fields and MMS ID (2nd result) https://catalog.princeton.edu/catalog/998574693506421. 

### Desired display
When a records has more than a single value in the electronic_portfolio_s array in the solr document instead of displaying hte first link mimic what we do for print records with multiple holdings and include a message next to the blue online badge stating "View Record for All Options": 

### Acceptance criteria
- [x] Create a list of possible wording choices for the single label and review with POs. 
- [x] When a record has multiple online links display in the search results the blue 'Online' badge and text: 'View record for full availability'.  
- [x] If there is ONLY one link display it. 
- [x] If there is more than one don't display a link, but instead display only the badge and the label. 
- [x] Remove the tooltip.
- [x] I can successfully use this feature with Voice Over in Safari.
- [x] I have checked this feature using one of the following browser extensions and confirmed that it does not add any level A or level AA issues:
    * axe dev tools
    * Lighthouse
    * Wave

### Notes
Example MMS ID with 8 porfolios: https://catalog.princeton.edu/catalog/99125253392806421
Current Search Results: https://catalog.princeton.edu/catalog?utf8=%E2%9C%93&search_field=all_fields&q=99125253392806421

kevinreiss:
When we implement this we should also implement it in parallel in the all search frontend.

ellen-aa:
The acceptance criteria is clear and easily corrects the misleading results page.

caroldh:
In a related comment, can we standardize the "view record for full availability" displays in blacklight. Currently, there is a difference for SCSB and PUL items: 
[...]

PR #4415, opened 2024-10-03
Upgrade lux and related npm packages

sandbergja:
This helps with #3824, and also gets us @ishasinha1 's fix for https://github.com/pulibrary/lux-design-system/issues/340


christinach (approved):
Thanks @sandbergja

Issue #2965, opened 2022-04-22
Move viewer location up the page on item level catalog records.


kelea99:
### Summary
_As a library patron searching the online catalog, I would like to see Princeton's online holdings and viewer above the physical holdings in the item level catalog record._ 

- Issues: We have been given feedback from many patrons and library staff about discovery of our digital holdings in the online catalog.  One of two particular painpoints is the location of the viewer at the bottom of the record.  Because the viewer is out of site unless a patron knows to scroll down they may (and have) missed the viewer and access to the digital surrogate entirely.  In conversations with users about the thumbnail, many do not know to click on the thumbnail in order to be taken down the page to the viewer. Until 856 fields/ELFs are created in marc for digital holdings, online content links and faceting also will not appear to help users.  
- Ideas: Move the viewer up the page to just above the physical holdings information in some way.  Users have commented that if the item they are looking for has a digital surrogate, that is what they will want to see first and foremoest. @escowles had a great suggestion, which was to investigate utilizing that 3rd column on the page differently, and to contain the viewer further up the page.  

### Impact 
_Please include hard deadlines, if the exhibit is part of an event, the issue is stopping work, etc._
Work is not halted, but user experience and discovery are directly impacted. 

### Priority recommendation

- [ ] asap
- [ ] within the next 3 weeks
- [ ] PO will prioritize

### Examples from other institutions (viewer placement)

- [Stanford University Libraries](https://searchworks.stanford.edu/view/wx275ct4832)

sandbergja:
One thing we like about the Stanford view is that it doesn't take 100% of the width of the page.  This means that if a user's mouse is one the side of the viewer, they can still scroll down the page.

escowles:
@sandbergja I think that's a good point — it's frustrating when you're trying to scroll and you wind up zooming in the viewer instead. I think putting the viewer in the Available Online section, and making it as wide as the column, would do that.

The other thought I had, which would be a bigger change, would be making the left column wider and putting the viewer where we put the thumbnail now. We would probably want a button to make the viewer larger, like Youtube and lots of other video sites have these days.

kelea99:
@sandbergja That was an actual painpoint given to us by folks in Mendel Music library, as well as one other user.  They noted frustration trying to scroll back up to the top of the page using their mouse or trackpad, only to be stuck zooming in and out of the viewer.  That is a great point.

kevinreiss:
[...]

PR #4417, opened 2024-10-03
[#2965] give some padding to the viewer so that the user can scroll easier

christinach:
part of #2965

give some padding to the viewer so that the user can scroll easier up and down on the page
Update viewer backround to be #ffff;

Example1 on catalog-staging: https://catalog-staging.princeton.edu/catalog/9934901833506421
Example2 on catalog-staging: https://catalog-staging.princeton.edu/catalog/9935115343506421


sandbergja (approved):
Thanks, @christinach 🐬

PR #4419, opened 2024-10-04
Only run coveralls in CI, to speed up local test runs

sandbergja:
On a sample of tests run locally, this reduced the median time running the tests by 3%.


christinach (approved):
Thanks @sandbergja

PR #4420, opened 2024-10-04
Remove unused uv__overlay class

christinach:


sandbergja (approved):
Thanks, @christinach 🐬

PR #4422, opened 2024-10-04
Only load datadog in production

sandbergja:
This speeds up tests significantly, a random sample of 100 tests takes 19 seconds less without loading datadog.

Identified using [test-prof](https://github.com/test-prof/test-prof) and [speedscope](https://www.speedscope.app/)


christinach (approved):
Thanks @sandbergja

PR #4424, opened 2024-10-04
Add microform icon

sandbergja:
Closes https://github.com/pulibrary/bibdata/issues/2498

Here's how the facets look with this icon:
an icon showing a strip of film with perforations on the side, next to the term Microform and a count of 18873 matching resources


christinach (approved):
Thanks @sandbergja !

Issue #4012, opened 2024-05-06
Add citation for SCSB Records


christinach:
### User story
As a researcher who will be citing SCSB records  I want to easily copy and paste a citation into my document

### Acceptance criteria
- [x] Display 'Cite' button
- [x] Update the code to build the citations for  SCSB records
- [x] I can successfully use this feature with Voice Over in Safari.
- [ ] I have checked this feature using one of the following browser extensions and confirmed that it does not add any level A or level AA issues:
    * axe dev tools
    * Lighthouse
    * Wave
- [ ] ...

### Concrete example

See the title at[ this link](https://catalog.princeton.edu/catalog/SCSB-5789141) from Columbia's holdings (SCSB) as an example. You cannot cite, nor export to a bibliographic management system (e.g. Zotero). 

### Implementation notes, if any
Related to https://github.com/pulibrary/orangelight/issues/4124
- Investigate using [citeproc-ruby](https://github.com/inukshuk/citeproc-ruby)

christinach:
related to https://github.com/pulibrary/orangelight/issues/4124

maxkadel:
I tried citeproc-ruby and ran into some problems. 
a) I could not figure out why, but when using the MLA format, it did not add the delimiter between the last and first names. This could be a bug or driver error?
b) The publisher is not in the Solr record without the place of publication, which is needed for most MLA book citations.
c) Again, this could be my own error, but I can't get all the fields I need without another call to Solr. At that point in the code, the fields included in the document are very minimal.

maxkadel:
Closed by #4427

PR #4427, opened 2024-10-08
Add citations for SCSB items

maxkadel:
I think this PR delivers value, but there are some things I'd like help improving.

Connected to #4012

- Right now, it is only tested for Books, I think we should add tests for other types of items, especially journals.
- It requires a second call to Solr, which seems like it shouldn't be necessary, but I've spent a lot of time in the catalog controller trying to add the `_citation_` fields so they are in the solr document at this point, but haven't succeeded yet. (See https://github.com/pulibrary/orangelight/issues/4429 for follow-up ticket to resolve this)
- ~Probably a lot of the methods in Blacklight::Document::Mla can be a) Private and b) moved into an abstract method for other CiteProc citations (APA, Chicago)~ - done
- Right now we have to take apart existing Solr fields to get the publication information - see https://github.com/pulibrary/bibdata/issues/2515 for follow-up ticket to add separate Solr fields so we do not have to process them here but can use them directly from Solr.
- ~For some reason I can't get the author name to show up correctly with the comma between the family name and the given name. I've added a pending test for this issue.~ (I'm wondering if this is because of the combining accent in the example? - it does work correctly in development for SCSB-8953469) - I have created a fork that has a test, this seems to only affect names ending in decomposed accents. This is a bug in CiteProc - see https://github.com/pulibrary/citeproc/tree/names_with_final_accents

image


app/components/orangelight/document/citation_component.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class Orangelight::Document::CitationComponent < Blacklight::Document::CitationComponent
+  DEFAULT_FORMATS = {
+    'blacklight.citation.mla': :export_as_mla,
```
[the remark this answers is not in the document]
maxkadel: Ticket to route all types of records through the same routines - https://github.com/pulibrary/orangelight/issues/4428

spec/models/apa_spec.rb
```diff
@@ -0,0 +1,53 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Blacklight::Document::Apa, citation: true do
```
sandbergja, line 5: Woohoo citation tag!

app/models/concerns/blacklight/document/apa.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+# Creates an html APA citation for non-Marc records
+module Blacklight::Document::Apa
```
sandbergja, line 4: I like this structure with the 3 modules for the 3 citation styles

sandbergja (changes requested):
Thanks, @maxkadel !  🦄  I saw one very small thing in the Gemfile, the actual code looks great to me.

sandbergja (approved):
Woohoo!  Thanks, @maxkadel 🦄

PR #4433, opened 2024-10-10
Remove confluence link

sandbergja:
This page is not accessible to the public, and has information that is more useful for staff members, so we don't need this link on our public help page.


christinach (approved):
Thanks @sandbergja

PR #4435, opened 2024-10-11
Display geographic facet

maxkadel:
Connected to https://github.com/pulibrary/bibdata/issues/2034

image

This does not add it to the advanced search page, since the other "Subject: x" searches are not on the advanced search page.


christinach (approved):
Thanks @maxkadel 🦄 

🌐

PR #4446, opened 2024-10-16
[#4081] Update to bootstrap 5

### LINKED ISSUE #4081: Upgrade to Bootstrap 5

Migrate from Bootstrap 4 to Bootstrap 5

### Level of urgency
- [ ] High
- [x] Moderate
- [ ] Low

### EOL upgrade

[Bootstrap 4's end of life was January 1, 2023](https://getbootstrap.com/docs/4.6/end-of-life/)

- We shouldn't rely on EOL software
- [Blacklight is considering stopping support for bootstrap 4](https://github.com/projectblacklight/blacklight/issues/3211)
- Bootstrap 5 removes a dependency on jquery, which could allow us to do the same
- Bootstrap 5 includes accessibility improvements

### Acceptance criteria
 - [ ] We use Bootstrap 5 instead of 4
 - [ ] The site largely functions and looks the same

Current versions of Blacklight support both bootstrap 4 and 5.

[Migrating to v5](https://getbootstrap.com/docs/5.0/migration/)

christinach:
closes #4081


christinach:
The next step for me is to update the dropdown buttons so that the advanced search  component works as expected.

christinach:
The color of the skip links is off but it shouldn't be a blocker.

christinach:
- @kevinreiss noticed that in the record page the show tools should be on the right side and not wrap. (Not blocker)

- cancel button in modal does not work.

Gemfile
```diff
@@ -75,7 +76,7 @@ gem 'rsolr'
 gem 'rspec-rails'
 gem 'rubyzip', '>= 1.2.2'
 # Use SCSS for stylesheets
-gem 'sass-rails', '~> 6.0'
```
sandbergja, line 78: Could we remove this outright, rather than commenting it?

app/assets/stylesheets/blacklight.scss
```diff
@@ -1,11 +1,4 @@
-//bootstrap variables override
-$table-bg-accent: #f5f4f1;
-$font-size-base: 1rem;
-$brand-primary: #3072AB;
-$brand-success: #3B863B;
-$gray-light: #757575;
 
-@import 'bootstrap';
```
sandbergja, line 8: Thanks for organizing these better!

app/components/orangelight/search_bar_component.html.erb
```diff
@@ -1,7 +1,7 @@
 <%= form_with url: search_catalog_url, local: true, method: @method, class: @classes.join(' '), scope: @prefix, role: 'search', aria: { label: scoped_t('submit') }, **@form_options do |f| %>
   <%= render Blacklight::HiddenSearchStateComponent.new(params: @params) %>
   <% if search_fields.length > 1 %>
-    <%= f.label :search_field, scoped_t('search_field.label'), class: 'sr-only visually-hidden' %>
+    <%= f.label :search_field, scoped_t('search_field.label'), class: 'visually-hidden visually-hidden' %>
```
sandbergja, line 4: Looks like the same class is here twice

```suggestion
    <%= f.label :search_field, scoped_t('search_field.label'), class: 'visually-hidden' %>
```

app/views/advanced/_numismatics_advanced_search_form.html.erb
```diff
@@ -0,0 +1,62 @@
+<%= form_tag search_catalog_path, :class => 'advanced form-horizontal', :method => :get do  %>
```
sandbergja, line 1: How is this partial used? Does it replace the existing numismatics search form component?

sandbergja (changes requested):
Thanks for all your work on this, @christinach 🐬 !  It was a big project, and it will be so great to be on Bootstrap 5!

* I think the skip links should be a blocker, I think that's a serious accessibility regression.  I think it might not be the color; in my devtools I see a 
```
position: absolute !important;
clip: rect(1px, 1px, 1px, 1px);
```
when I uncheck them I can see the skip links again.

* I see a very small difference in height in the search bar in the keyword dropdown vs. the search input:
the search input is slightly taller than the keyword dropdown

Gemfile
christinach: yes I will remove it. I will add it in the commit for the Gemfile. I want to keep them tidy and clean.

christinach: Thanks for looking into the skip links. I will check what you pointed out.

app/components/orangelight/search_bar_component.html.erb
christinach: oh yes! probably when I replaced sr-only with visually-hidden.

app/views/advanced/_numismatics_advanced_search_form.html.erb
christinach: I'm not sure I will look into this one.

christinach:
@kevinreiss this branch is rebased against current main.

app/views/advanced/_numismatics_advanced_search_form.html.erb
christinach: @sandbergja the partial was removed while I was away. Rebasing the branch fixed it.

christinach:
> Thanks for all your work on this, @christinach 🐬 ! It was a big project, and it will be so great to be on Bootstrap 5!
> 
> * I think the skip links should be a blocker, I think that's a serious accessibility regression.  I think it might not be the color; in my devtools I see a
> 
> ```
> position: absolute !important;
> clip: rect(1px, 1px, 1px, 1px);
> ```
> 
> when I uncheck them I can see the skip links again.
> 
> * I see a very small difference in height in the search bar in the keyword dropdown vs. the search input:
>   the search input is slightly taller than the keyword dropdown

@sandbergja I rebased the bootstrap 5 branch on main which does not have the blacklight 8 upgrade. We reverted it yesterday. 
I don't see the skip links issue after rebasing.

I also don't see the '...small difference in height in the search bar in the keyword dropdown vs. the search input....'. See the screenshot below:
search bar and dropdown

kevinreiss:
I noticed that Facet values in QA are underlined. 
Screenshot 2024-11-18 at 8 15 40 AM

sandbergja:
@christinach the small difference between the search bar in the keyword dropdown vs. the search input looks like it's only in Firefox (and more subtly in Safari) -- I don't see it in Edge or Chrome.

Another subtle thing: If you go to "Your Account" > "Library Account", the horizontal rule is on top of the text "or".   Do we even need that horizontal rule, I wonder?

the screen gives a choice between logging in with netID or alma account.  Between them is the word or with a line on top of it

christinach:
> @christinach the small difference between the search bar in the keyword dropdown vs. the search input looks like it's only in Firefox (and more subtly in Safari) -- I don't see it in Edge or Chrome.
> 
> Another subtle thing: If you go to "Your Account" > "Library Account", the horizontal rule is on top of the text "or". Do we even need that horizontal rule, I wonder?
> 
> the screen gives a choice between logging in with netID or alma account. Between them is the word or with a line on top of it

@sandbergja thanks. I fixed the height discrepancy. I missed an input-group-prepend which was causing that.
I don't think we need the separator line. I will remove it in another PR against main.

sandbergja (approved):
Thanks, @christinach 🐬 !  This is so great!

Issue #4453, opened 2024-10-18
[Record Page] Update position and display of button 'Where to find it'


christinach:
### Current display
Where to find it has white background color and is positioned below the Library location

### Desired display
Where to find follows the same color scheme as the Request button (background-color: #3072AB; border-color: #3072AB; color: #ffffff;).
Where to find it is positioned in the same line and on the right side of the request button.
 
### Acceptance criteria
- [ ] Rename the button caption 'Where to find it' to 'Find on shelf'.
- [x] 'Find on shelf' follows the same color scheme as the 'Request' button (background-color: #3072AB; border-color: #3072AB; color: #ffffff;).
- [ ] 'Find on shelf'  is positioned in the same line and on the right side of the 'Request' button.
- [ ] I can successfully use this feature with Voice Over in Safari.
- [ ] I have checked this feature using one of the following browser extensions and confirmed that it does not add any level A or level AA issues:
    * axe dev tools
    * Lighthouse
    * Wave
- [ ] ...

### Notes
The [figma diagrams were collectively reviewed](https://www.figma.com/design/BTU6c0uNhVFf4wLgB5p8BC/Orangelight-PO-workcycle?node-id=0-1&node-type=canvas&t=rxB9IEVeVMN6HD9r-0) during the [Orangelight meeting 10/17/2024.](https://docs.google.com/document/d/1GxpYrcRc6w3kMFU1XU6CEpnumFzvxFrj2GkCq_XvoWw/edit?tab=t.0#heading=h.s4fjyi2thf3c)

kevinreiss:
The stackmap integration looks for html like this to decide where to inject the "Where to find it" button

```
<span data-map-location="firestone$stacks" data-location-library="Firestone Library" data-location-name="Stacks"></span>```

I believe it also scrapes the call number value looking for this selector class "holding-call-number" on the record page and "call-number" on the search results page.

caroldh:
This will be addressed as part of the Show Page redesign

PR #4458, opened 2024-10-21
[#4453] Update Where to find 

christinach:
- Update Where to find it color scheme to be the same as the bootstrap btn-primary button

part of #4453


sandbergja (approved):
Thanks, @christinach 🐬

PR #4467, opened 2024-10-25
Remove font-display: fallback; unknown property

christinach:
Remove font-display: fallback; unknown property from the dev tools in chrome and firefox

### DIFF
--- a/app/assets/stylesheets/components/record.scss
+++ b/app/assets/stylesheets/components/record.scss
@@ -14,7 +14,6 @@
   .main-content {
     font-family: $font-sans;
     font-size: 1rem;
-    font-display: fallback;
     line-height: 1.5;
 
     h2,


sandbergja (commented):
Where is it saying that `fallback` is an unknown property?  It [is in MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/font-display).

christinach:
> Where is it saying that `fallback` is an unknown property? It [is in MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/font-display).

in the dev console. It doesn't recognize it both on chrome and firefox

maxkadel (commented):
I'm not sure I understand this PR - from the mdn docs this looks like a valid value? https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/font-display#fallback

christinach:
According to the docs and the rest of the scss files we have these properties ( font-display) should be wrapped in a @font-face {}. 
the record.scss should have this wrapped in a @font-face but it also needs an src to get the font. Otherwise it complains.

christinach:
If there are no objections I'm in favor to remove it since it doesn't do anything as a descriptor in this place.

sandbergja (approved):
Thanks, @christinach ! 🐬  And thanks for sharing what you found about `font-display` and `@font-face`

PR #4469, opened 2024-10-25
Upgrade lux to 5.9.0

sandbergja:
Includes a fix to the header menu bar from @ishasinha1 to close an open menu when you click outside of it.


christinach (approved):
Thank @ishasinha1 and @sandbergja ! 🍿 🍄

PR #4474, opened 2024-10-28
[#4344] Remove mm solr parameter when the user's query contains boolean OR

### LINKED ISSUE #4344: JSON Query DSL: OR searches are behaving the same as AND searches
### Expected behavior
When the JSON Query DSL feature is on, searches for [activism AND "social justice"](https://catalog.princeton.edu/catalog?search_field=all_fields&q=+activism+AND+%22social+justice%22) should return different results than [activism OR "social justice"](https://catalog.princeton.edu/catalog?search_field=all_fields&q=+activism+OR+%22social+justice%22)

### Actual behavior
They return the same results when the JSON Query DSL is on.

### Steps to replicate

1. Turn on the JSON query dsl in staging
2. Search for `activism AND "social justice"`
3. Search for `activism OR "social justice"`

### Impact of this bug

We can't turn on the new features in production, and migrating to Blacklight 8 is blocked

### Implementation notes

* I would expect Solr's edismax parser to recognize the all-caps AND and OR, maybe we have some code that is adjusting the queries before they reach solr?  Or something wrong with our solr configuration?.

sandbergja:
`mm` has a particular meaning when using the Boolean query parser, which is used when we send JSON queries like this to solr:

```
{"query":{"bool":{"must":[{"edismax":{"query":"activism OR \"social justice\""}}]}}}
```

In these cases, according to Solr's documentation, mm specifies the number of should clauses required for a document to be included in the results.  In our configuration, the default mm is 6<90%, meaning that if there are 6 or fewer clauses, all are required (and if there are 7 or more, 90% of the clauses are required).  This conflicts with and overrides the OR that the user intended.

Therefore, if we notice that the user's query includes OR in all caps, set the mm to 0 (meaning no minimum number of the clauses needs to be included).

Many thanks to @maxkadel for troubleshooting this with me!

closes #4344

### DIFF
--- a/app/models/search_builder.rb
+++ b/app/models/search_builder.rb
@@ -12,7 +12,8 @@ class SearchBuilder < Blacklight::SearchBuilder
                                      cjk_mm wildcard_char_strip excessive_paging_error
                                      only_home_facets prepare_left_anchor_search
                                      series_title_results pul_holdings html_facets
-                                     numismatics_facets numismatics_advanced]
+                                     numismatics_facets numismatics_advanced
+                                     adjust_mm]
 
   # mutate the solr_parameters to remove words that are
   # boolean operators, but not intended as such.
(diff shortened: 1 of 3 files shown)


maxkadel (approved):
Well done @sandbergja! This was a tricky one! 🍄‍🟫🍄🍄‍🟫

PR #4478, opened 2024-10-29
[#4472] Provide an abstraction of SearchService#fetch

### LINKED ISSUE #4472: Figure out why spec/requests/linked_records_spec.rb:6 is failing in blacklight 8
### Expected behavior

spec/requests/linked_records_spec.rb:6 passes on both the main and blacklight-8-gem branch

### Actual behavior

It fails on the blacklight-8-gem branch

### Steps to replicate

1. Check out the blacklight-8-gem branch
2. `bundle exec rspec spec/requests/linked_records_spec.rb:6`

```
Failure/Error: render json: document.linked_records(field: params[:field], maximum_records: 500).decorated

NoMethodError:
  undefined method `linked_records' for nil:NilClass

        render json: document.linked_records(field: params[:field], maximum_records: 500).decorated
                             ^^^^^^^^^^^^^^^
./app/controllers/concerns/orangelight/catalog.rb:82:in `linked_records'
./lib/orangelight/middleware/invalid_parameter_handler.rb:16:in `call'
./spec/requests/linked_records_spec.rb:6:in `block (2 levels) in <top (required)>'
```

### Impact of this bug

Can't upgrade to blacklight 8

sandbergja:
This method from the search service has a different API in Blacklight 7 vs. Blacklight 8.  This commit provides a small abstraction that checks whether we are using Blacklight 7 or 8, and then uses the appropriate API.

Closes #4472

### DIFF
--- a/app/services/retrieve_record_service.rb
+++ b/app/services/retrieve_record_service.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 # We can remove this service (and just call search_service.fetch directly)
 # after we migrate to Blacklight 8
+# :reek:UncommunicativeVariableName
 class RetrieveRecordService
   # :reek:DuplicateMethodCall
   # :reek:FeatureEnvy
@@ -15,8 +16,7 @@ class RetrieveRecordService
 
   private
 
-    # :reek:UtilityFunction
     def using_blacklight7?
-      Gem.loaded_specs['blacklight'].version.to_s.start_with? '7'
+      @using_blacklight7 ||= Gem.loaded_specs['blacklight'].version.to_s.start_with? '7'
     end
 end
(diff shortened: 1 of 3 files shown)


maxkadel (approved):
Thanks @sandbergja! 🍄

app/services/retrieve_record_service.rb
```diff
@@ -0,0 +1,22 @@
...
+    else
+      search_service.fetch(...)
+    end
+  end
+
+  private
+
+    # :reek:UtilityFunction
+    def using_blacklight7?
+      Gem.loaded_specs['blacklight'].version.to_s.start_with? '7'
```
maxkadel, line 20: This doesn't need to be blocking, but would it make sense to memoize this method? Or maybe even use an initializer to set this, so we don't have to check every time?

PR #4480, opened 2024-10-29
Use ServerItemPaginationComponent

### LINKED ISSUE #4187: Migrate our customized _previous_next_doc.html.erb to a view component

Migrate our customized `_previous_next_doc.html.erb` partial to a view component

### Level of urgency
- [ ] High
- [x] Moderate
- [ ] Low

- This partial is deprecated in Blacklight 7 and removed in Blacklight 8

### Acceptance criteria
 - [ ] All the logic and styles of our custom `_previous_next_doc.html.erb` partial are added to a view component that inherits from a view component in stock blacklight
 - [ ] We keep `_previous_next_doc.html.erb`, but all it does is render the new component

* As of today, Blacklight 8 uses a component named `Blacklight::SearchContext::ServerItemPaginationComponent` to provide the previous/next doc functionality.  Unfortunately, this component has not yet been backported to Blacklight 7.  See https://github.com/projectblacklight/blacklight/issues/3313
* Blacklight 7 does have a component called `Blacklight::SearchContextComponent` (it's actually the old version of `ServerItemPaginationComponent`, so it has a similar API), but which is not present in Blacklight 8.

maxkadel:
Closes #4187


sandbergja (approved):
Thanks, @maxkadel ! 🦄

PR #4493, opened 2024-10-31
Use search_service wrapper for bookmarks

maxkadel:
We have a rake task to update any existing Voyager ID bookmarks in the database, so we no longer need to migrate the ID number on the fly.


sandbergja (approved):
Thanks, @maxkadel 🦄

PR #4494, opened 2024-10-31
Have consistent window size

maxkadel:
Fixes some tests in BL8


sandbergja (approved):
Thanks, @maxkadel 🦄 !

PR #4501, opened 2024-10-31
Remove bookmarks customizations that are no longer needed

sandbergja:
This allows us to use Blacklight's built-in #index method, rather than needing to keep our own custom method compatible with both Blacklight 7 and 8.


kevinreiss (changes requested):
This all looks good. One small complaint from reek about a variable name.

kevinreiss (approved):

PR #4510, opened 2024-11-01
Move email tests to prepare for BL8 refactor

maxkadel:
Since BL8 uses a view component rather than model methods to generate the email text, move the tests to view tests so they will test the output, which should be the same for either BL7 or BL8.

Also factor out `using_blacklight7?` into initializer. I thought I was going to use it in this PR but wasn't able to get the BL8 email view component working yet, but I figured it was worth keeping anyway.


sandbergja (approved):
Thanks, @maxkadel 🦄

PR #4514, opened 2024-11-01
[#4513] Allow users to edit facet-only searches

### LINKED ISSUE #4513: Editing a basic facet search drops the facet
### Expected behavior
If a user does a facet search (for example, one that they found through a Subject (browse)), then edits the search, the search results will still be limited to the initial facet.

### Actual behavior
After the user edits their search, the search results are no longer limited to the facet

### Steps to replicate

In prod:
1. Do a Subject (browse) search for Romanticism
2. Choose the one with roughly 9000 results
3. Press the "Edit search" button
4. Add the keyword painting
5. Press the Search button
6. Note that now you have 267,112 results, and not all of them are about romanticism!

Locally:
1. Go to http://localhost:3000/?f[subject_topic_facet][]=Manuscripts%2C+Arabic&search_field=all_fields
3. Press the "Edit search" button
4. Add the keyword literature
5. Press the search button
6. Note that it is no longer limited to Arabic Manuscripts

### Impact of this bug

Users can't refine certain searches

### Notes

Thanks to Rebecca for [reporting the issue in the catalog channel](https://pulibrary.slack.com/archives/CA0UY0Q07/p1730496048302989).

sandbergja:
We need to include current facet values as hidden fields on the advanced search form, so that they are included in the subsequent search request.

Closes #4513

### DIFF
--- a/app/components/orangelight/advanced_search_form_component.html.erb
+++ b/app/components/orangelight/advanced_search_form_component.html.erb
@@ -7,7 +7,7 @@
   </div>
 <% end %>
 <%= form_tag search_catalog_path, method: @method, class: @classes.join(' '), role: 'search', 'aria-label' => t('blacklight.search.form.submit') do %>
-  <%= render Blacklight::HiddenSearchStateComponent.new(params: {'advanced_type' => 'advanced'}) %>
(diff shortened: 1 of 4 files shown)


christinach (approved):
Thanks @sandbergja !

Issue #4486, opened 2024-10-30
BL8: undefined method `to_email_text' for #<SolrDocument


maxkadel:
### Expected behavior
It can generate email text, and related tests pass

### Actual behavior
Tests for generating email text fail (`spec/models/email_spec.rb`) 

```
      NoMethodError:
        undefined method `to_email_text' for #<SolrDocument:33940 _source: {"id"=>"9947055653506421", "contained_in_s"=>["12345"]}>

                expect(solr_doc.to_email_text).to match('Holdings:')
                               ^^^^^^^^^^^^^^
        Did you mean?  to_semantic_values
      # ./spec/models/email_spec.rb:85:in `block (4 levels) in <top (required)>'
```

### Steps to replicate
Run ~`rspec spec/models/email_spec.rb:85`~ - moved test - run `rspec spec/views/record_mailer/email_record.text.erb_spec.rb`

### Impact of this bug
Can't upgrade to Blacklight 8

### Implementation notes, if any
It looks like the Blacklight 8 implementation of exporting as various formats is pretty different. See [email_action method on 8.x branch](https://github.com/projectblacklight/blacklight/blob/4ed913aa649a824605faaae58350a3d3bc6da4ec/app/controllers/concerns/blacklight/catalog.rb#L230-L237)

PR #4515, opened 2024-11-04
Inherit record mailer

maxkadel:
By inheriting the RecordMailer from Blacklight, rather than overlaying it, we make our code more upgrade-proof.

Connected to #4486


sandbergja (approved):
Thanks, @maxkadel 🦄

Issue #4518, opened 2024-11-04
Blacklight 8 - the search magnifying glass icon is not centered


sandbergja:
### Expected behavior

On the `bl8_iv` branch, the magnifying glass icon does not look super un-centered.

### Actual behavior
It is in the top left corner of the button.  I'm not convinced the icon is centered on main either, but it's not this bad!

![the icon is almost at the top of the button](https://github.com/user-attachments/assets/0634c3d1-7a38-4506-9ed8-58b8d0d9d7c4)

### Steps to replicate

Checkout the `bl8_iv` branch locally or deploy it to staging/qa.

### Impact of this bug

Looks very odd

sandbergja:
closed by #4519

PR #4519, opened 2024-11-04
[#4518] Center magnifying glass icon in search button on Blacklight 8

sandbergja:
This does not change the appearance of the blacklight 7 search button, but fixes the blacklight 8 button appearance.

The magnifying glass icon is centered, although it does look a little top heavy


christinach (approved):
thanks @sandbergja

PR #4532, opened 2024-11-05
[#4525] Display holdings data from bound-withs in record email and bookmark print

### LINKED ISSUE #4525: Blacklight 8: printing bookmarks doesn't have any metadata
### Expected behavior

The print view of bookmarks has useful information like the title, publisher, author, format, and holdings.

### Actual behavior

In blacklight 8, this view only contains the URL

### Steps to replicate

1. On catalog qa, add some bookmarks
2. Go to your bookmarks
3. Press the Print button

### Impact of this bug

This feature is pretty much useless without any data.  I'm not sure how often it is used, or for what use case, but I think anybody who does use it currently would not get their needs met by the current Blacklight 8 display.

sandbergja:
Refactor to use a View Component rather than a presenter, since it made it easier to render multiple holding blocks with the appropriate delimiters.

closes #4525 , also fixes the last failing test on the bl8_iv branch for me locally.


app/models/concerns/blacklight/document/email.rb
```diff
@@ -9,7 +9,6 @@ module Email
       def to_email_text
         body = []
         add_bibliographic_text(body)
-        add_holdings_text(body)
```
sandbergja, line 12: Removing the Blacklight 7-specific holdings display code, so that we can display the view component in either Blacklight 7 or 8.

kevinreiss (approved):

PR #4537, opened 2024-11-05
[#4529] Maintain the same selected facet heading color in Blacklight 7 and 8

### LINKED ISSUE #4529: Blacklight8: Selected facet header text does not have sufficient contrast with background
### Expected behavior
When you select a facet to search by, the resulting facet header should be accessible.

### Actual behavior
The facet header has insufficient contrast with the background

On BL7 (passes contrast tests)
image

On BL8 (fails contrast tests)
image

### Steps to replicate
Either
a. Go to https://catalog-qa.princeton.edu/?f%5Blocation%5D%5B%5D=Mudd+Manuscript+Library
b. Go to the library homepage and select one of the facets to limit by

### Impact of this bug
Not accessible, cannot upgrade to BL8

sandbergja:
Closes #4529

### DIFF
--- a/app/assets/stylesheets/components/facets.scss
+++ b/app/assets/stylesheets/components/facets.scss
@@ -174,6 +174,14 @@ a.more_facets_link {
   }
 }
 
+.facet-limit-active {
+  .card-header {
+    .btn {
+      color: $dark;
+    }
+  }
+}
+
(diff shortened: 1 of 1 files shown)


maxkadel (approved):
Thanks @sandbergja! 🍄

PR #4538, opened 2024-11-05
[#4534] Make Bookmark All button compatible with Blacklight 8

### LINKED ISSUE #4534: Blacklight 8: Bookmark all button doesn't work
### Expected behavior

The Bookmark all button adds all search results to the bookmarks

### Actual behavior

It doesn't add any search results to the bookmarks on catalog-qa (which is using blacklight 8).  It works okay on Blacklight 7.

### Steps to replicate

1. Go to catalog-qa
2. Do a search
3. Press the bookmark all button

### Impact of this bug

Blocks Blacklight 8 upgrade

sandbergja:
Our jquery code attempts to trigger an event, but it appears that Blacklight 8 is not listening to the events we send.  There are many changes in the JS between Blacklight 7 and Blacklight 8 (7 is in jquery while 8 is in vanilla js, 7 puts the event listener on the specific checkbox/button while 8 puts the event listener on the entire document).

I'm not sure why the jquery event triggering doesn't work with the Blacklight 8 js, but migrating from jquery to vanilla js seems to do the trick in both Blacklight 7 and 8.

Closes #4534

### DIFF
--- a/app/javascript/orangelight/bookmark_all.es6
+++ b/app/javascript/orangelight/bookmark_all.es6
@@ -1,19 +1,21 @@
 export default class BookmarkAllManager {
-
   constructor() {
-    this.element = $("#bookmark_all_input");
+    this.element = $('#bookmark_all_input');
     this.prepopulate_value();
     this.bind_element();
   }
 
   prepopulate_value() {
-    if ($("input.toggle-bookmark:checked").length == $("input.toggle-bookmark").length) {
+    if (
+      $('input.toggle-bookmark:checked').length ==
+      $('input.toggle-bookmark').length
+    ) {
       this.element.prop('checked', true);
     }
   }
 
   bind_element() {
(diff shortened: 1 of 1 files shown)


app/javascript/orangelight/bookmark_all.es6
```diff
@@ -1,19 +1,21 @@
 export default class BookmarkAllManager {
-
```
sandbergja, line 2: Some automated lint fixes

app/javascript/orangelight/bookmark_all.es6
```diff
@@ -30,10 +32,14 @@ export default class BookmarkAllManager {
   }
 
   bookmark_all() {
```
sandbergja, line 34: The actual changes are in these two methods: `bookmark_all` and `unbookmark_all`

PR #4540, opened 2024-11-05
[#4531] Remove pivot facet icon customization

### LINKED ISSUE #4531: Facet Expansion Icon on Classification Facet is Doubled
### Expected behavior
When I open the classification facet I expect to see only one expansion icon. 

### Actual behavior
I see two expansion items. Using the facet works as expected. 

[
](url)

### Steps to replicate
1. Run a search
2. Expand the Classification Facet. 

### Impact of this bug
Interface looks wonky.

sandbergja:
Since we're moving away from the pivot facet anyway (see #4505), I thought we might not mind moving away from our customized pivot facet icon.

The default Blacklight 7 icon doesn't look the greatest (it's a little small), but the default Blacklight 8 icon looks pretty similar to our customized one.

In both Blacklight 7 and 8, the icon is lined up vertically with the facet text, which is a nice improvement over main.

Blacklight 8 default icon:
large plus and minus icons

Blacklight 7 default icon:
Small plus and minus icons

Our customized icon:
large plus and minus icons, but the icon is much higher vertically than the text

Closes #4531

### DIFF
--- a/app/assets/stylesheets/components/facets.scss
+++ b/app/assets/stylesheets/components/facets.scss
@@ -211,3 +211,18 @@ h3.facet-field-heading {
   margin-top: 0px;
   padding: .3rem .9rem !important;
 }
+
+/* Blacklight 7 facet customization, we can remove when Blacklight 8 is merged
+   (this rule specifically matches the Blacklight 7 DOM, it does not match anything
+   in Blacklight 8) */
+li:not(.treeitem) button.facet-toggle-handle {
+  font-size: 1.75rem;
+}
+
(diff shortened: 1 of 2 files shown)


maxkadel (commented):
Is there a way to check that the BL7 icon is a big enough target for people with physical difficulty targeting a mouse?

kevinreiss:
I agree the Blacklight 8 looks fine and there is definitely no need to keep our custom icon here.

kevinreiss:
To make the BL 7 icon look a little better as a short time fix could use CSS to increase how it displays since it's a vector graphic?

kevinreiss (approved):

Issue #4539, opened 2024-11-05
Blacklight 8: the Bookmarks count doesn't update when you add/remove bookmarks


sandbergja:
### Expected behavior

When I add or remove a bookmark from the search results page, the "Bookmarks (4)" counter will adjust to match the number of bookmarks.

### Actual behavior
It works on Blacklight 7, but not Blacklight 8 unfortunately.  The number does not automatically adjust, you have to refresh the page.

### Steps to replicate

1. Go to catalog qa
2. Do a search
3. Add / remove some bookmarks and watch the counter to see if it updates

### Impact of this bug

User gets conflicting information about whether or not their bookmark saved.

sandbergja:
I believe this is the same issue as https://github.com/projectblacklight/blacklight/issues/3350 -- perhaps the easiest fix would be to backport https://github.com/projectblacklight/blacklight/pull/3351 and upgrade Blacklight once it is released.

sandbergja:
This is now working on the branch of #4546, so I'm closing it.

PR #4544, opened 2024-11-06
[#4539] Add test to confirm that bookmark count adjusts when you remove a bookmark.

sandbergja:
Currently, this test passes on Blacklight 7, fails on Blacklight 8.3, and passes on Blacklight 8.6.1 thanks to @kirkkwang's work in https://github.com/projectblacklight/blacklight/pull/3351

Helps with #4539


christinach (approved):
Thanks @sandbergja ! 🍄

PR #4553, opened 2024-11-09
[#4552] More reliable boolean operators on the advanced search form

### LINKED ISSUE #4552: Advanced search: "OR" is not working sending the wrong clause operators
### Expected behavior
If a user enters "Academic dissertations" OR "Academic theses" into the advanced search form, we translate it into `{should:"Academic dissertations"}` and `{should:"Academic theses"}` before sending it to solr.

### Actual behavior

It is translating it into `{must:"Academic dissertations"}` and `{should:"Academic theses"}`

This gives the user the exact same results they would get from just searching "Academic dissertations" by itself, rather than sprinkling in the "Academic theses" results as well.

### Steps to replicate

1. Go to the advanced search screen
2. In the first and second dropdowns, select "Subject"
3. In the first input, enter `"Academic dissertations"`
4. In the second input, enter `"Academic theses"`
5. Set the first radio button to "OR"
6. Search.  Note the number of results
7. Remove the "Academic theses" constraint
8. Note that the number of results didn't change.  In other words, the user's query `"Academic theses"` didn't influence the search at all.

### Impact of this bug

Difficult to do a boolean OR search in the advanced search form (the only way to do it is: leave the first input blank, set both radio buttons to "OR", and then type your queries into the second and third inputs)

### Notes

Thanks to @minjiec for reporting this issue!

sandbergja:
Rather than trying to passing the boolean operator parameters that Blacklight expects directly from the advanced search form (e.g. `clause[0][op]=must` in the URL query params), instead pass the boolean operator that the user sees in the UI (e.g. `boolean_operator1=OR`).  Then, pass it along to a new class `AdvancedBooleanOperators`, which handles the various different combinations a user might enter and converts them to the parameters Blacklight expects.

closes #4552

### DIFF
--- a/.reek.yml
+++ b/.reek.yml
@@ -50,6 +50,7 @@ detectors:
     - Requests::ApplicationHelper#hidden_service_options
     - Requests::ApplicationHelper#single_pickup
     - Requests::RequestHelper#reshare_type
+    - AdvancedBooleanOperators#initialize
     - Blacklight::Document::JsonLd#check_role
     - Blacklight::Document::JsonLd#date
     - Requests::Aeon#holding_location_to_site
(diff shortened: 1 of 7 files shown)


christinach (approved):
Thanks @sandbergja 🍄

PR #4558, opened 2024-11-11
[#4555] Hide skip links when not focused

### LINKED ISSUE #4555: Styles have changed for skip links
### Expected behavior

Skip links look as they do in the v7.0-blacklight-7-final tag

### Actual behavior

They look different: they display even if not focused.  Also, they are very small vertically.

### Steps to replicate

Go to the catalog

### Impact of this bug

Visual regression, it makes the skip links smaller to use

I thought #4547 would fix this, but it didn't.

sandbergja:
Closes #4555


christinach (approved):
Thanks @sandbergja

PR #4574, opened 2024-11-14
4081 bootstrap5 on blacklight7

christinach:


christinach:
I rebased the initial branch because it includes the review and also I would like to merge it as soon as it is ready and not waiting for the BL 8 upgrade fixes.

PR #4583, opened 2024-11-18
Change to deliver_later and some test fixes

kevinreiss:


rladdusaw:
Everything works in test, but when I deployed it to qa I was unable to see any emails being delivered. I'm wondering if we need to add a job queue like Sidekiq for this to work properly.

christinach (approved):
Thanks @rladdusaw ! 🐈‍⬛

PR #4592, opened 2024-11-19
pull-right pull-left are from the fontawesome library.

christinach:
currently pull-right pull-left are from the fontawesome library.
Replace then with float-left float-right from Bootstrap 4.x. Helps with the migration to Bootstrap v5
Keeps a consistency with the libraries we use to float items


regineheberlein (approved):
Thanks @christinach !

PR #4602, opened 2024-11-27
[#4566] Move Next and Previous text in the actual <a> tag so that the

### LINKED ISSUE #4566: "previous" and "next" navigation buttons do not remain when looking through catalog records from a list of results
### Expected behavior
I would expect that navigation buttons remain as an option, so that I can continue to view next (or previous) records in my list of results. 

### Actual behavior
I have a list of search results, and when I click into the 1st record, I can see the  "previous" and "next" available for me to navigate through the list. But if I click "next" to view the next record in the list, I no longer see the "previous" and "next" buttons. 
1st record, where navigation is possible:
{400431BF-1B8C-430E-8BFA-A05559CBA262}

After clicking "next" to see next record, buttons are no longer available:
{64E94009-F76C-4EB4-98E1-FD1EDF9E796E}

### Steps to replicate
Execute a search in catalog and get a list of results. Click on one of the records, and use "previous" or "next" to move in the list of results. On next record, you will no longer see options for "previous" or "next"

### Impact of this bug
Cannot easily move through a list of search results.

[Related Slack thread](https://pulibrary.slack.com/archives/C031U4XCUH2/p1731683169255359)

christinach:
closes #4566 

Tested and works both on main and the Blacklight8-upgrade-v2 branch .

Move Next and Previous text in the actual <a> tag so that the event target in

https://github.com/projectblacklight/blacklight/blob/main/app/javascript/blacklight-frontend/search_context.js#L2 always succeeds on the check regardless of where in the button the user clicks.

### DIFF
--- a/config/locales/blacklight.en.yml
+++ b/config/locales/blacklight.en.yml
@@ -1,8 +1,8 @@
 en:
   views:
     pagination:
-      previous: '&laquo; <span class="d-none d-lg-inline"> Previous</span>'
-      next: '<span class="d-none d-lg-inline">Next</span> &raquo;'
+      previous: '&laquo; Previous'
+      next: 'Next &raquo;'
   blacklight:
     advanced_search:
       form:


sandbergja (approved):
Thanks, @christinach 🐬 🍄  !

PR #4604, opened 2024-11-27
Update search_options.scss to apply the flex attributes only to class search-pagination-widgets

christinach:


sandbergja (approved):
Thanks, @christinach 🐬 🍄

PR #4605, opened 2024-11-27
Change harmful-content-bar not to be sticky

christinach:
This came up in one of the Fable user testing sessions where the user, testing with a screen magnifier accessibility tool, could not read the content of the page since the sticky bar was covering a large percentage of the page.

### Current record page in production:

Current record page in production with sticky bar

### Record page with no sticky bar:

Record page with no sticky bar


sandbergja (approved):
Looks good to me!  Thanks, @christinach 🐬

Issue #4569, opened 2024-11-13
Blacklight 8 [bookmarks]: In the bookmarks page when clicking the 'Next' button it triggers '..something went wrong'


christinach:
### Expected behavior
When I view the bookmarks page I can click on the Next or Previous button and view my bookmarks.

### Actual behavior
When I view the bookmarks page and I click on the Next or Previous button I get the error page 'Looks like something went wrong'.

### Steps to replicate
1. checkout blacklight 8 branch
1. Bookmark more than 10 books.
1. Go to https://catalog.princeton.edu/bookmarks?per_page=10.
1. Click the 'Next' button

### Impact of this bug
The user cannot view the bookmarks on the next page.

Related bug: #4568.  #4569 is regarding the "Next" and "Previous" buttons, rather than the "2", "3", "4" pagination at the bottom.

@maxkadel identified the cause of this issue to be the same as https://github.com/pulibrary/orangelight/issues/4568. 
Most likely this ticket will be resolved when https://github.com/pulibrary/orangelight/issues/4568 is resolved.

christinach:
closed by https://github.com/pulibrary/orangelight/pull/4625/files
Thanks @sandbergja !

Issue #4568, opened 2024-11-13
Blacklight 8: Pagination in Bookmarks page does not work


christinach:
### Expected behavior
When I view the bookmarks page I can click on the pagination to view them.

### Actual behavior
When I view the bookmarks page and I click on the pagination to view them I get the error page 'Looks like something went wrong'.

### Steps to replicate
1. Checkout the blacklight 8 branch
1. Bookmark more than 10 books.
1. Go to https://catalog.princeton.edu/bookmarks?per_page=10.
1. Click the second page in the pagination.

### Impact of this bug
The user cannot paginate through the bookmarks.

Related ticket: #4566

maxkadel:
Since I am leaving for the week, I'm putting this back down.

I wrote a [test that fails against Blacklight 8](https://github.com/pulibrary/orangelight/pull/4612) and passes against Blacklight 7. 

On blacklight 8, it seems that the request is going through the search_builder and hitting the [excessive_paging](https://github.com/pulibrary/orangelight/blob/8ddfbea8537489bde0a85779db72793f8b5fb06b/app/models/search_builder.rb#L76-L90) checks, which raises an error.

My best guess is that this is because either
a) In Blacklight 7 it doesn't go through the search_builder in the same way, and so doesn't hit those checks, or
b) In Blacklight 8 it's not getting passed with its full context, and so fails those checks.

PR #4612, opened 2024-12-04
Test failing against Blacklight 8 for the right reason

maxkadel:
Should pass against Blacklight 7

Connected to #4568
Connected to #4569


christinach (approved):
Thanks @maxkadel ! 🦄

PR #4483, opened 2024-10-30
Remove advanced search gem

maxkadel:
- Doing this off of main, since the old branch was getting stale

Closes #3679

PR #4618, opened 2024-12-06
Remove old numismatics form

sandbergja:
It is no longer possible to reach it after #4483 was merged


christinach (approved):
thanks @sandbergja

PR #4621, opened 2024-12-06
add catalog5 to config/deploy/production.rb

Beck-Davis:
related to https://github.com/pulibrary/princeton_ansible/issues/5604


sandbergja (approved):
Thanks, @Beck-Davis 🐻‍❄️ !

PR #4632, opened 2024-12-11
4557 honeybadger env

### LINKED ISSUE #4557: Honeybadger should reflect Rails environment
### Expected behavior
You should be able to filter Honeybadger to production errors, and see only errors from Production servers

### Actual behavior
Errors from staging servers are included in production search results
image

### Steps to replicate
- Go to [Honeybadger search, filtered to "production" environment](https://app.honeybadger.io/projects/54399/faults?sort=last_seen_desc&q=environment%3A%22production%22%E2%80%81-is%3Aresolved%E2%80%81-is%3Aignored%E2%80%81)
- Click into some of the errors and look at the "Location" section in the summary. It should look something like "Location: production on catalog-staging1" - note that the host is not a production host

### Impact of this bug
Makes the runner job more difficult

See - https://docs.honeybadger.io/lib/ruby/getting-started/environments/ - it's supposed to inherit from the `RAILS_ENV` environment variable, so perhaps this is not getting set correctly? Or perhaps `HONEYBADGER_ENV` is being set incorrectly?

We could also use a configuration file - see e.g. [allsearch_api](https://github.com/pulibrary/allsearch_api/blob/main/config/honeybadger.yml)

rladdusaw:
Resolves #4557 

The setup of Honeybadger was incomplete and we missed requiring Honeybadger in the Capfile. 
https://docs.honeybadger.io/lib/ruby/getting-started/introduction/

### DIFF
--- a/Capfile
+++ b/Capfile
@@ -16,6 +16,8 @@ require "capistrano/passenger"
 
 require 'whenever/capistrano'
 
+require 'capistrano/honeybadger'
+
 # Includes tasks from other gems included in your Gemfile
 #
 # For documentation on these, see for example:
--- a/config/honeybadger.yml
+++ b/config/honeybadger.yml
@@ -1,4 +1,3 @@
 
 exceptions:
   rescue_rake: true
-env: "<%= Rails.env %>"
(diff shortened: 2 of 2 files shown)


christinach (approved):
Thanks @rladdusaw !

PR #4633, opened 2024-12-11
Use the new API slot with_* for ModalComponent

christinach:
It became available in view_component version 2.54 https://viewcomponent.org/CHANGELOG.html#2540. Currently we're using version 2.83.0


regineheberlein (approved):
Thanks @christinach !

PR #4643, opened 2024-12-13
[feedback form] Move modal-footer section outside of modal-body

christinach:


sandbergja (approved):
Thanks, @christinach ! 🐬

Issue #4641, opened 2024-12-12
[Blacklight 8 and main ] Feedback form should display an error to the user when the input fields are invalid and submit the form once


christinach:
### Expected behavior
Feedback form should display an error to the user when the input fields are invalid
It should submit the form once.

### Actual behavior

Feedback form will not display an error to the user when the input fields are invalid

It submits the form twice. fixed in https://github.com/pulibrary/orangelight/pull/4645/files

### Steps to replicate
Click on Ask a question.
Add an invalid email or no email. The modal will close without validating first the value.

### Impact of this bug
E.g. "I can't work until this is fixed" or "I have a workaround"

### Acceptance criteria
- [ ]  Feedback form displays an error to the user when the input fields are invalid

- [x]  It submits the form once. fixed in https://github.com/pulibrary/orangelight/pull/4645/files

### Implementation notes, if any
I think this is because of the remote: true in combination with the [data-bl-dismiss](https://github.com/pulibrary/orangelight/pull/4582/files#diff-0f36f38ec0a3b7d7e102a8aad24abc69b6e0c4aa5f4a1267b8de047ff610e429R22) that was added here so that the modal will close when the user submits the form.

It will not submit anything in the queue if the values are invalid. But the modal will close and there will be no indication to the user that they have typed an invalid value or they missed a required one.

#### Update Friday 13th, 2024
The main branch is fixed.
The blacklight 8 branch will close the modal and there will be no indication to the user that they have typed an invalid value or they missed a required one.

[fixes the double submission](https://github.com/pulibrary/orangelight/pull/4645/files)

[Fix main to have again client side validation errors](https://github.com/pulibrary/orangelight/pull/4646)

christinach:
[Experiment and working branch](https://github.com/pulibrary/orangelight/tree/experiment-with-remote-true-modal-close)

christinach:
resolved by https://github.com/pulibrary/orangelight/pull/4654

PR #4645, opened 2024-12-13
[#4641] Partof #4641 - submit form only once

christinach:
part of #4641


sandbergja (approved):
Thanks, @christinach 🐬

PR #4646, opened 2024-12-13
[#4641]Bring production to the state where we have client side validation errors

christinach:
part of #4641


sandbergja (approved):
Thanks, @christinach 🐬

PR #4647, opened 2024-12-13
Catch when patron_hash is TrueClass

### LINKED ISSUE #4581: In Requests, handle errors from Bibdata gracefully when building patron
### Expected behavior
When Requests cannot build a patron because Bibdata is returning an html error from the load balancer (perhaps 5XX errors?), the application should show a helpful message to the user, and fail gracefully.

### Actual behavior
The application raises an error and probably just shows the generic "what went wrong" page.

### Steps to replicate
- Not sure - happened because when the Catalog was trying to build the patron, when it requested patron info from bibdata, it got the "Something went wrong" page from the load balancer instead, and was not caught by any of the rescues.

### Impact of this bug
If a person is not able to create a request because Bibdata is down, they do not know why the request failed or what the current status is, or what their next steps might be.

https://app.honeybadger.io/projects/54399/faults/114615544/01JCWW9RV765NT1H0AVMEQ50F2
```
NoMethodError: undefined method `[]' for true:TrueClass patron_hash[:patron_group] ^^^^^^^^^^^^^^^
44 |  
-- | --
45 | def patron_group
46 | patron_hash[:patron_group]
47 | end
```

When the load balancer passes on an error, it is usually an html page, and may not raise any of the errors in the Bibdata service

rladdusaw:
Resolves #4581


app/models/requests/patron.rb
```diff
@@ -94,7 +94,10 @@ def to_h
 
       def load_patron(user:)
         patron_hash = current_patron_hash(user.uid)
-        errors << "A problem occurred looking up your library account." if patron_hash.blank?
+        if patron_hash.instance_of?(TrueClass) || patron_hash.blank?
```
maxkadel, line 97: For me, it would make more sense if we caught this error lower down in the stack trace, and make sure that this method doesn't get a patron_hash that's an instance of TrueClass.

sandbergja: +1

sandbergja (changes requested):
Thanks, @rladdusaw -- I agree with @maxkadel that we should catch this at a lower level if possible.  I think it would be beneficial for two reasons:
* Hopefully we could avoid having a type check, and
* TrueClass gives the vibes of something happened successfully -- at least to me-- so I find it counterintuitive here that it actually means that getting the hash errored.  It would be nice if we could avoid that.

sandbergja (approved):
Thanks, @rladdusaw and @maxkadel ! 🐈‍⬛ 🦄

maxkadel (approved):
Thanks @rladdusaw!

PR #4658, opened 2024-12-19
Remove unused Stackmap concern

### LINKED ISSUE #4477: Can we remove Orangelight::Stackmap concern?

- [ ] Check if we still need the code in the file app/controllers/concerns/orangelight/stackmap.rb
- [ ] If we don't need it, delete it

### Level of urgency
- [ ] High
- [ ] Moderate
- [x] Low

This doesn't appear to be called when opening the stackmaps modal.  If it is no longer needed, we can simplify the code base a little!

maxkadel:
Closes #4477


christinach (approved):
Thanks @maxkadel

PR #4660, opened 2024-12-20
Include the solr parameters in Publisher and Notes fields for the advanced search

christinach:
Include the solr parameters in Publisher and Notes fields for the advanced search

Without these they behave like the keyword search
This is fixing a regression related to removing the advanced search gem and replacing it with the built-in advanced search

Thanks to @caroldh who found this bug! 🐛


sandbergja (approved):
🐬 🍄 🐛

PR #4661, opened 2024-12-20
Remove unused method hidden_fields_holding from application_helper

christinach:


sandbergja (approved):
Thanks, @christinach ! 🐬

PR #4662, opened 2024-12-24
[#4655] Server side validation for report harmful language modal

### LINKED ISSUE #4655: [Blacklight 8 branch] Report harmful language - server side email validation error -  (optional field)
### Expected behavior
The form has an optional field for email. If the email does not end in a top level domain then it will display `Email is not a valid email address`

### Actual behavior
The form has an optional field for email. When the email does not end in a top level domain then it will not display `Email is not a valid email address`

### Steps to replicate
Go to a record page.
Click on 'Report Harmful Content'

### Impact of this bug
If a user misspells their email and has no top level domain then the form will appear to submit successfully but never arrive in the queue.

sandbergja:
* Make sure that the form has data-remote="true" when in a modal, but not when it is outside a modal.  This way, Rails UJS will handle form submission as an AJAX request while in a modal, but as a normal form submit while outside the modal.
* When form results are invalid, return HTTP status code 422, so that the front end knows to show the validation messages.

Closes #4655

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -26,8 +26,18 @@
 
 // Wait for the modal to open
 document.addEventListener('show.blacklight.blacklight-modal', function () {
+  // Add data-remote=true so that Rails UJS knows to submit the form as an AJAX request
+  document
+    .querySelector('#new_report_harmful_language_form')
+    ?.setAttribute('data-remote', true);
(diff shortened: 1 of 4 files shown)


christinach (approved):
Thanks @sandbergja ! 🍄

christinach (commented):
@sandbergja I missed that we forgot to remove the remote true from the other forms.

Issue #4025, opened 2024-05-30
Language and Facet subjects are not very user-friendly in A-Z version


sandbergja:
### User story

* As a catalog user, I'd like to limit my results to Tagalog language results, when Tagalog is not one of the most common languages in my search results.
* As a catalog user, I'd like to limit my results to the subject Medicine, when Medicine is not of the most common subjects in my search results

### Current display

Currently a user has to:

1. Press "more" in the facet
2. Press "A-Z"
3. Keep paging through until they get to the desired facet value.  Loading each page of results can take some time, too.

Power users may open the "more" link in a new tab, which allows them to adjust the URL to get closer to the desired page, but even if you are aware of this workaround, you still have to page through a few results until you find just the right page.

### Desired display

In the [#catalog channel](https://pulibrary.slack.com/archives/CA0UY0Q07/p1717084264845829), Joshua suggested:

> If we could have an input bar in which you start typing and a hotlist that changes based on the input while typing, this would make selecting a language so much easier. Perhaps looking  similar to the second image .
![suggested facet display of languages, with a typeahead combobox to limit the language results](https://github.com/pulibrary/orangelight/assets/7086943/c3e4df21-7943-45a6-8ad8-d4ddecd860af)

### Acceptance criteria
- [ ] Create a wireframe
- [ ] Once this wireframe is approved, implement it and resolve this ticket.
- [ ] I can successfully use this feature with Voice Over in Safari.
- [ ] I have checked this feature using one of the following browser extensions and confirmed that it does not add any level A or level AA issues:
    * axe dev tools
    * Lighthouse
    * Wave
- [ ] The UI is easier and more efficient to use when trying to limit to a specific, less common Subject or Language.

### Notes
Related [blacklight/issues/2123](https://github.com/projectblacklight/blacklight/issues/2123)

ellen-aa:
This feature seems to be the way the Language facet works in Advanced Search. Carol and I think that the A-Z list and the Numerical sort are useful, but if there's a way to add the ability to type (as in the Language facet for Advanced Search), that would be very helpful (for the Catalog and for DPUL!).

sandbergja:
@ishasinha1 , @christinach , and I collaborated on an upstream PR: https://github.com/projectblacklight/blacklight/pull/3367

kevinreiss:
@sandbergja trying to backport work from BL9 to BL8/7. Working on some failing tests.

maxkadel:
[...]

PR #4671, opened 2024-12-31
Don't filter out non-home facets on facet suggest searches

sandbergja:
This allows us to do facet.contains searches, which are needed for #4025, for facet fields like language, which don't show up on the home page.

Helps with #4025


christinach (approved):
Thank you @sandbergja ! 🍄 🍄🍄 🍄🍄 🍄🍄 🍄🍄 🍄🍄 🍄

PR #4672, opened 2025-01-01
Improve performance of the solr query used to generate the advanced search form

sandbergja:
Running this locally against the production solr sped the solr query up from a median 10,691 ms to a median 1,851 ms.

This removes several potentially expensive parameters from the solr query we send to generate the advanced search page:
* facet.query
* facet.pivot
* stats
* stats.field
* facet.field, except for the facet field values we need to retrieve to render the advanced search form

Since the logic is different on the numismatics form vs advanced form, this commit
splits them into two separate search builders.

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -841,8 +841,10 @@ class CatalogController < ApplicationController
     end
 
     def search_service_context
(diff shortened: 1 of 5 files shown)


christinach (approved):
Thanks @sandbergja !

PR #4678, opened 2025-01-02
Simplify query for facet more link

sandbergja:
When we use the "more" link on a facet, we send a solr query to gather the facet values.  Prior to
this commit, we also sent some extra parameters for other facets.  For example, we always sent the facet.query params for the publication date and cataloged date, even if we were looking at a totally different facet, say, language.

Working locally with a branch for #4025 and catalog-staging, the median speed for a suggestion query goes down from 2192 ms to 1258 ms.

Helps with #4025, so that the feature is performant enough to be useful

### DIFF
--- a/app/models/search_builder.rb
+++ b/app/models/search_builder.rb
@@ -12,7 +12,7 @@ class SearchBuilder < Blacklight::SearchBuilder
                                      only_home_facets prepare_left_anchor_search
                                      series_title_results pul_holdings html_facets
                                      numismatics_facets numismatics_advanced
-                                     adjust_mm]
+                                     adjust_mm remove_unneeded_facets]
 
   # mutate the solr_parameters to remove words that are
   # boolean operators, but not intended as such.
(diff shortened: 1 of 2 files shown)


app/models/search_builder.rb
```diff
@@ -107,6 +107,18 @@ def adjust_mm(solr_parameters)
     solr_parameters['mm'] = 0
   end
 
+  # When the user is viewing the values of a specific facet
+  # by clicking the "more" link in a facet, solr doesn't
+  # need to perform expensive calculations related to other
+  # facets that the user is not displaying
```
christinach, line 113: 👍🏻

app/models/search_builder.rb
```diff
@@ -137,4 +149,11 @@ def add_edismax(advanced_fields:)
         blacklight_config.search_fields[field]['clause_params'] = { edismax: }
       end
     end
+
+    # :reek:FeatureEnvy
+    def remove_unneeded_stats(solr_parameters)
```
christinach, line 154: 👍🏻

christinach (approved):
Thanks @sandbergja ! 🍄 🍄

PR #4680, opened 2025-01-04
Advanced form: only process facets that have include_in_advanced_search: true

sandbergja:
Before this commit, the AdvancedSearchFormComponent processed any field in the CatalogController that did not have include_in_advanced_search: false.

In theory, this should not be particularly expensive, but a subtle bug in Blacklight causes some performance issues in cases where we try to process fields that aren't included in the Solr response and when the solr response contains many facet values -- basically the exact situation of our advanced search form!  See
https://github.com/projectblacklight/blacklight/issues/3484 and https://github.com/projectblacklight/blacklight/pull/3485 for further details.


kevinreiss (approved):
Thanks @sandbergja!

christinach (approved):
Thanks @sandbergja !

PR #4682, opened 2025-01-06
Remove stackmap route and template

### LINKED ISSUE #4667: Remove Stackmap Route and Views that are no longer used
### Expected behavior
Bots are calling unused stackup URL is visited it should not produce a 500 error. We can just remove this code and let these requests go to 404 errors. 

### Actual behavior
A 500 error is produced when certain parameters are submitted to the Stackmap Endpoint 

### Steps to replicate
Visit https://catalog.princeton.edu/catalog/99108120423506421/stackmap?loc=mendel$stacks&cn=ML275.5+.P475+2018  a 500 error will be produced. 

### Impact of this bug
It's unclear to me if actual users or bots are generating this error as there does not seem to be a referring URL logged in any of the honeybadger alerts this error has generated. 

See https://app.honeybadger.io/projects/54399/faults/115594155/01JGCPT26SRDAAYXB6AMYV439P

```
<% titlevalue = @document['title_vern_display'] %>
```
Is the line that repeatedly produces the "NoMethodError: undefined method `[]' for nil " error.

It isn't clear to me what the referring URLs are that are producing these errors. Visiting the catalog show page for the MMS IDs in question does not generate an error. All the pages where I see a stackmap button in the holdings page seem to work fine. We think the URL these errors are consulting is not actually used in the application any longer. If it exists as a route we should just remove it and let these go to 404 errors. 

### Acceptance criteria

- [x] Remove the stackmap route entirely as we no longer utilize it in the application.
- [x] Remove the associate erb file related to this.
- [x] Remove specs associated with this feature.

rladdusaw:
Resolves #4667


sandbergja (approved):
Thanks, @rladdusaw ! 🐈‍⬛

PR #4683, opened 2025-01-06
Don't request documents from solr for the advanced search form

sandbergja:
Running the solr query with `debug=timing` doesn't indicate any time spent on selecting the documents or retrieving their metadata.  I did however see a slight performance improvement when running `curl` with `rows=0` over the previous `rows=20`. After running it 20 times, it was consistently about 100ms faster with `rows=0`. At the very least, it's a smaller response to send over the network or to write/read to the on-disk cache.


christinach (approved):
Thank you @sandbergja ! 🍄

PR #4694, opened 2025-01-10
Ruby 3.4.1

sandbergja:


christinach (approved):
Thanks @sandbergja !

christinach (commented):
@sandbergja will you do a bundle update?

PR #4700, opened 2025-01-14
Use sidekiq as email delivery queue

### LINKED ISSUE #4609: Use SideKiq for email ActiveJob queues
### User story
While the default Rails queue seems to work, SideKiq would give us better visibility and potentially better reliability for jobs

### Acceptance criteria
The application uses SideKiq for queues and can be deployed

- [ ] I can send emails using SideKiq
- [ ] I can track asynchronous email jobs in SideKiq
- [ ] Emails sent on the staging and qa environments are caught by mailcatcher
- [ ] Deployment does not cause emails to get lost.
- [x] Use the catalog-indexers for sidekiq
- [x] Run it a systemd service (ansible update) 

The endpoint for /sidekiq will need some form of authentication to prevent unauthorized users from accessing the web frontend.

sandbergja:
Also, put all indexers in an indexer capistrano role

closes #4609

### DIFF
--- a/.lando.yml
+++ b/.lando.yml
@@ -22,7 +22,7 @@ services:
     type: postgres:15
     portforward: true
   orangelight_redis:
-    type: redis:6.0.16
+    type: redis:7
     portforward: true
 proxy:
   orangelight_test_solr:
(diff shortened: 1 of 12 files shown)


.lando.yml
```diff
@@ -22,7 +22,7 @@ services:
     type: postgres:15
     portforward: true
   orangelight_redis:
-    type: redis:6.0.16
+    type: redis:7
```
sandbergja, line 25: This matches the redis version we use in prod

christinach (approved):
Thanks @sandbergja @maxkadel 🍄 🦄 !

PR #4702, opened 2025-01-14
Revert "Use sidekiq as email delivery queue"

sandbergja:
Reverts pulibrary/orangelight#4700

It seems that we will need to refactor much of requests emailing to pass in simpler, more sidekiq-compatible classes, we can reopen this then.


maxkadel (approved):
Thanks @sandbergja

PR #4706, opened 2025-01-16
Add Requests::Submission#to_h method that is sidekiq-compatible

sandbergja:
Sidekiq is [very restrictive about which data structures we can pass in with jobs, including to email jobs](https://github.com/sidekiq/sidekiq/wiki/Best-Practices#1-make-your-job-parameters-small-and-simple).

Our `Requests::Submission` object was not built with sidekiq's restrictions in mind, so this commit adds a method that can convert the object into one that is compatible with Sidekiq.

Note that we do not yet use this method anywhere, this is just a preparatory step.  The next steps will be to start incrementally teaching our mailer views to use the `to_h` output rather than `Requests::Submission` directly.

Helps with #4609

### DIFF
--- a/app/models/requests/submission.rb
+++ b/app/models/requests/submission.rb
@@ -84,7 +84,7 @@ module Requests
     end
 
     def pick_up_location
-      Requests::BibdataService.delivery_locations[items.first["pick_up"]]["library"]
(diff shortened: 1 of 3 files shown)


kevinreiss (approved):

christinach (approved):
Thanks @sandbergja 🍄 !

PR #4707, opened 2025-01-16
Use redis 7 in lando to match our prod redis

sandbergja:
This is an incremental step toward #4609, so we can actually run sidekiq locally


christinach (approved):
Thanks @sandbergja and @maxkadel !

PR #4708, opened 2025-01-16
Add sidekiq gem to project

sandbergja:
This is an incremental step toward #4609

Hopefully it also fixes CI 😬


christinach (approved):
Thanks @sandbergja and @maxkadel

PR #4710, opened 2025-01-16
exclude vite server from connect_src so that we resolve

christinach:
Refused to connect to 'ws://localhost:3036/vite-dev/' because it violates the following Content Security Policy directive: "connect-src 'self' *.princeton.edu http://localhost:* https:".

@sandbergja thanks for catching the extra comma.


sandbergja (approved):
Thanks, @christinach ! 🐬

Issue #4666, opened 2024-12-30
Migrate from jquery-ujs to rails-ujs


christinach:
### What maintenance needs to be done?
In our Ruby on Rails app use the rails-ujs library.

### Level of urgency
- [ ] High
- [ ] Moderate
- [ ] Low

### Why is this maintenance needed?
- Start removing the JQuery dependency https://github.com/pulibrary/orangelight/issues/4584
- We're on Rails 7.1.4. We want to use the UJS that is part of the Rails framework.
- 
### Acceptance criteria
 - [x] [In application.js replace jquery_ujs with rails-ujs. Require it after jquery.](https://github.com/pulibrary/orangelight/pull/4711/files)
 - [x] Replace instances of $.rails.<something> with Rails.<something>
 - [ ] Use new event parameter syntax. For example:
 instead of 
```
$(document).on("ajax:success", function(event, data, status, xhr) {
  ...
});
```
use 
```
$(document).on("ajax:success", function(event) {
  var data = event.detail[0];
  var status = event.detail[1];
  var xhr = event.detail[2];
  ...
});
```
 - [ ] Migrate JQuery events to UJS events (separate ticket)
 - [ ] Correct AJAX type (separate ticket)
 - [ ] Replace JQuery trigger() with native events dispatching: click() method or dispatchEvent() method. (separate ticket)  

### Implementation notes, if any
Reference for this ticket: https://dev.to/nejremeslnici/migrating-from-jquery-ujs-to-rails-ujs-k9m

Assigned @christinach  to complete writing the rest of the tickets that are in the acceptance criteria.

related to #4584

PR #4711, opened 2025-01-16
Remove jquery_ujs from application.js

christinach:
to resolve 'Uncaught Error: If you load both jquery_ujs and rails-ujs, use rails-ujs only.' in the dev console
part of #4666


christinach:
closing in favor of https://github.com/pulibrary/orangelight/pull/4723

PR #4715, opened 2025-01-21
Copy seleced items instead of moving to the top

### LINKED ISSUE #4675: Advanced search combobox: copy, rather than move, selected values to the top
### Current display

When the user selects an option from the combobox, it disappears from its current location and moves to the top of the screen.

### Desired display

When the user selects an option, it keeps its place, but is also copied to the top.  [See this prototype for the behavior.](https://xd.adobe.com/view/d6e1f0fa-704b-4f0f-ac51-32b363ab8c54-83b2/)

### Acceptance criteria

- [x] When the user selects an option, it keeps its place
- [x] When the user selects an option, it is copied to the top
- [x] There is a separation between the selected options at the top and the list of all options below
- [x] When the user de-selects an option, it is removed from the top and keeps its place in the list below
- [x] We do not re-sort items when options are selected
- [x] Rename pul to "All Princeton Holdings"
- [x] When "All Princeton Holdings" option is selected, it is at the top of the selected options.
- [x] PUL holding libraries appear in the menu first _before_ the more specific locations.
- [x] When "All Princeton Holdings" option is not selected, it is at the top of the unselected options.
- [x] I have checked this feature using one of the following browser extensions and confirmed that it does not add any level A or level AA issues:
    * axe dev tools
    * Lighthouse
    * Wave

rladdusaw:
Resolves #4675

### DIFF
--- a/app/assets/stylesheets/components/search--advanced.scss
+++ b/app/assets/stylesheets/components/search--advanced.scss
@@ -100,6 +100,12 @@
   white-space: normal;
 }
 
+.dropdown-item.divider {
+  pointer-events: none;
+  border-bottom: 1px solid black;
+  margin-bottom: 4px;
+}
+
 .row .range_limit input.range_begin,
 .row .range_limit input.range_end {
   width: 5em;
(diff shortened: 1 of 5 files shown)


kevinreiss (approved):
thanks @rladdusaw

christinach (approved):
Thanks @rladdusaw ! 🐈‍⬛

PR #4717, opened 2025-01-22
Upgrade to blacklight 8.8

sandbergja:
Helps with #4025, since this version contains the upstream feature we need.


christinach (approved):
Thanks @sandbergja !

PR #4721, opened 2025-01-23
Queue requests emails with data in a hash

sandbergja:
Then, once it has been queued, turn the hash back into a `Requests::Submission` object.

This helps with #4609, since sidekiq [only allows us to queue emails with data in a very specific list of data types.](https://github.com/sidekiq/sidekiq/wiki/Best-Practices#1-make-your-job-parameters-small-and-simple)


christinach (approved):
Thanks @sandbergja 🍄

maxkadel (approved):
Thanks @sandbergja! 🍄

PR #4723, opened 2025-01-23
[#4666] Remove jquery_ujs from application.js

christinach:
to resolve 'Uncaught Error: If you load both jquery_ujs and rails-ujs, use rails-ujs only.' in the dev console
part of #4666


christinach:
I did a QA test on catalog-staging. It looked ok. Please deploy and review on catalog-staging.

sandbergja (approved):
Thanks, @christinach ! 🐬

Issue #4719, opened 2025-01-23
Limit results by section is not accessible with tabkey navigation


christinach:
### Steps to recreate
Go to https://catalog.princeton.edu/advanced
Use the tabkey to find the Limit results fields : Access, Format, Language, Holding Location 
Press the up or down arrow. The list will not expand.
Press spacebar. The list will expand. Press the Up or Down arrow you cannot select an item from the list. 

### URL
https://catalog.princeton.edu/advanced

### Issues
Cannot use the 'Limit results by' section with tabkey or voiceover

### Severity
High

### WCAG Conformance level (A, AA, AAA)
A
[Success Criterion 2.1.1 Keyboard](https://www.w3.org/TR/WCAG21/#keyboard)

### User impact
Content cannot be operated through a keyboard

### Acceptance criteria
- [ ] I can successfully use this feature with Voice Over in Safari.
- [ ] I have checked this feature using one of the following browser extensions and confirmed that it does not add any level A or level AA issues:
    * axe dev tools
    * Lighthouse
    * Wave
- [ ] ...

maxkadel:
From what I've seen so far, it looks like the keyboard event listeners are not being added correctly. The [keyup listener code](https://github.com/pulibrary/orangelight/blob/6632176ca261510b4b57b52fe07816ddbf7f56c1/app/javascript/orangelight/multiselect_combobox.es6#L63-L77) is never reached, and the [block inside `this.inputElement.addEventListener`](https://github.com/pulibrary/orangelight/blob/6632176ca261510b4b57b52fe07816ddbf7f56c1/app/javascript/orangelight/multiselect_combobox.es6#L57-L58) is never reached.

One thing to note is that there are two `addEventListener` methods, one public and one private. The private one is the one in the class, the public one, which is called on the inputElement, is standard JavaScript.

maxkadel:
In the working commit (38d74429), which is still on Bootstrap 4, we're using the class "form-group", and it is firing the keyup event. We want to look at the javascript from this version of Bootstrap to see how it is querying the item and the class. It's possible it cannot add the event on the element in Bootstrap 5.

See also: [Bootstrap 5 migration notes - dropdown breaking change](https://getbootstrap.com/docs/5.0/migration/#dropdowns).

maxkadel:
[...]

PR #4727, opened 2025-01-29
Allows you to use arrow keys to open a combobox and navigate within it

maxkadel:
- After you hit enter to select an item, trying to navigate with the arrow keys just scrolls the page

Connected to #4719


sandbergja (approved):
Thanks, @maxkadel, @christinach , and @carolyncole  ! 🦄

PR #4730, opened 2025-02-04
[#4025] Enable facet suggest for language and subject topic facets

sandbergja:
Helps with #4025 

I think the only remaining question is if we want to enable this on any other facets.


christinach (approved):
Thanks @sandbergja ! we can enable it as needed

Issue #4731, opened 2025-02-05
Facet suggest: filtering facets, choosing a value, then attempting to filter again does not work


sandbergja:
Thanks to @caroldh for identifying and reporting this issue!

### Expected behavior

You can filter facets, choose a value, then filter again.

### Actual behavior
You can only do it once, and then the UI breaks :-(

### Steps to replicate

1. Search catalog.princeton.edu for economic policy
2. In the Language facet, click "more"
3. Filter the languages to "russ"
4. Select Russian
5. Note that your URL contains `only_values=true` and `query_fragment=russ`
6. In the Language facet, click "more" again
7. Note that you don't get a filter, or really much content in the modal at all!
![The modal contains only the word Russian, an X, and the number 10,935](https://github.com/user-attachments/assets/08124bfb-aaf8-4f96-b98c-3a977267ef01)

### Impact of this bug

Makes it hard to use facets to limit your search

### Implementation notes, if any

* The search results page url should not contain `only_values=true` or `query_fragment=russ` -- these are only meant to be sent in the javascript `fetch`
* See upstream bug https://github.com/projectblacklight/blacklight/issues/3507

maxkadel:
PR against Blacklight main has been merged, now have and [open PR against release-8.x branch](https://github.com/projectblacklight/blacklight/pull/3510)

maxkadel:
Closed by #4736

PR #4736, opened 2025-02-06
Upgrade to latest Blacklight 8.x release, including bugfix for facet suggest

maxkadel:
Connected to #4731


sandbergja (approved):
Thanks, @maxkadel  🦄

PR #4740, opened 2025-02-10
Load_facet checks length of CSV before truncating columns

### LINKED ISSUE #4724: load_facet in browse lists should not proceed if the CSV file is empty

The `load_facet` method in the BrowseLists module should check to make sure that the CSV is a reasonable size before it does its work.

### Level of urgency
- [ ] High
- [x] Moderate
- [ ] Low

This way, if the previous task has failed and produces an empty CSV, we do not end up with a completely empty browse list.

Currently, we can experience the following situation:

- there is an error in, say, the `browse:name_titles` rake task
- the error causes the `/tmp/alma_orangelight_name_titles.csv` file to be totally empty
- `load_facet` doesn't check the file size, and happily `TRUNCATE`s the entire table
- Since the csv file is empty, `load_facet` cannot add any data back into the table
- The production database has an empty browse list :-( 

### Acceptance criteria
 - [ ] `load_facet` does not `TRUNCATE` the current data if the relevant csv file is empty
 - [ ] `load_facet` does not `TRUNCATE` the current data if the relevant csv file is much smaller than it should be

maxkadel:
- Made expected length configurable, based on environment

Closes #4724

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -1200,7 +1200,6 @@ Style/RedundantRegexpEscape:
     - 'app/models/solr_document.rb'
     - 'app/services/faceted_query_service.rb'
     - 'app/services/online_holdings_markup_builder.rb'
-    - 'lib/orangelight/browse_lists.rb'
     - 'lib/orangelight/browse_lists/call_number_csv.rb'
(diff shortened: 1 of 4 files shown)


kevinreiss (approved):

kevinreiss:
Look good, thanks @maxkadel.

PR #4755, opened 2025-02-18
Update bootstrap to 5.3

christinach:
Install gem 'dartsass-sprockets'

Deployed and tested on [catalog-staging](https://catalog-staging.princeton.edu/catalog?search_field=all_fields&q)=


rladdusaw (approved):
Thanks @christinach!

PR #4756, opened 2025-02-19
[#4733] Enable facet suggest for publication facet

### LINKED ISSUE #4733: Add suggest: true for the Place of publication facet
### User story

As an orangelight user, I'd like to limit my results to items published in a specific place.  The place isn't among the most common places in the search, and it is near the end of the alphabet, so the "A-Z Sort" and "Numerical Sort" options aren't helpful.  Instead, I can filter all places of publication to find the one I want.

### Acceptance criteria

- [x] I can click the "More" link in the Place of publication facet, and it allows me to filter facet values to find the one I want.  This is similar to the Language and Subject: Topic facets

You can add `suggest: true` in the catalog controller.

Reported by @caroldh  -- thanks!

christinach:
closes #4733 

Publication facet with suggest feature. It has a 'more link'

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -142,7 +142,7 @@ class CatalogController < ApplicationController
     }, include_in_advanced_search: false
 
     config.add_facet_field 'instrumentation_facet', label: 'Instrumentation', limit: true, include_in_advanced_search: false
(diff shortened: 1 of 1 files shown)


regineheberlein (approved):
Thanks @christinach !

PR #4767, opened 2025-02-28
Remove special conditionals for Internet Explorer and IEMobile

sandbergja:
Internet Explorer and IEMobile have not been supported for years.  We no longer need these Internet Explorer conditional comments.


christinach (approved):
thanks @sandbergja 🍄

PR #4770, opened 2025-03-04
Bookmark all can be activated by spacebar

### LINKED ISSUE #4760: Bookmark All button is not in the tab order
### Steps to recreate

Tab through a search results page.  Note that after the "Sort" and "Per Page" buttons, the tab order goes directly to the first title, skipping "Bookmark All".

Also, the font size is slightly larger on "Bookmark All" than it is on the "Sort" and "per Page" buttons -- not super related to this issue, but it would be nice to fix.

### URL

https://catalog.princeton.edu/catalog?search_field=all_fields&q=stone+circles

### Issues

[SC 2.1.1 Keyboard](https://www.w3.org/TR/WCAG22/#keyboard)

### Severity

High

### WCAG Conformance level (A, AA, AAA)

Level A

### User impact

Keyboard users can't bookmark all efficiently

### Recommended fix

Use a button, rather than the current `div` and hidden checkbox

### Acceptance criteria
- [ ] I can tab to the Bookmark All button and activate it using the keyboard
- [ ] The font size on the Bookmark All is consistent with other nearby buttons
- [ ] I can successfully use the Bookmark All button with Voice Over in Safari.
- [ ] I have checked this button using one of the following browser extensions and confirmed that it does not add any level A or level AA issues:
    * axe dev tools
    * Lighthouse
    * Wave

rladdusaw:
Resolves #4760


app/views/catalog/_bookmark_all.html.erb
```diff
@@ -1,4 +1,4 @@
-<div class="checkbox bookmark_all">
+<div class="checkbox bookmark_all" tabindex='0'>
```
sandbergja, line 1: Could we use a button, rather than a div here?  Then we could listen for a click event to catch either the space bar or the enter key.  It would also get added to the tab order automatically, without having to specify tabindex.

sandbergja (commented):
Thanks, @rladdusaw !  I had one question here.

app/views/catalog/_bookmark_all.html.erb
sandbergja: Thanks for looking into this!  I think the button would be better long term (we wouldn't need to remember the tabindex if we refactor this code in the future, for example), if you can get the CSS sorted.  Another benefit is that the button would be navigable by assistive technologies that allow you to navigate by form control (like NVDA).

app/javascript/orangelight/bookmark_all.es6
```diff
@@ -22,6 +22,16 @@ export default class BookmarkAllManager {
         this.prepopulate_value();
       }
     });
+    $('.bookmark_all').on('keydown', (e) => {
+      e.preventDefault();
```
sandbergja, line 26: This line causes a keyboard trap -- once a user has tabbed to the "Bookmark all" button, they cannot Tab or Shift-Tab to any other element after that!  This is a [WCAG level A violation](https://www.w3.org/WAI/WCAG21/Understanding/no-keyboard-trap.html), and I would argue that it is even more impactful than the original bug.

If we listen for the `click` event instead, we don't have to prevent any defaults or identify the pressed key.  It will provide the spacebar and enter key automatically.

sandbergja (changes requested):
Thanks, @rladdusaw !  This branch has a serious accessibility issue: a keyboard trap -- could you please look into an alternative implementation?  For example, I think a `click`, rather than `keydown` event listener, would be simpler and not introduce this issue.

app/javascript/orangelight/bookmark_all.es6
rladdusaw: Will pressing spacebar or enter trigger the click? I prevented the default because pressing space would scroll the screen down.

sandbergja: [Yes, MDN says](https://developer.mozilla.org/en-US/docs/Web/API/Element/click_event):

> An element receives a click event when any of the following occurs:[...]
> Any user interaction that is equivalent to a click, such as pressing the Space key or Enter key while the element is focused. Note that this only applies to elements with a default key event handler, and therefore, excludes other elements that have been made focusable by setting the [tabindex](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex) attribute.

sandbergja (approved):
Thanks for this great PR, @rladdusaw !  🐈‍⬛

Issue #4765, opened 2025-02-25
Editing an advanced search should result in updated search results


maxkadel:
### Expected behavior
When someone does an advanced search, sees the results, and then edits that search with another keyword and hits search, the original search should be updated with the new keyword.

### Actual behavior
The original search is re-done without any updates.

### Steps to replicate
1. Go to the [search results from an advanced search](https://catalog.princeton.edu/catalog?advanced_type=advanced&boolean_operator1=AND&boolean_operator2=AND&range%5Bpub_date_start_sort%5D%5Bbegin%5D=&range%5Bpub_date_start_sort%5D%5Bend%5D=&clause%5B0%5D%5Bfield%5D=all_fields&clause%5B0%5D%5Bquery%5D=alphabet&boolean_operator1=OR&clause%5B1%5D%5Bfield%5D=author&clause%5B1%5D%5Bquery%5D=azbuka&boolean_operator2=AND&clause%5B2%5D%5Bfield%5D=title&clause%5B2%5D%5Bquery%5D=&f_inclusive%5Badvanced_location_s%5D%5B%5D=rare%24ctsn&f_inclusive%5Badvanced_location_s%5D%5B%5D=rare%24hsvc&advanced_type=advanced&boolean_operator1=AND&boolean_operator2=AND&clause%5B0%5D%5Bfield%5D=all_fields&clause%5B0%5D%5Bquery%5D=alphabet&clause%5B1%5D%5Bfield%5D=author&clause%5B1%5D%5Bquery%5D=&clause%5B2%5D%5Bfield%5D=title&clause%5B2%5D%5Bquery%5D=&f_inclusive%5Badvanced_location_s%5D%5B%5D=rare%24ctsn&f_inclusive%5Badvanced_location_s%5D%5B%5D=rare%24hsvc&f_inclusive%5Badvanced_location_s%5D%5B%5D=rare%24ctsn&f_inclusive%5Badvanced_location_s%5D%5B%5D=rare%24hsvc&range%5Bpub_date_start_sort%5D%5Bbegin%5D=&range%5Bpub_date_start_sort%5D%5Bend%5D=&commit=Search)
2. Click "Edit search"
3. In the second drop-down, select "Author/Creator" and type in "Delaunay" (an author in the search results)
4. Click "Search"
5. Note that the filters at the top of the page and the search results have not updated to reflect the new search term.

### Impact of this bug
E.g. "I can't work until this is fixed" or "I have a workaround"

[...]

PR #4772, opened 2025-03-05
I4765 edit advanced search

maxkadel:
The hidden search state was being duplicated, meaning duplicate parameters were being sent, and only one set of parameters were being used, meaning that editing the advanced search was not working.

Connected to #4765


christinach (approved):
Thank you @maxkadel @sandbergja !

PR #4780, opened 2025-03-07
Fix error when there are no authors in Solr doc

maxkadel:
Was raising [Honeybadger error](https://app.honeybadger.io/projects/54399/faults/118028630) when there was no author

```ruby
NoMethodError: undefined method 'map' for nil

      citation_fields_from_solr[:author_citation_display].map do |author|
```


sandbergja (approved):
Thanks, @maxkadel ! 🦄

Issue #4745, opened 2025-02-12
Block known bots from searches that have a lot of facets


sandbergja:
### What maintenance needs to be done?

Block known bots (i.e. bots with a user agent defined as a bot in the [voight-kampff gem](https://github.com/biola/Voight-Kampff)) from doing searches with 5 or more facets.

### Level of urgency
- [ ] High
- [x] Moderate
- [ ] Low

### Why is this maintenance needed?

Bots DOS us, and often do so using queries with lots of facets.  This would be another tool in our toolkit to keep the site available for humans and well-behaved bots

### Acceptance criteria
 - [ ] When a known bot does a search that includes 5 or more facets, we do not send the request to solr or attempt to render the page

### Implementation notes, if any

* We should check facets applied with the `f` url param and the `f_inclusive` url param
* We can use `agent_is_crawler?` or `request.bot?` to identify if a request came from a known bot

PR #4789, opened 2025-03-10
[#4745] Fix a bug in my code -- need to check for a minimum of 5 facet values before rejecting

sandbergja:
I made a mistake while refactoring this in #4787, and forgot the check for a minimum of 5 facet values.


maxkadel (approved):
Thanks @sandbergja !

PR #4790, opened 2025-03-10
[#4429] Call solr once to render citations

### LINKED ISSUE #4429: Reduce calls to Solr for CiteProc citations
Make it so that displaying citations for SCSB items using CiteProc does not require a second call to Solr. (see https://github.com/pulibrary/orangelight/issues/4012 for original implementation).

### Level of urgency
- [ ] High
- [x] Moderate
- [ ] Low

- Having a second call to Solr slows down the user experience, puts more load on Solr, and should not be necessary (there should be a way to make all fields available for display on the initial call to Solr).

### Acceptance criteria
 - [x] After the show page is loaded, there should not be an additional call to Solr to get the fields needed to display the citation

See `citation_fields_from_solr` method in the Blacklight::Document::CiteProc class. (not yet merged as of this ticket writing).

sandbergja:
Prior to this commit, we called solr twice: once in the controller via Blacklight's `#action_documents` methods (which do not include the correct fields), and again in our `#citation_fields_from_solr` method (which does).

Neither `#action_documents` implementation allows us to pass in specific solr options, so this commit instead brings the `#action_documents` methods' logic into the `#citation` actions.

Also, remove some Blacklight 7 backwards compatibility that is no longer needed.

Closes #4429

### DIFF
--- a/app/controllers/bookmarks_controller.rb
+++ b/app/controllers/bookmarks_controller.rb
@@ -19,6 +19,12 @@ class BookmarksController < CatalogController
     render('orangelight/record_mailer/email_record', formats: [:text])
   end
 
+  def citation
+    bookmarks = token_or_current_or_guest_user.bookmarks
+    bookmark_ids = bookmarks.collect { |bookmark| bookmark.document_id.to_s }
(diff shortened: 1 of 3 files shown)


christinach (approved):
Thanks @sandbergja !

maxkadel (approved):
Yay! Thanks @sandbergja! 🍄

PR #4797, opened 2025-03-11
Use View Components version 3

### LINKED ISSUE #3890: Upgrade view_component to 3.X

Currently, we are using view_component 2.  We should use view_component 3.

### Level of urgency
- [ ] High
- [ ] Moderate
- [x] Low

### EOL upgrade

I didn't see a formal EOL policy from view_component.  However, there hasn't been a release in the 2.X series for almost a year.

So that we don't get too far behind, and so that we can use the [ViewComponent documentation](https://viewcomponent.org/), which is for version 3 

### Acceptance criteria
 - [ ] We use version 3.0.0 or higher of view_component

sandbergja:
closes #3890


christinach (approved):
Thanks @sandbergja !

PR #4798, opened 2025-03-11
Add style to limit button height

### LINKED ISSUE #4768: Use a consistent height for button-like elements on the search results page
### Current display
The "Start over", "Edit search", and search constraint buttons / links are all different heights, adding visual noise and cognitive overhead to the display. 

Image

### Desired display
These three elements should have a consistent height, which is also consistent with the search bar height.

### Acceptance criteria
    * axe dev tools
    * Lighthouse
    * Wave
- [ ] When the page loads, the three elements appear to be the same height.
- [ ] When tabbed through the currently focused element is clear and has sufficient contrast.

rladdusaw:
Resolves #4768


maxkadel (approved):
Thanks @rladdusaw! 🐈‍⬛

PR #4812, opened 2025-03-18
Remove babel

sandbergja:
We were only using it for eslint.  However, the eslint default parser is sufficient for our needs, since we don't use experimental javascript features.

[See the "when should I use @babel/eslint-parser" section](https://www.npmjs.com/package/@babel/eslint-parser)


christinach (approved):
Thanks @sandbergja ! bye bye babel

PR #4816, opened 2025-03-19
bundle update dartsass-sprockets

christinach:
removes bootstrap sass deprecation warnings


sandbergja (approved):
Thanks, @christinach !  🐬   This gets rid of the deprecation warnings for me!

PR #4819, opened 2025-03-20
Move from sneakers to kicks

sandbergja:
Sneakers was [renamed/forked to kicks for the 3.0.0 version](https://github.com/ruby-amqp/kicks/releases/tag/3.0.0).


christinach (approved):
Thanks @sandbergja ! 🍄

Issue #4774, opened 2025-03-05
Question mark in Title search results in false empty result set


regineheberlein:
### Expected behavior
When I search for a title containing a question mark in Advanced Search that PUL has holdings for, I get a non-zero result set.

### Actual behavior
No matches found.

### Steps to [replicate](https://catalog.princeton.edu/catalog?advanced_type=advanced&boolean_operator1=AND&boolean_operator2=AND&range%5Bpub_date_start_sort%5D%5Bbegin%5D=&range%5Bpub_date_start_sort%5D%5Bend%5D=&clause%5B0%5D%5Bfield%5D=all_fields&clause%5B0%5D%5Bquery%5D=Parrish&boolean_operator1=AND&clause%5B1%5D%5Bfield%5D=author&clause%5B1%5D%5Bquery%5D=Trollope%2C+Anthony&boolean_operator2=AND&clause%5B2%5D%5Bfield%5D=title&clause%5B2%5D%5Bquery%5D=Can+You+Forgive+Her%3F&f_inclusive%5Badvanced_location_s%5D%5B%5D=Special+Collections&range%5Bpub_date_start_sort%5D%5Bbegin%5D=&range%5Bpub_date_start_sort%5D%5Bend%5D=&commit=Search)
Go to Advanced Search. Execute a search for
- Keyword: Parrish
- Author: Trollope, Anthony​​​
- Title: Can You Forgive Her?
- Holding Location: Special Collections

Then do the same search, but without the question mark.

### Impact of this bug
False negatives even though we have holdings

kevinreiss:
Possible the "?" is being treated as an operator or isn't encoded correctly.

PR #4820, opened 2025-03-20
Refactor SearchBuilder: extract a general method for mutating all query strings

sandbergja:
Helps with #4774


kevinreiss (approved):

kevinreiss:
Thanks @sandbergja.

PR #4825, opened 2025-03-20
Remove spec support method that is no longer used

sandbergja:


christinach (approved):
Thanks @sandbergja 🍄 !

PR #3386, opened 2023-01-27
Adds vocabulary column to subject table

christinach:
closes https://github.com/pulibrary/orangelight/issues/1982
closes https://github.com/pulibrary/orangelight/issues/1376

update multiple genre vocab
Update the browse_facet task to check for subject facet update browselists load_facet to look for subject facet Include fixtures
Updates subjects.sorted

PR #4826, opened 2025-03-20
Update fixtures with new facet fields

christinach:
Helps with https://github.com/pulibrary/orangelight/pull/3386


spec/system/catalog_show_spec.rb
```diff
@@ -142,7 +142,7 @@
     let(:details_fields) do
       ['printer', 'binder', 'former-owner', 'notes_display', 'binding_note_display',
        'provenance_display', 'references_url_display', 'other_format_display',
-       'other_title_display', 'recap_notes_display']
```
sandbergja, line 145: Why are we removing `recap_notes_display`?

sandbergja (approved):
Thanks, @christinach !  🐬 Looks good, I had a question which is not a blocker.

spec/system/catalog_show_spec.rb
christinach: it is not in the updated fixture

PR #4831, opened 2025-03-21
Remove Blackight 7 compatibility method

sandbergja:
The `Orangelight.using_blacklight7?` method was created to smooth the transition between breaking API changes during our Blacklight 8 migration.  We've been on Blacklight 8 for a while, so this PR removes this method and everything that calls it.


maxkadel (approved):
🎉  Love getting rid of code!

PR #4843, opened 2025-03-27
We moved away from Voyager 4 years ago.

christinach:
We don't use the same item status in orangelight.


maxkadel (approved):
Thanks @christinach! 🐬

PR #4858, opened 2025-04-02
Bundle update 01042025

christinach:


app/models/requests/aeon_url.rb
```diff
@@ -72,11 +72,11 @@ def item_from_document
       end
 
       def at_marquand?
-        holding_location&.fetch('library', nil)&.fetch('code', nil) == 'marquand'
```
sandbergja, line 75: It's cool that rubocop can do this now! 🤖

sandbergja (approved):
Thanks, @christinach ! 🐬

maxkadel (approved):
Thanks @christinach!

PR #4860, opened 2025-04-03
Add a small Solr core for boolean search testing

maxkadel:
It is hard to test complex boolean queries with our current large test corpus, so we create a new, small corpus in its own solr core for complex boolean testing.


sandbergja (approved):
Thanks, @maxkadel !  🦄 😎

PR #4863, opened 2025-04-04
Improve hierarcical facet styles

rladdusaw:


kevinreiss:
Viewing this in staging now looks good in the major browsers and when resized to a small screen.

kevinreiss (approved):

PR #4893, opened 2025-04-30
Remove duplicate visually-hidden from elements

christinach:


maxkadel (approved):
Thanks @christinach! 🐬

PR #4915, opened 2025-05-20
Use small circleci resource class for jobs that need fewer resources

sandbergja:
Still use the large resource class for rspec tests and lighthouse.

This will save around 120 circleci credits per CI run, assuming that the timings are relatively consistent and the [pricing here](https://circleci.com/pricing/price-list/#docker)


regineheberlein (approved):
nice! thanks @sandbergja !

Issue #4909, opened 2025-05-13
JSON-LD should not default to English when the record has "no linguistic content" or similar


sandbergja:
### Expected behavior

When the underlying MARC record has a code 'zxx', 'mul', 'sgn', 'und', or '|||', the JSON-LD should reflect that code.

### Actual behavior

The JSON-LD says English in these cases.

### Steps to replicate

* Go to https://catalog.princeton.edu/catalog/99106517293506421
* Note that the record says that the language is "No linguistic content"
* Go to the [JSON-LD version of the record](https://catalog.princeton.edu/catalog/99106517293506421)
* Note that the language code is "en"

### Impact of this bug

Many records in DPUL erroneously say they are in English.

### Implementation notes

* The language_iana_s field, which we use for the JSON-LD, [specifically maps these languages to "en"](https://github.com/pulibrary/bibdata/blob/6a1e6ac87c98453ef884b52a8fb0737ae44e196f/marc_to_solr/lib/language_service.rb#L9) to provide a best effort fallback language for assistive technologies in the Orangelight UI.

tpendragon:
@sandbergja You can add the 639-3 codes I don't really care, we'll handle it.

maxkadel:
Once https://github.com/pulibrary/bibdata/pull/2786 is merged and a reindex is done with that code in place, we should merge https://github.com/pulibrary/orangelight/pull/4952

PR #4952, opened 2025-06-27
Use new field for json-ld

maxkadel:
This new field still needs to be indexed as of 6/27/2025

Connected to #4909


sandbergja (approved):
Thanks, @maxkadel ! 🦄 We miss you!

PR #4955, opened 2025-06-30
Disable Stackmap

kevinreiss:
Adds a feature to temporarily disable the 'Where to find it' button. Also remove plasma physics from list of stackmap libs as it doesn't exist any longer.


christinach (approved):
thanks @kevinreiss

PR #4967, opened 2025-07-08
Update online options display

rladdusaw:


kevinreiss (commented):
Thanks @rladdusaw. We are exempting most of the components from FeatureEnvy so this one isn't alone 😂. In the future that is probably a bigger re-factor.

Issue #4927, opened 2025-05-28
[Results page] Create a pop-up to prompt users with library accounts to sign-in when they select bookmarks on the Results Page


ellen-aa:
### Current display
Currently, users can select bookmarks, but are not prompted to sign-in, which creates some potential loss of desired search results.

### Desired display
The pop-up will mitigate the potential for loss of search results by prompting users with accounts to sign-in. It will also warn users without accounts that their bookmarks will not be saved. The pop-up will only appear the first time that a user who is not logged in bookmarks an item.

![Image](https://github.com/user-attachments/assets/46aea289-2c83-4b1d-9030-7153ecac735f)

Updated text for the pop-up: 
- "Princeton Users" should be changed to "Library Account Holders". 
- "To save bookmarks for future sessions, please log in using your Princeton account" should be changed to "To save bookmarks for future sessions, please log in using your library account".

Link to the wireframe in Figma: https://www.figma.com/design/jWuYZ9oOrTW39A7CHeHRLh/Princeton-Library?node-id=548-1289&p=f&t=0UJCBMy0tu1tvkB6-0

### Acceptance criteria

- [ ] I can successfully use this feature with Voice Over in Safari. After I press the "bookmarks" button, the focus moves to the popup message and starts reading it. 
- [ ] I have checked this feature using one of the following browser extensions and confirmed that it does not add any level A or level AA issues:
    * axe dev tools
    * Lighthouse
    * Wave
- [ ]

sandbergja:
Closed by #4969

Issue #4910, opened 2025-05-14
[Results page] Adapt the Bookmark feature on the Results Page to the new wireframe


ellen-aa:
### Current display - 

The current display says "Bookmark" with a checkbox. 

![Image](https://github.com/user-attachments/assets/9a8c3286-378e-4eb1-85c3-27188cd8b14e)

### Desired display - 

The desired display will be a Bookmark icon that is clickable.

![Image](https://github.com/user-attachments/assets/07f6dad9-b0dc-4be5-b2c3-b35b2b085b71)

The Bookmark icon will turn orange when clicked. When the user's cursor is over the Bookmark icon, it turns grey.

![Image](https://github.com/user-attachments/assets/2e14f4a7-d4e1-4fbc-b609-a25b2045c99c)

Link to Figma wireframe: https://www.figma.com/design/jWuYZ9oOrTW39A7CHeHRLh/Princeton-Library?node-id=548-1289&p=f&t=0UJCBMy0tu1tvkB6-0

### Acceptance criteria

- [ ] The desired display will be a bookmark icon that is clickable
- [ ] I can successfully use this feature with Voice Over in Safari.
- [ ] I have checked this feature using one of the following browser extensions and confirmed that it does not add any level A or level AA issues:
    * axe dev tools
    * Lighthouse
    * Wave
- [x] We create an .svg image for the bookmark icon, if it does not exist already

sandbergja:
Closed by #4969

PR #4969, opened 2025-07-10
Add a new bookmark button component

sandbergja:
This bookmark button replaces the bookmark checkbox on the search results page.

The first time a non-logged in user presses the button, they are shown a dialog that encourages them to sign in.  The bookmark is added, whether or not they choose to sign in.  We then record the timestamp in localStorage, and do not show them the dialog again as long as that localStorage is present.

Advances #4910
Advances #4927


kevinreiss:
The failing test may need to be moved from a feature test to a system test, so the javascript can be rendered.

Issue #4912, opened 2025-05-14
On search results page, put holding information into cards


maxkadel:
### Current display

![Image](https://github.com/user-attachments/assets/685ed005-4ab9-4b07-b16c-0feff36b1507)

### Desired display

![Image](https://github.com/user-attachments/assets/884ccd62-0e04-477c-9835-40712aff3939)

For when there are more than 3 holding locations:

![Image](https://github.com/user-attachments/assets/bad28dbc-a72d-49a1-abf0-9263dbbe110e)

The cards should be clickable, and they should take you to the Show page

### Acceptance criteria
- [ ] The wording for the availability badges is out of scope for this ticket and will be addressed in another ticket (TICKET NUMBER HERE)
- [ ] The pin icon is used for physical locations
- [ ] I can successfully use this feature with Voice Over in Safari.
- [ ] I have checked this feature using one of the following browser extensions and confirmed that it does not add any level A or level AA issues:
    * axe dev tools
    * Lighthouse
    * Wave
- [ ] ...

### Notes
Requesting designers to add the sub-location back to the holdings - for now, put it on the following line in whatever way looks best to you

PR #4976, opened 2025-07-11
4912 search results lux card

christinach:
part of https://github.com/pulibrary/orangelight/issues/4912


christinach:
An example of 4 lux-cards and a record with one lux-card but long call number content.
An example of 4 lux-cards and a record with one lux-card but long call number content.

PR #4980, opened 2025-07-14
Revert "Status unavailable 4929"

christinach:
Reverts pulibrary/orangelight#4971

We want different display in the record and the show page. For now we will override this in availability.js. 
We will revisit bibdata to refactor the api call response when we will have a better picture of the catalog display.

Friends with https://github.com/pulibrary/bibdata/pull/2798/files


sandbergja (approved):
Thanks, @christinach ! 🐬

Issue #4929, opened 2025-05-28
[Results page] Modify the current availabilty badges on the Results Page


ellen-aa:
### Current display
The current display has 2 badges (Available and Unavailable). The current display also has an On-site Access badge for certain locations.

![Image](https://github.com/user-attachments/assets/06749a93-53b7-4fc3-a82c-a448555e33ca)

![Image](https://github.com/user-attachments/assets/8a856692-606f-4eed-9103-aaf099449bee)

### Desired display
The desired display displays Available, Some Available, and Request. Formerly "On-site Access" items will display as Available. 

![Image](https://github.com/user-attachments/assets/a5d99961-c442-4b45-a43b-a2de4a5951e8)

![Image](https://github.com/user-attachments/assets/c25482bc-5e0b-42e8-b11c-8b926c6329e6)

Link to the [Figma wireframe](https://www.figma.com/design/jWuYZ9oOrTW39A7CHeHRLh/Princeton-Library?node-id=1395-2437&t=Q4CMYbNDTbgrnX9P-1)

### Acceptance criteria

- [ ] I can successfully use this feature with Voice Over in Safari.
- [ ] I have checked this feature using one of the following browser extensions and confirmed that it does not add any level A or level AA issues:
    * axe dev tools
    * Lighthouse
    * Wave
- [ ] ...

christinach:
@ellen-aa @caroldh is there a document with notes on the mapping of the current badges to the new badges? what did we decide for ‘Ask staff’ ? per @kevinreiss [ in slack](https://pulibrary.slack.com/archives/CA891BNR0/p1751922321317049?thread_ts=1751921872.463509&cid=CA891BNR0)

christinach:
@caroldh @ellen-aa one more case: when we have bound-with records we display ‘View record for full availability’. The reason we've been doing this is that bibdata does not know about bound-with records and  we don't get availability information for holdings coming from the host record. For those holdings we ask the user to check the record. What should we display in this case?

mzelesky:
This constituent record has multiple host items: https://catalog.princeton.edu/catalog?search_field=all_fields&q=9997173506421

For a constituent record that is associated with multiple hosts, availability could be different for different items.

christinach:
For the review of this Pr you need also to 'turn on' the feature change_status in bibdata https://bibdata-qa.princeton.edu/features/

christinach:
There are two more cases to handle on the results page: 1. numismatics and 2. theses
[...]

PR #4986, opened 2025-07-15
Coins and Thesis display as Available in the search results page

christinach:
helps with https://github.com/pulibrary/orangelight/issues/4929


regineheberlein (approved):
Thanks @christinach

PR #4987, opened 2025-07-15
Allow icons to overflow to prevent getting cut off

rladdusaw:


sandbergja (approved):
Thanks, @rladdusaw ! 🐈‍⬛

PR #4988, opened 2025-07-15
Fix Bookmark All button

sandbergja:
#4910 changed the DOM structure for the bookmark controls, but did not update the Bookmark all button's selectors accordingly.  This broke the Bookmark all button.

This PR updates those selectors, and also rewrites the Bookmark all javascript from jquery to Vue.


christinach (approved):
Thanks @sandbergja !

PR #4993, opened 2025-07-16
Refactor holding_location_scsb methods into their own view component

sandbergja:
Noting that the new component is named `HoldingAvailabilityScsbComponent` (availability rather than location), since this logic actually renders the availability and has nothing to do with locations.


christinach (approved):
Thanks 🍄 🐝 🥝 🐬  !!!

PR #4995, opened 2025-07-16
Refactor build_holding_notes into a View Component

sandbergja:


christinach (approved):
thanks @sandbergja !!

PR #5006, opened 2025-07-18
[record page] don't include the library name if already present in the location name

sandbergja:
Thanks @ellen-aa for finding this case, where Lewis Library has a location whose name starts with "Lewis Library", meaning that we previously got the redundant label "Lewis Library - Lewis Library - Serials (Off-Site)"


christinach (approved):
Thanks @sandbergja !

Issue #4945, opened 2025-06-11
[Record page] Only display item availability badge when the grouped holdings are collapsed 


rladdusaw:
Ticket https://github.com/pulibrary/orangelight/issues/4933 should be completed before we start working on this one. 

### Current Display
Earlier wireframes had the availability badge always visible.

![Image](https://github.com/user-attachments/assets/fa5a9c8a-402e-4f74-98a6-31196dab52ba)

### Desired Display
The Availability badge is only displayed when the grouped holdings are collapsed

![Image](https://github.com/user-attachments/assets/fee23431-951f-4355-8186-49c533c33720)

Link to wireframe in Figma: https://www.figma.com/design/jWuYZ9oOrTW39A7CHeHRLh/Princeton-Library?node-id=548-1289&p=f&t=ezkxr7Koz5sQ8yRU-0

- [ ] The badges calculated for the collapsed holdings should take the same business logic that we use for holdings display on the search results page. 
- [ ] When all holdings in a grouped location are marked as "Available" we display a label of "Available" with a green background. Holdings marked for "On-Site Access" should be treated the same as "Available Locations" for this display logic. 
- [ ] When all holdings in a grouped location are "Unavailable" we should display "Request" with a grey background. Holdings labeled "Ask Staff" should also be included in this display logic. 
- [ ] When 1 or more holdings in a grouped location is either marked as "Unavailable" and other holdings in that grouped location are marked as "Available" or "Some Available" we should display "Some Available with a green background. 
- [ ] When 1 or more holdings are judged to be unavailable 
- [ ] I can successfully use this feature with Voice Over in Safari.
- [ ] I have checked this feature using one of the following browser extensions and confirmed that it does not add any level A or level AA issues:
    * axe dev tools
    * Lighthouse
    * Wave

sandbergja:
Closed by #4945

PR #5008, opened 2025-07-18
[#4945] Display a holdings groups availability badge if the group is collapsed

sandbergja:
Advances #4945


spec/features/thesis_spec.rb
```diff
@@ -19,10 +19,6 @@
       within 'dd.blacklight-restrictions_note_display' do
         find_link 'Mudd Manuscript Library'
       end
-      find('span', text: 'Mudd Manuscript Library')
```
sandbergja, line 22: These assertions were failing (there are now multiple spans that contain the library name), and were not very meaningful, so I removed them.

PR #5011, opened 2025-07-19
Update available green color

christinach:
partof #4912


sandbergja (approved):
Thanks, @christinach !

PR #5017, opened 2025-07-21
Fix several issues with sytles on results page

### LINKED ISSUE #5014: Issue 4912 has some visual differences from the wireframe
### User story
The application should look like the wireframe

### Acceptance criteria

- [ ] The location has the pin icon
- [ ] There is not excessive white space in the cards
- [ ] The green color used is the same as in the wireframe
- [ ] The color of the call numbers is gray
- [ ] The "see 3 locations" text is centered and not at the bottom of the card
- [ ] the bookmark icon and thumbnail are aligned
    * axe dev tools
    * Lighthouse
    * Wave
- [ ] ...

rladdusaw:
Resolves #5014

### DIFF
--- a/app/assets/stylesheets/components/availability.scss
+++ b/app/assets/stylesheets/components/availability.scss
@@ -3,7 +3,8 @@
 @import "variables/typography";
 
 .holdings-card .lux-card .lux-text-style {
-  padding: 0 1rem 1rem 0;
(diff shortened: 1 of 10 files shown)


sandbergja (approved):
Thanks, @rladdusaw ! 🐈‍⬛

PR #5020, opened 2025-07-22
Don't show the text Location has

sandbergja:
This makes it compatible with the figma

Before:
holding group on the show page shows a science call number, call no browse, the text location has, then information about which journal volumes are at Forrestal

After:

all data is the same except there is no text location has under call no browse


regineheberlein (approved):
Thanks @sandbergja !

Issue #5021, opened 2025-07-22
Incorporate UI changes identified in July 22nd standup


sandbergja:
### Acceptance criteria

- [ ] Within a search result availability card, the status should be bottom-justified, and always a consistent distance from the bottom of the card
- [ ] Within a search result availability card, the location and availability information need to pop a bit more (try reducing the font size of the call number)
- [ ] Within a search result availability card, the map icon and the library name should always be on the same name
- [ ] Search result availability cards should be a consistent width
- [ ] Within a search result availability card, the text Some Available needs to be green
- [ ] Within a search result availability card, the text Some Available needs to be on one line on typical screen widths
- [ ] Rather than saying "See 3 locations", the final card should say "See 3 more locations"
- [ ] There needs to be a bit more space between the format icon/book jacket and the bookmark button.
- [ ] The format icon/book jacket and the bookmark button should be right-aligned
- [ ] Adjust the physical badge icon
- [ ] The physical badge font is a bit hard to read.  Try a darker color and/or heavier font weight
- [ ] Adjust the availability card hover and focus styles.  The whole card should appear activatable, not just some underlined text within it

PR #5023, opened 2025-07-22
5021.2 call num reduction

RyanAJensen:
Resolves second checkbox of #5021


sandbergja (approved):
Thanks, @RyanAJensen ! 🥝

PR #5027, opened 2025-07-22
5021.5 some available green

RyanAJensen:
Resolves fifth checkbox of #5021


sandbergja:
@RyanAJensen I've been doing the following to check my work locally using the data from catalog-qa:

- Make sure you are on VPN
- In one terminal window, `ssh -L 1234:localhost:8983 deploy@lib-solr-staging5d.princeton.edu`
- In another terminal window start rails like so: `SOLR_URL=http://localhost:1234/solr/catalog-qa be rails s`

Then if you go to http://localhost:3000 and search for "science", you should find an example of "Some items not available"

PR #5028, opened 2025-07-22
Improve holding card focus styles

sandbergja:
Rather than styling the text with an underline, we now style the entire card with a lux-style focus indicator and a light grey background.

Part of #5021


christinach (approved):
thanks @sandbergja !

Issue #5031, opened 2025-07-23
Incorporate UI changes identified in July 23rd standup


sandbergja:
#### Search result page
- [x] If there is one holding for a record, expand horizontally.  Else make them a consistent width - Jane
- [x] Reduce top and bottom spacing in each holding card - Jane
- [x] For online holdings, the globe icon does not have enough spacing - Jane

#### Record show page

- [x] Holdings information should be gray (i.e. information underneath "Call no. browse"). See example:

<img width="1306" height="331" alt="Image" src="https://github.com/user-attachments/assets/f2aaf787-5598-469a-886c-5628d4ce694c" />

ellen-aa:
A new solution to the color of the holdings information was decided and implemented on 07/23/2025.

PR #5032, opened 2025-07-23
Holdings cards: consistent width unless there is a single holding

sandbergja:
Per today's standup, if there is a single holding, we allow it to expand so that it does not need to wrap any text.  If there are multiple holdings, we keep them a consistent width.

Part of #5031


kevinreiss (approved):
Thanks @sandbergja.

Issue #5038, opened 2025-07-23
Remove redundant globe icon from Online Options component


sandbergja:
### Current display

The Online Options vue component includes a globe icon

### Desired display

We should remove the globe icon per @ellen-aa and @caroldh , since it is redundant in the search results view (there is already a globe icon in a badge to indicate that the title is an online title)

PR #5044, opened 2025-07-23
Remove extra globe icon

kevinreiss:
For issue #5038.

### DIFF
--- a/app/javascript/orangelight/vue_components/online_options.vue
+++ b/app/javascript/orangelight/vue_components/online_options.vue
@@ -1,8 +1,6 @@
 <template>
   <div v-if="count > 1" class="online-wrapper">
-    <LuxHyperlink :href="url"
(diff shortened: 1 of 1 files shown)


app/javascript/orangelight/vue_components/online_options.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <div v-if="count > 1" class="online-wrapper">
     <LuxHyperlink :href="url"
-      ><span class="icon icon-web d-inline"></span>{{ text }}</LuxHyperlink
+      ><span class="icon d-inline"></span>{{ text }}</LuxHyperlink
```
sandbergja, line 4: Can we remove the entire span while we're at it?

sandbergja (approved):
Thanks, @kevinreiss !  I had a small question which is not a blocker/

PR #5051, opened 2025-07-24
5033 adjust search

### LINKED ISSUE #5033: Mobile View UI Adjustments Search Results Page
### Current display
Single holding display
Image

Multiple Holding locations
Image

### Desired display
Both single and multiple holding location cards are displayed as a single vertical list when displayed on a mobile screen. Each data point on the card should have it's own line. 

### Acceptance criteria

- [ ] Location Cards are displayed in a vertical list
- [ ] When there are multiple locations all cards have the same width. 
    * axe dev tools
    * Lighthouse
    * Wave

RyanAJensen:
Resolves #5033

### DIFF
--- a/app/assets/stylesheets/components/availability.scss
+++ b/app/assets/stylesheets/components/availability.scss
@@ -12,7 +12,7 @@
 .holdings-card {
   display: flex;
   gap: 0.5rem;
-  justify-content: space-between;
+  justify-content: flex-start;
   flex-direction: row;
   border-radius: 0.375rem;
   padding: 0.5rem 0rem 0.5rem 0rem;
@@ -500,7 +500,6 @@
 @media screen and (max-width: $bp-large) {
   .holdings-card {
     display: grid;
-    justify-content: start;
     width: 100% !important;
     grid-template-rows: repeat(1, 1fr);
   }


app/assets/stylesheets/components/availability.scss
```diff
@@ -12,6 +12,7 @@
 .holdings-card {
   display: flex;
   gap: 0.5rem;
+  justify-content: space-between;
```
sandbergja, line 15: Could we add a `justify-content: start` within the media query below too?  Otherwise, on a wide screen, we get this appearance, where the holding cards are far from one another:

on a record with two holdings, the first holding is at the start of the screen, then there is a large gap, and then the second holding is right next to the cover image

RyanAJensen: Is that the correct fix?

sandbergja: Ah, sorry, I meant `flex-start`, not `start`.  I played around with it a bit locally, and I think we can change this line to `justify-content: flex-start`, and then remove the `justify-content` within the media query below.  How does that look for you?

RyanAJensen: Sounds good to me, I made the changes you suggested

PR #5076, opened 2025-07-28
Fix spacing in 'Copies in library' box so location and call number don't appear so far apart

### LINKED ISSUE #5069: [Record page] Location font size should be larger in accordian
### Current display
Font in Location appears small in relation to the box. Feedback from Beth: It looks like there is something in the font-family within summary that is adjusting the font-size to be smaller. Also - could be that the box is just too large?

Image

### Desired display
Text is more in proportion to size of box.

Image

### Acceptance criteria

    * axe dev tools
    * Lighthouse
    * Wave
- [ ] ...

RyanAJensen:
Resolves #5069

### DIFF
--- a/app/assets/stylesheets/components/details.scss
+++ b/app/assets/stylesheets/components/details.scss
@@ -35,8 +35,8 @@ details:not([open]) summary .when-open {
     display: none;
 }
 
-details>* {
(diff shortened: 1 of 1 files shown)


sandbergja (approved):
Thanks, @RyanAJensen ! 🥝

Issue #5074, opened 2025-07-28
[Record page] when the holding location label is blank remove the `-`


christinach:
### Expected behavior
Hyphen should not display when there holding location label is blank

### Actual behavior
Hyphen displays when the holding location label is blank.

### Steps to replicate
see https://catalog-qa.princeton.edu/catalog/99131517408106421

<img width="1055" height="250" alt="screenshot of the holding location section in the record page" src="https://github.com/user-attachments/assets/b219fb48-1ad8-4732-bc1b-c85da8e9011c" />

PR #5077, opened 2025-07-28
Don't display a hyphen in holding group names if location name is empty

sandbergja:
Previously, the location eastasian appeared as 'East Asian Library -'.  With this commit, it simply appears on the show page as 'East Asian Library'.

Resolves https://github.com/pulibrary/orangelight/issues/5074


christinach (approved):
Thanks @sandbergja

PR #5078, opened 2025-07-28
Add rounded corners to 'Other versions' sidebar box

### LINKED ISSUE #5068: [Record page] Round corners for Other Versions sidebar box
### Current display

Other Versions sidebar box has square corners:

Image

### Desired display

Sidebar box with information on Other Versions should have rounded corners to match the other elements with rounded corners on Record page.
This was not included on original Figma board.

### Acceptance criteria

    * axe dev tools
    * Lighthouse
    * Wave
- [ ] ...

RyanAJensen:
Resolves #5068

### DIFF
--- a/app/assets/stylesheets/components/availability.scss
+++ b/app/assets/stylesheets/components/availability.scss
@@ -416,7 +416,7 @@
 }
 
 .availability--other-versions {
-  border-radius: var(--font-size-x-small);
+  border-radius: var(--border-radius-pill);
   .linked-block {
     margin-bottom: .75em;
   }


app/assets/stylesheets/components/availability.scss
```diff
@@ -416,6 +416,7 @@
 }
 
 .availability--other-versions {
+  border-radius: var(--font-size-x-small);
```
sandbergja, line 419: What would you think about using
```suggestion
  border-radius: var(--border-radius-pill);
```

instead?  It's slightly more rounded (16px vs 12px), but I think it would be nice to not tie this box's border-radius to the font size (I think we'd find it unexpected in the future if we were to change the font size, and it has an effect on the box corner rounding).

sandbergja (commented):
Thanks, @RyanAJensen !  It looks good, I just had a question and an alternative variable I'd like your feedback on.

app/assets/stylesheets/components/availability.scss
RyanAJensen: Yes that sounds like a good change. I was set on 12px for some reason because I think one of the other rounded boxes had that border-radius. I was definitely iffy on using a font size variable for the border-radius but I didn't know where to find more appropriately-suited variables. Thanks for the fix!

PR #5079, opened 2025-07-28
Refactor PhysicalHoldingsMarkupBuilder to be compatible with frozen string literals

sandbergja:
Also, add a comment describing the class to fix the last remaining reek issue on this class.


christinach (approved):
Thanks @sandbergja !

Issue #5062, opened 2025-07-25
Return to Record Page from Request Form should open the same Location that the form displayed


kevinreiss:
### Current display
When one follows the a request button on a record with multiple holding locations to look at options available for a specific holding location and you "return" that page after looking at the request form the record page displays it's "default" state which is to have the holding listed first expanded with the others contracted. If you've followed the request button from a location that is collapsed by default this is confusing. You'd expect to see the one you last selected opened. 

### Desired display
When returning from the request form the record should display the holding involved in that request as expanded. 

### Example
1. Visit a record with [multiple holdings](https://catalog-qa.princeton.edu/catalog/9952684423506421)
2. Expand the second location at Recap and select the [request button](https://catalog-qa.princeton.edu/requests/9952684423506421?aeon=false&mfhd=22694545730006421). 
3. From the record page select "Return to Record"
4. Observe that the first location "Marquand" is expanded rather than the one you selected "ReCAP."

### Acceptance criteria

- [ ] When navigating from the request form to the record page the holding location that was viewed on the form should be expanded. 
- [ ] I can successfully use this feature with Voice Over in Safari.
- [ ] I have checked this feature using one of the following browser extensions and confirmed that it does not add any level A or level AA issues:
    * axe dev tools
    * Lighthouse
    * Wave

sandbergja:
Closed by #5087

PR #5081, opened 2025-07-28
Show page: open holding group on page load if specified in the open_holdings param

sandbergja:
Helps with #5062


christinach (approved):
Thanks @sandbergja !

PR #5087, opened 2025-07-28
Keep holding group open when user returns from the request form

### LINKED ISSUE #5062: Return to Record Page from Request Form should open the same Location that the form displayed

sandbergja:
Before this commit, when a user opened a holding group on the show page, pressed the Request button, then pressed the Return to Record button, the holding group was no longer open.

This commit adds a param to the requests form that allows this Return button to take the user back to the show page with the desired groups open.

Resolves #5062

### DIFF
--- a/app/components/holdings/location_services_component.rb
+++ b/app/components/holdings/location_services_component.rb
@@ -3,17 +3,19 @@
 # which services are available for a holding, given
 # its location (typically some type of Request button)
 # :reek:TooManyMethods
+# :reek:TooManyInstanceVariables
 class Holdings::LocationServicesComponent < ViewComponent::Base
-  def initialize(adapter, holding_id, location_rules, holding)
+  def initialize(adapter, holding_id, location_rules, holding, open_holdings = nil)
     @adapter = adapter
     @holding_id = holding_id
     @location_rules = location_rules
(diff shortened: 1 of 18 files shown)


christinach (approved):
thanks @sandbergja !

Issue #5061, opened 2025-07-24
[Record Page] SCSB Unavailable Item Displays no status on Record Page


kevinreiss:
### Expected behavior
I expect to see status (i.e. Available/Unavailable) displayed in the Copies in Library section of the record page. 

### Actual behavior
Status isn't displayed sometimes. 

<img width="1058" height="239" alt="Image" src="https://github.com/user-attachments/assets/193d6f9b-712f-4ad3-b8fe-3a2f157ca86f" />

### Steps to replicate
Visit [record](https://catalog-qa.princeton.edu/catalog/SCSB-10763824) in QA. 

### Impact of this bug
Confusing display

### Notes
I checked the [availability response ]([url](https://bibdata-qa.princeton.edu/availability?scsb_id=10763824))from bibdata-qa for the ID and it seems correct.

PR #5091, opened 2025-07-29
Update availability.es6 to add bg-danger when status is unavailable

christinach:
fixes[#5061]

Update availability.es6 to add bg-danger when status is unavailable


sandbergja (approved):
Thanks, @christinach 🐬

PR #5099, opened 2025-07-30
Remove Stackmap health check

sandbergja:
We no longer use Stackmaps, so we don't need to know its health status


christinach (approved):
🍄

PR #5125, opened 2025-07-31
Adjust Spacing in Copies in Library

### LINKED ISSUE #5097: [Record Page] Mobile View - Adjust Spacing in Copies in Library
### Current display
In most standard mobile Iphone and Android sizings the element flow in this table is bit odd and not aligned. 

Image

### Desired display
To my eye we should make this display more proportionally and symmetrically where all three major elements are at the same level as much as we can. 

### Acceptance criteria

- [ ] Spacing in the "Copies in Library" table aligns better. 
    * axe dev tools
    * Lighthouse
    * Wave

RyanAJensen:
Resolves #5097

### DIFF
--- a/app/assets/stylesheets/components/record.scss
+++ b/app/assets/stylesheets/components/record.scss
@@ -207,6 +207,8 @@
     .availability-table td {
       width: 100%;
       flex: auto;
+      justify-content: flex-start;
+      padding: 10px 15px 10px 0;
     }
 
     .availability-table thead {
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/components/record.scss
```diff
@@ -207,6 +207,8 @@
     .availability-table td {
       width: 100%;
       flex: auto;
+      justify-content: left;
```
christinach, line 210: @rladdusaw left is not a valid value. Can you change it to `justify-content: flex-start`? thanks

christinach (changes requested):
@RyanAJensen one comment

app/assets/stylesheets/components/record.scss
RyanAJensen: Yes I can fix that, sorry about that!

PR #5134, opened 2025-07-31
Address deprecation warning for render_document_partials method

sandbergja:
This method will be removed in Blacklight 9.  We should probably eventually move from these partials to our own interpretation of a `Blacklight::DocumentComponent` or something similar, but this at least gets us past the deprecation warning noise.


christinach (approved):
thanks @sandbergja

PR #5135, opened 2025-07-31
Stop rendering the Blacklight constraints view partial

sandbergja:
It will be removed in Blacklight 9, see https://github.com/projectblacklight/blacklight/pull/3331


christinach (approved):
thanks @sandbergja !

PR #5146, opened 2025-08-04
Create individual methods to set the status display on the

christinach:
availability element
Extract them in a new default class


sandbergja (approved):
Really nice refactor!  Thanks, @christinach ! 🐬

PR #5150, opened 2025-08-05
Additional status display methods

christinach:
Replace the relevant ones in availability.es6


sandbergja (approved):
Thanks, @christinach ! 🐬

PR #5154, opened 2025-08-05
Refactor: move numismatics-specific search building logic into the NumismaticsFormSearchBuilder

sandbergja:
Prior to this commit, this logic was in the general `SearchBuilder` class, but only used for the numismatics form.

This commit moves the logic to `NumismaticsFormSearchBuilder`, allowing us to simplify the numismatics action in the `CatalogController`, remove some Reek ignores, and make the general `SearchBuilder` class a bit smaller and more approachable.


regineheberlein (commented):
thanks @sandbergja !

Issue #4634, opened 2024-12-12
Enable Blacklight to Discover Translated Works by the Source Language of Translations


minjiec:
### User story
Researchers, educators, and translators need to locate translated works, specifically by the source language from which the work is translated.

For example, a researcher may want to survey publications on philosophy translated from German into English in year 2023;
An instructor may want to locate titles (in a specific genre such as fiction or juvenile fiction) translated from Japanese into English, so that they can make a recommended reading list for students who are learning Japanese language and culture;
A translator may want to harvest a list of queer novels translated from French, or a list of poetry translated from Persian, or a list of children's literature about the Holocaust translated from Hebrew, to examine "biases and gaps in the representation of a specific discipline, country or community" ([Fricker 2024](https://www.linkedin.com/pulse/librarians-oclc-clarivate-discover-translation-christophe-fricker-mmf3e/)).
#### Current barrier
No library discovery system currently supports this kind of queries, even though catalogers have diligently entered the information in bib records, in the form of `041 1# $h` (source language of a translated resource). There is no way to search the data.
#### Reference librarians’ imperfect solutions
In order to assist users’ need for locating translated works, reference librarians occasionally offer search tips. Because none is leveraging the most precise information embedded in 041$h, those methods are not only cumbersome but tend to yield incomplete, outdated, and inaccurate results. For example–

- [Harvard](https://ask.library.harvard.edu/faq/82549): How do I find all existing translations of a book? - Ask a Librarian

- [NYU](https://guides.nyu.edu/translations): How to Find Translations into and from English - Research Guides at New York University

- [University of Richmond](https://libguides.richmond.edu/c.php?g=41853&p=8252233): Identifying Translations

#### More user stories shared by selectors and CaMS members

The proposed feature would be particularly useful when I answer reference questions about translations from languages I am less familiar with. I may not know the print reference sources of those languages as I would with Chinese, for example. (_Martin Heijdra, Director of the East Asian Library_) 

The [UNESCO Index Translationum](https://www.unesco.org/xtrans/bsform.aspx), which I often relied on for finding translations, is woefully outdated in contents, technical standards, and search functions. I often have students asking me for secondary sources by Chinese scholars in English.  A search feature like this would be a great help for finding these. (_Joshua Seufert, Chinese Studies Librarian_) 

[...]

PR #5155, opened 2025-08-05
Add the original language of translation facet

sandbergja:
We do not display it yet, since we have not done a full re-index.

This is an incremental step toward #4634


regineheberlein (approved):
thanks @sandbergja !

PR #5159, opened 2025-08-06
Add a new field: Source language of translation, behind a feature flipper

sandbergja:
This field displays as a dropdown on the advanced search form. It only displays if you have enabled the relevant
FlipFlop feature -- we will not want to do this in production until we have done a full reindex.

If you want to try this locally, you will need to first run: `bundle exec rake cache:clear`.

Advances #4634


christinach (approved):
Thanks @sandbergja !

PR #5161, opened 2025-08-07
Display the translation source language on the show page

sandbergja:
Helps with #4634

It's okay to start displaying this on the show page before our full reindex.  It will simply add some more information to the show page, but will not affect the advanced search page or displayed facet counts.


christinach (approved):

christinach (approved):
Thanks @sandbergja !

PR #5162, opened 2025-08-07
Source language of translation: show all on the advanced search screen

sandbergja:
Prior to this commit, only the 10 most popular source languages displayed on the advanced search form.

This commit makes this more general: any facet field configured to display on the advanced search screen will not limit its values.

Helps with #4634


christinach (approved):
Thanks @sandbergja !

PR #5168, opened 2025-08-08
Add orange highlight around box on hover, reformat html to facilitate change

### LINKED ISSUE #5148: [Results page] Make the entire box around Online resources highlight in orange when a user hovers over it
### Current display
Currently, only the title of the online resource in the online links card is clickable (and it highlights in orange when a user hovers over it).
Image

### Desired display
The desired display would not highlight the text in the online card. Instead the entire online links card would highlight in orange when a user hovers over it, as we do for the cards for physical holdings. See example here:

Image

### Acceptance criteria

    * axe dev tools
    * Lighthouse
    * Wave
- [ ] ...

### Notes

This will further clarify and standardize what is clickable and what is not on PUL's results page.

RyanAJensen:
Resolves #5148

### DIFF
--- a/app/assets/stylesheets/components/availability.scss
+++ b/app/assets/stylesheets/components/availability.scss
@@ -10,7 +10,7 @@
   font-size: 0.75rem;
 }
 
-.holdings-card {
+.holdings-card, .online-holdings-list {
   display: flex;
   gap: 0.5rem;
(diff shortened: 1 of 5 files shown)


rladdusaw (approved):
Thanks @RyanAJensen!

Issue #5167, opened 2025-08-08
[Results page] Adjust the location name for items that are In Library Use Only


ellen-aa:
### Current display
Users at branch libraries report that, "if a patron sees only the Mendel location with the call number below, it seems like they could get it from the stacks here in Mendel. It's only when you click to the record page that you see it's in ReCAP." This can create confusion for users who are not anticipating having to request and wait for a title to arrive at the location.

See full Slack message from Lisa Read: https://pulibrary.slack.com/archives/CA0UY0Q07/p1754404520681089 

### Desired display

The desired display is to change the location names for In Library Use Only items for the results page only. The changes can be found on the Google sheet at this link: https://docs.google.com/spreadsheets/d/1Lzksnwcw9YgXsHENSBHjIWS--SLdU0PxTxmZiZzAAuo/edit?usp=sharing. 

Beth German has kindly whipped up this wireframe for us:

<img width="720" height="571" alt="wireframe of desired display" src="https://github.com/user-attachments/assets/a8d1b6b4-41b1-46f1-aebb-f2fca61de756" />

### Acceptance criteria

- [ ] I can successfully use this feature with Voice Over in Safari.
- [ ] I have checked this feature using one of the following browser extensions and confirmed that it does not add any level A or level AA issues:
    * axe dev tools
    * Lighthouse
    * Wave
- [ ] ...

### Notes

This is a sudden priority since it will impact the branch libraries and the influx of users coming back to campus after the summer break.

christinach:
@ellen-aa @caroldh there are more location that are remote in recap. To see all of them check [Bibdata holding locations](https://bibdata.princeton.edu/locations/holding_locations) and search for 'recap_rmt' in 'Remote Storage' column.

PR #5169, opened 2025-08-08
Only in the search results, display specific label for in library use locations

christinach:
Only in the search results, display specific label for in library use locations.

'arch$pw': "Archictecture (Remote Storage)",
'eastasian$pl': "East Asian (Remote Storage)",
'engineer$pt': "Engineering (Remote Storage)",
'firestone$pb': "Firestone (Remote Storage)",
'firestone$pf': "Firestone (Remote Storage)",
'lewis$pn': "Lewis (Remote Storage)",
'lewis$ps': "Lewis (Remote Storage)",
'mendel$pk': "Mendel (Remote Storage)",
'stokes$pm': "Stokes (Remote Storage)"

part of #5167

I also need to change the javascript . Will do in a second PR

### DIFF
--- a/app/components/holdings/search_location_component.rb
+++ b/app/components/holdings/search_location_component.rb
@@ -28,7 +28,7 @@ class Holdings::SearchLocationComponent < ViewComponent::Base
           'mendel$pk': "Mendel (Remote Storage)",
           'stokes$pm': "Stokes (Remote Storage)"
         }.freeze
-        library_in_use_locations[location_code.to_sym] if library_in_use_locations.key?(location_code.to_sym)
+        library_in_use_locations[location_code.to_sym]
       end
 
       def call_number
--- a/app/components/holdings/search_location_component.rb
+++ b/app/components/holdings/search_location_component.rb
@@ -8,13 +8,30 @@ class Holdings::SearchLocationComponent < ViewComponent::Base
 
     private
 
+      attr_reader :holding_hash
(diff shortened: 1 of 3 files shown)


app/components/holdings/search_location_component.rb
```diff
@@ -8,13 +8,30 @@ def initialize(holding_hash)
...
+          'eastasian$pl': "East Asian (Remote Storage)",
+          'engineer$pt': "Engineering (Remote Storage)",
+          'firestone$pb': "Firestone (Remote Storage)",
+          'firestone$pf': "Firestone (Remote Storage)",
+          'lewis$pn': "Lewis (Remote Storage)",
+          'lewis$ps': "Lewis (Remote Storage)",
+          'mendel$pk': "Mendel (Remote Storage)",
+          'stokes$pm': "Stokes (Remote Storage)"
+        }.freeze
+        library_in_use_locations[location_code.to_sym] if library_in_use_locations.key?(location_code.to_sym)
```
sandbergja, line 31: Could this be simplified to not use the `if`?  I think it would have the same behavior (i.e. returning `nil` if the hash does not contain the key)

```suggestion
        library_in_use_locations[location_code.to_sym]
```

sandbergja (approved):
Thanks, @christinach ! 🐬  Looks great, I just had a small idea that is not a blocker.

PR #5171, opened 2025-08-12
Proxy links and don't display finding aids as online

rladdusaw:


sandbergja (approved):
Thanks, @rladdusaw !  🐈‍⬛

PR #5177, opened 2025-08-13
mendel$qk and eastasian$ql should display a Remote storage label

christinach:
eastasian$ql: 'East Asian (Remote Storage)',
mendel$qk: 'Mendel (Remote Storage)',


sandbergja (approved):
Thanks, @christinach ! 🐬

Issue #5176, opened 2025-08-13
Aeon titles with more than 500 items display as Unavailable on the requests form, even when they are available


sandbergja:
### Expected behavior

If there are more than 500 items, we do not show an availability badge on the requests form

### Actual behavior

We show an "Unavailable" badge on the requests form for titles that are only requestable via aeon, even if all 500+ items are available.

### Steps to replicate

1. Go to [this record in the catalog](https://catalog.princeton.edu/catalog/9930682063506421)
2. Press the request button
3. Note that it says "Unavailable"

### Impact of this bug

Users might think that they cannot access items even though they absolutely can.

### Implementation notes, if any

This happens because of the following sequence of events:
* In `Requests::Form`, `too_many_items?` returns true
* As a result, `load_items` does an early return and does not load any items
* As a result, `build_requestable` takes the path `build_requestable_from_data`, which calls `build_requestable_from_holding`, which creates a new `Requests::Requestable` without specifying any item data.
* `Requests::Requestable#initialize` notices that there is no item, and supplies a NullItem instead
* `NullItem` implements `#status` to return the string `'Unavailable'`

@christinach and I looked at this issue, and confirmed that we don't want to change `NullItem` since it is used in other places as well.  We could instead use a different Item class polymorphically for this scenario that does not provide an Unavailable status.

PR #5178, opened 2025-08-13
Refactor: Requestable class should accept a Requests::Item, rather than construct a new one

sandbergja:
This will simplify the constructor slightly and, more importantly, make `Requestable` more flexible to accept other classes similar to `Requests::Item` and `Requests::NullItem`.

Helps with #5176


christinach (approved):
thanks @sandbergja !

rladdusaw (approved):
Thanks @sandbergja

PR #5183, opened 2025-08-14
Search results page: fix text overflow on Chrome

### LINKED ISSUE #5166: [Results page] The location name for Special Collections items extends beyond the card
### Expected behavior
As a user, I would expect the location information to be contained in the card.

### Actual behavior
The location name is bleeding into the next card, both when a user hovers over it and when they are not.

screenshot of search results page with actual display

screenshot of search results page with actual display

### Steps to replicate

See examples above.

### Impact of this bug

The display isn't great for SC items.

```
```

sandbergja:
On Chrome, long library names like Special Collections extended past the edge of the card, thanks to a bootstrap CSS rule for all direct descendants of .row that set this element's width to 100% (which, combined with 30px of offset on the left caused the issue).

Firefox was unaffected.

Closes #5166

Marking as a draft until @ellen-aa and @caroldh have an opportunity to review it on catalog-qa

### DIFF
--- a/app/assets/stylesheets/components/availability.scss
+++ b/app/assets/stylesheets/components/availability.scss
@@ -46,6 +46,7 @@
   .search-result-library-name {
     margin-left: 6px;
     padding: 0;
+    width: calc(100% - 30px);
   }
 }
 


christinach (approved):
Thanks @sandbergja !

PR #5189, opened 2025-08-19
Remove unused locals from render partial call

sandbergja:
We also already are storing the result of `@request.default_pick_ups` in a local called `default_pick_ups`, there is no need to call `@request.default_pick_ups again`.


christinach (approved):
Thanks @sandbergja 1

PR #5191, opened 2025-08-19
Refactor requestable_form partial into a view component

sandbergja:
This partial was pretty much a long if-elsif block.  This commit refactors it into a view component, where the logic is a bit easier to see and it is a bit easier to test (although it still requires so many mocks!)


kevinreiss (approved):
Thanks @sandbergja 🍄

PR #5195, opened 2025-08-19
Replace !empty? with any?

sandbergja:
Removing the double negative makes it less confusing for me.


kevinreiss (approved):
Thanks @sandbergja.

Issue #4886, opened 2025-04-29
Display Viewer for Indexed Figgy Ephemera Items. 


kevinreiss:
### User story
As a user I would like to follow a link in the Online section of a record page and then view the Ephemera item content as I would a Figgy Item linked to a MMS ID value. 

### Acceptance criteria

- [ ] We process the electronic_holdings_display1 value with a ephemera UUID into a request that enables the rendering fo the viewer and the updating the online content. 
- [ ] The online section clearly points the user to the content. 
- [ ] The viewer with content is rendered at the bottom of the page. 
- [ ] I can successfully use this feature with Voice Over in Safari.
- [ ] I have checked this feature using one of the following browser extensions and confirmed that it does not add any level A or level AA issues:
    * axe dev tools
    * Lighthouse
    * Wave

### Concrete example
See [example JSONLD data](https://figgy-staging.princeton.edu/catalog/33fc03db-3bca-4388-84d3-6b48092199d6.jsonld).
The key will be the @ID value in the jsonld. Example data: { "https://figgy-staging.princeton.edu/catalog/33fc03db-3bca-4388-84d3-6b48092199d6":  ["Online Content" : "Born Digital Monographs, Serials, & Series Reports"] }

### Implementation notes, if any
A iiif manifest url can also be constructed using the id value. Example: https://figgy-staging.princeton.edu/concern/ephemera_folders/af4a941d-96a4-463e-9043-cfa512e5eddd/manifest. See [worksheet](https://docs.google.com/spreadsheets/d/1UlfW-sb125tnL62kJFQtKCgzz9TuUkT3NFJKlZaLC7c/edit?gid=2039455207#gid=2039455207) more mapping details.

kevinreiss:
1. One thing to consider is loading the uv in OL rather than waiting to load it from Figgy. 
2. There is a unique config in Figgy, that we would need to replicate.

PR #5197, opened 2025-08-25
Build a viewer_component to render the iiif viewer for the ephemera resources

christinach:
related to #4886


rladdusaw (approved):
Thanks @christinach!

Issue #3900, opened 2023-12-11
Make All Marquand (Non-Rare Book) Items Always requestable regardless of Alma Status


kevinreiss:
### User story
As a user I can obtain Marquand items even if they are charged since all are stored onsite, so I should not be able to make a resource sharing request for these items. 

### Acceptance criteria

- [x] All non-rare marquand items (except those that have a base status of Item Not In Place and are not "On Order" or "In Process") .should ALWAYS be eligible for [Clancy](https://github.com/pulibrary/orangelight/blob/main/app/models/requests/service_eligibility/clancy_in_library.rb), [Clancy_unavailable](https://github.com/pulibrary/orangelight/blob/main/app/models/requests/service_eligibility/clancy_unavailable.rb), and [Marquand In Library Requests](https://github.com/pulibrary/orangelight/blob/main/app/models/requests/service_eligibility/marquand_in_library.rb). However if the item has an unavailable status instead of creating a hold request in Alma, Clancy or SCSB (depending on the item location) ~~the system should instead just send an email to Marquand staff notifying them a user wants the item~~ Marquand requested we display a message asking users to email their marquand shared email address rather than allowing the form to actually be sumbitted. . 
- [x] All non-rare marquand items except those that have a base status of Item Not In Place and are not "On Order" or "In Process") should ALWAYS be eligible for [Clancy EDD](https://github.com/pulibrary/orangelight/blob/main/app/models/requests/service_eligibility/clancy_edd.rb) and [Marquand EDD](https://github.com/pulibrary/orangelight/blob/main/app/models/requests/service_eligibility/marquand_edd.rb). However if the item has an unavailable status instead of creating a hold request in Alma, Clancy or SCSB (depending on the item location) the system should instead just send an email to Marquand staff notifying them a user wants a portion of the item digitized. 
- [x]  ~~The email the system sends to Marquand Staff should have the subject line of "Paging Request for Marquand Item in Carrel"~~ Marquand decided they didn't want this. 
- [ ] I can successfully use this feature with Voice Over in Safari.
- [ ] I have checked this feature using one of the following browser extensions and confirmed that it does not add any level A or level AA issues:
    * axe dev tools
    * Lighthouse
    * Wave

### Concrete example
After implementing this ticket: Items in Alma that are located in Marquand, not in a SC location, have a base status of Item Not In Place and are not "On Order" or "In Process".  Behavior for "On Order" and "In Process" items will not change as part of this ticket.

[...]

PR #5200, opened 2025-08-26
When Marquand title is checked out, present a way to email Marquand on the request form

sandbergja:
Advances #3900, although it does not fully automate it since Marquand is in the middle of its large move.

Here is how it looks on a form with multiple holdings: checked out volumes do not get a select checkbox, and the status "Item in Use, Ask Staff for access" is displayed in yellow.  In the Delivery Options column, there is a link to email marquand for access:

the form with 2 checked out marquand items and 1 checked in one, as described above


kevinreiss (approved):
Thanks @sandbergja.

PR #5203, opened 2025-08-27
Refine display of Marquand Contact info on the requests form

sandbergja:
Follow up to #5200, from Feedback from Marquand.

* If the item is on loan, say "Item in use"
* If the item has any other process type, or no process type, say "Item unavailable"


christinach (approved):
Thanks @sandbergja !

PR #5213, opened 2025-09-02
Update thumbnail_display method to read the text value from the solr document.

christinach:
Update thumbnail_display method to read the text value from the solr document.
We changed solconf to store thumbnail_display as a text not an array


rladdusaw (approved):
Thanks @christinach!

PR #5214, opened 2025-09-08
Don't generate an external icon when there is an empty external label

christinach:
Don't display an external icon only, for cases where the external link is indexed but there is no label. 
I've seen this in production but couldn't find an example quickly.

I'm adding this to cover cases from ephemera resources with the addition in https://github.com/pulibrary/bibdata/pull/2917/ and also cases where for any reason we may not have a label for an external link url. The link_text in ElectronicAccess is indexed as an empty string and not a None so it returns an [""] for a label.
```
"electronic_access_1display":"{\"\":[\"\"],\"https://catalog-staging.princeton.edu/catalog/f4e58e9f-623d-41aa-9a8c-0919f3f7d4ed#view\":[\"Digital content\"],\"iiif_manifest_paths\":{\"ephemera_ark\":\"https://figgy.princeton.edu/concern/ephemera_folders/f4e58e9f-623d-41aa-9a8c-0919f3f7d4ed/manifest\"}}"
```

### DIFF
--- a/app/services/online_holdings_markup_builder.rb
+++ b/app/services/online_holdings_markup_builder.rb
@@ -33,7 +33,7 @@ class OnlineHoldingsMarkupBuilder < HoldingRequestsBuilder
                  link_to(texts.first, ::Regexp.last_match(0), class: 'electronic-access-link')
                end
              else
-               link_text = new_tab_icon(texts.first)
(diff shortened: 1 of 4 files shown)


christinach:
closing for now

PR #5222, opened 2025-09-24
Allow standard circulating location materials to be picked up at multiple branches

christinach:
Remove check for pending?
It does not make sense as a check anymore

related to https://github.com/pulibrary/bibdata/issues/2926

Friends with https://github.com/pulibrary/bibdata/pull/2963


christinach:
We will deploy both PRs on Monday morning October 20th

Issue #5221, opened 2025-09-19
Catalog HTML Email Pick-up Request Email is Currently Malformed


kevinreiss:
### Expected behavior
When a user receives an email telling them an pick-up request for an on-campus item has been successfully placed the text for the message should be visible in the email message. It should be part of the template and display below the page header.

### Actual behavior
Here is what the message looks like in outlook and when you try and open the HTML message in a browser:

<img width="1560" height="859" alt="Image" src="https://github.com/user-attachments/assets/bf6e353d-f723-49fd-a21f-1b13dbfdad57" />

You can see in the screenshot that the actual email text seems to be not properly placed in the HTML. 

### Steps to replicate

1. Find an item in Firestone Library available for request
2. Request the item.
3. Review the email generated in your email client. 

### Impact of this bug
Confusing user experience. Users think the email is broken. 

### Implementation notes, if any

The message should be after header but above the pick-up location. This only seems to happen for on campus materials not recap items. Recap request confirmation mails are formatted properly. See the example below from a recap item:

<img width="1195" height="550" alt="Image" src="https://github.com/user-attachments/assets/7b6202be-97b6-49d3-ab67-8f10efe79492" />

christinach:
resolved by https://github.com/pulibrary/orangelight/pull/5227

PR #5227, opened 2025-10-01
Wrap class body-content in a <tr><td> with the same styling as the rest of the templates

christinach:
related to #5221


christinach:
With this change:

screenshot with fixed hetml template confirmation

PR #5229, opened 2025-10-06
Update honeybadger.yml

christinach:
exclude RuntimeError: Sorry, the catalog does not support file uploads


christinach:
closing in favor of https://github.com/pulibrary/orangelight/pull/5228/files

Issue #5234, opened 2025-10-10
Every Firestone On the Shelf Paging Email and Patron Confirmation message display the wrong Pickup Location


kevinreiss:
### Expected behavior
The pick-up location displayed in this email should reflect the Firestone location that is appropriate. 

### Actual behavior
Right now every Firestone request regardless of location displays "Firestone Library - PF" in the email sent to circulation staff. 

<img width="796" height="328" alt="Image" src="https://github.com/user-attachments/assets/d77efd4c-6fca-4f5b-be58-ec7bef3ba38b" />

### Steps to replicate

Find an item in Firestone that is available for pick-up. Request the item. Consult the On the Shelf Paging Request email that goes to circ staff and the confirmation email that patrons receive. Both of the emails have this issue. I tried with MMS ID: 9911500553506421. 

### Impact of this bug
Doesn't seem to be impacting actual requests, but the display is confusing for staff. 

### Implementation notes, if any
Staff said this started after 9/16 and is seen on every request for Firestone materials. In this example the correct Pick-up location to display would be Firestone Library, PA instead of Firestone Library, PF.

christinach:
I'm seeing the issue in the email that is send for Partner items as well. 

<img width="987" height="766" alt="Partners item request email template" src="https://github.com/user-attachments/assets/c316443d-337c-44ff-9e7e-7aae2f8e192a" />

kevinreiss:
The recap item example should display the "QX" value.

PR #5243, opened 2025-10-20
Update logic of building pickup locations

christinach:
Illiad eligible items should be picked up from Firestone Library with gfa_pickup of PA
Before soring default_pick_ups filter them based on the location code: If it's firestone$pf use delivery location gfa_pickup for any other location including firestone (not$pf) reject delivery location with gfa_pickup PF

related to #5234


kevinreiss (approved):
Thanks @christinach

PR #5246, opened 2025-10-21
Move ctx generation out of the Requests::Form#initialize

sandbergja:
This initializer already does a lot of work -- creating a `ctx` does not need to happen in the initializer, so let's move it out.


christinach (approved):
thanks @sandbergja !

PR #5252, opened 2025-10-22
Refactor SCSB availability check into its own class

sandbergja:
This was formerly known as barcode_sort (though it did much more than sorting them).

### DIFF
--- a/app/models/requests/recap_item_availability.rb
+++ b/app/models/requests/recap_item_availability.rb
@@ -1,0 +1,71 @@
+# frozen_string_literal: true
+module Requests
+  # This class is responsible for finding availability data from SCSB for items in ReCAP (whether owned by PUL or a partner)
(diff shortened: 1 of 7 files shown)


christinach (approved):
Thanks @sandbergja 🍄 !

app/models/requests/recap_item_availability.rb
```diff
@@ -0,0 +1,71 @@
...
+
+    # This inner class is responsible for adding availability data that has already been fetch from SCSB to an item hash
+    class ItemWithAvailability
+      def initialize(availability_data:, item:)
+        @availability_data = availability_data
+        @item = item
+      end
+
+      def call
+        # Not sure why we set the string and symbol versions to different values...
```
christinach: I think because we need to do some work in bibdata first and bring the status label that we want.

PR #5254, opened 2025-10-23
Refactor: move Requestable creation out of Requests::Form

sandbergja:
Prior to this commit, one of the `Form` class' many responsibilities was to create many of the `Requestable` instances needed by the Requests system. It created four different kinds of `Requestable`, and it was hard to trace the logic throughout the Form class for each one of those four paths.

This commit takes a different approach: introducing a factory class named `RequestablesList`, which determines what type of `Requestable` to create, then delegates the creation to one of four smaller factories: `AlmaItemsFactory`, `ScsbItemsFactory`, `NoItemsFactory`, and `TooManyItemsFactory`.

### DIFF
--- a/app/models/requests/requestables_list/alma_items_factory.rb
+++ b/app/models/requests/requestables_list/alma_items_factory.rb
@@ -34,7 +34,11 @@ module Requests
           end
 
           def location_code
-            location['code'] if location
+            location_object.code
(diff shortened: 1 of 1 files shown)


christinach (approved):
Thanks @sandbergja !

Issue #5236, opened 2025-10-15
[Requests] Remove method no_services?


christinach:
### What work needs to be done?
Remove method [no_services](https://github.com/pulibrary/orangelight/blob/00f264088060b1b22bbd3fc83f84805de64f5766/app/models/requests/requestable_decorator.rb#L151)?

### Why is this work needed?
- We need to remove this method and clear the codebase. It was added during Covid era and it has no meaning to keep it

### Acceptance criteria
 - [ ] Tests are updated and passing.

PR #5260, opened 2025-10-23
Remove no_services?

christinach:
This was added during Covid when the library was closed.

related to #5236


sandbergja (approved):
Thanks, @christinach ! 🐬

Issue #4410, opened 2024-10-02
eastasian$ql Holding Location should allow request to the delivery locations that are configured for this location code in bibdata


kevinreiss:
### Expected behavior
All active public delivery locations for a holding location should display on the request form. The pickup locations are configured in bibdata. Eastasian$ql current has two [public pickup locations](https://bibdata.princeton.edu/locations/holding_locations/eastasian$ql) configured that fit this description and 5 staff locations.

### Actual behavior
See this [example](https://catalog.princeton.edu/requests/9934002193506421?aeon=false&mfhd=22669804680006421) record from the location. 

<img width="1445" alt="Screenshot 2024-10-02 at 5 10 03 PM" src="https://github.com/user-attachments/assets/58ba23cf-9d05-4201-b210-edd45cc4a656">

### Steps to replicate
1. Pull up any [eastasian$ql item available](https://catalog.princeton.edu/requests/9934002193506421?aeon=false&mfhd=22669804680006421) for loan. 
2. Open the request form. Only East Asian Library is available as a choice. 

### Impact of this bug
Users cannot send items in this location to the Firsetone Library when they want to. 

### Acceptance Criteria
- [ ] Items in eastasian$ql have both Firestone Library and East Asian Library as pick-up choices and the Staff locations

### Notes
See: [_requestable_delivery_option_in_library_use.html.erb](https://github.com/pulibrary/orangelight/blob/main/app/views/requests/form/_requestable_delivery_option_in_library_use.html.erb#L15)
[requestable_decorator.rb](https://github.com/pulibrary/orangelight/blob/main/app/models/requests/requestable_decorator.rb#L137).

christinach:
@kevinreiss [Eastasian$ql](https://bibdata.princeton.edu/locations/holding_locations/eastasian$ql) has pickup locations: true the following: 
[Technical Services 693](https://bibdata.princeton.edu/locations/delivery_locations/2)
[East Asian Library](https://bibdata.princeton.edu/locations/delivery_locations/4)
[Firestone Library](https://bibdata.princeton.edu/locations/delivery_locations/6)
[Firestone Library, Resource Sharing](https://bibdata.princeton.edu/locations/delivery_locations/12)
[Technical Services HMT](https://bibdata.princeton.edu/locations/delivery_locations/23)

which attribute defines that it's a public pickup location?Is it `GFA Pickup` PL and PF or is it the `staff_only` that we are looking to say that a location is a public pickup?

[...]

PR #5262, opened 2025-10-23
Recap confirmed that the following gfa_pickups

christinach:
PF, QL, QP, QC, QT are allowed for this location
Include them in valid_recap_annex_pickup method

related to #4410


rladdusaw (approved):
Thanks @christinach!

PR #5268, opened 2025-10-27
Improve performance of online_holding_note?

sandbergja:
* Rename method for clarity
* Memoize the parsed JSON so that it does not need to be parsed again for every field in the catalog controller (this is the main performance gain)
* Symbolize names to reduce memory allocations
* Use `any?` with a block, rather than `collect` then `any`, so we don't need to iterate through every holding if we get a match before that
* Check in a microbenchmark
* Upgrade the json gem to a newer version that has some performance optimizations

Microbenchmark before:
```
$ bundle exec ruby benchmarks/app/controllers/concerns/orangelight/catalog.rb
ruby 3.4.1 (2024-12-25 revision 48d4efcb85) +YJIT +PRISM [arm64-darwin24]
Warming up --------------------------------------
                        36.000 i/100ms
Calculating -------------------------------------
                        363.356 (± 1.4%) i/s    (2.75 ms/i) -      1.836k in   5.053892s
23428 objects allocated
```

Microbenchmark after:
```
Warming up --------------------------------------
                       374.839k i/100ms
Calculating -------------------------------------
                          3.749M (± 0.7%) i/s  (266.76 ns/i) -     18.742M in   4.999956s
2 objects allocated
```

* More iterations/second is good!
* Fewer allocations is good!

The user-visible effect is a slight increase in speed on the show page for titles with many holdings.  In my local dev setup, this method used to take ~10% of all time spent displaying the show page for the journal "Science".  With this commit, it takes ~1% of all time.

### DIFF
--- a/spec/controllers/catalog_controller_spec.rb
+++ b/spec/controllers/catalog_controller_spec.rb
@@ -39,8 +39,8 @@ RSpec.describe CatalogController do
       expect(email).to be_nil
     end
   end
-  describe '#online_holding_note?' do
-    subject(:note) { described_class.new.online_holding_note?(nil, document) }
+  describe '#show_location_has?' do
+    subject(:note) { described_class.new.show_location_has?(nil, document) }
 
     let(:link) { 'field not blank' }
     let(:holdings) do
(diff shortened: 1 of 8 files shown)


spec/controllers/catalog_controller_spec.rb
```diff
@@ -39,30 +39,30 @@
...
+    subject(:note) { described_class.new.show_location_has?(nil, document) }
 
     let(:link) { 'field not blank' }
     let(:holdings) do
       '{"1":{"location_has":["blah"]}}'
     end
 
     describe 'for document with link and holding note' do
-      let(:document) { { electronic_access_1display: link, holdings_1display: holdings } }
+      let(:document) { SolrDocument.new({ electronic_access_1display: link, holdings_1display: holdings }) }
```
sandbergja: Use a `SolrDocument` in this test rather than a hash, since it is more realistic.

PR #5274, opened 2025-10-28
Add hover effect for the LibMap Button

kevinreiss:


christinach (approved):
Thanks @kevinreiss !
I ran axe tools and the color contrast looks ok . I forced the state to be on hover and it passes again the test.

PR #5290, opened 2025-10-30
Replace more vcr with stub_catalog_raw

sandbergja:


christinach (approved):
Thanks @sandbergja 🍄 🍄

PR #5294, opened 2025-10-30
Return correct message when there is no valid service

### LINKED ISSUE #5220: Affliate and Affliate-P Patrons who Authenticate Via CAS Receive don't see Proper Message in Request Form
### Expected behavior
When any patron who signs in via CAS/SSO in the patron groups of Affliate and Affliate-P tries to request an item that they cannot request for pickup they should see a message that indicates the item is not available for request. 

### Actual behavior
When patrons of this type tries to request items not requestable to them they see the following message

Image

Users should see the following message stored in this [key](https://github.com/pulibrary/orangelight/blob/cbfc926ac694ab29573dc1768ec73c91b66a2bf2/config/locales/requests.en.yml#L237C19-L237C100) in the requests translation file. Instead they see "Translation missing: en.requests.brief_msg".

### Steps to replicate
1. Obtain credentials to an Alma External account that is in either the Affliate or Affliate-P patron group. The easiest way to do this is have your own account changed to one of these patron groups. 
2. Find an item that is not requestable to patrons of this type. [Example record](https://catalog.princeton.edu/catalog/99125539872406421). 
3. Click on the request button. You'll see the incorrect message an each item attached to this holding location.

### Impact of this bug
Confusing user experience. 

This ONLY happens for patrons in these groups that sign in via CAS. Patrons who sign in with internal Alma credentials see the correct message.

rladdusaw:
Resolves #5220


kevinreiss (approved):
thanks @rladdusaw

PR #5300, opened 2025-10-31
Remove unused views

sandbergja:
The Requests::FormController#index action simply redirects to the catalog home page, so there is no need for it to have its own view.


christinach (approved):
Thanks @sandbergja

PR #5304, opened 2025-10-31
Simplify Requestable constructor a little

sandbergja:
This way, we don't have to create an instance of the Illiad class until it's needed.


christinach (approved):
thanks @sandbergja 🍄

PR #5306, opened 2025-10-31
Electronic delivery mode - activate the request button 

christinach:
Electronic delivery mode 

* In electronic delivery mode activate the request button
when the user has added a title in the input field
and the checkbox is selected.
Otherwise deactivate it.

* Update existing validation-error class to use --color-red variable from lux
* Add client side error when title is missing


christinach:
https://github.com/user-attachments/assets/6b4e2272-bcf7-4544-a53e-a2e03a545b08

Issue #4183, opened 2024-09-03
Annex Requests for Holdings that Have no Items Should notify the Annex Email. 


kevinreiss:
### What maintenance needs to be done?
When an Annex holding that has no item data is requested the annex staff should receive an email notification. Currently this scenario notifies "recapproblems@princeton.edu". 

### Level of urgency
- [ ] High
- [ ] Moderate
- [ ] Low

### Why is this maintenance needed?
- Annex staff are just forwarded the email from the minders of recapproblems. 

### Acceptance criteria
 - [ ] When an Annex holding that has no item data is requested the message should go to forranx@princeton.edu

### Honeybadger link and code snippet, if applicable
Alma error is ```Hold can not be created, message: Invalid Response.```. 

### Implementation notes, if any
[Need an example item]

kevinreiss:
Here is another example record: https://catalog.princeton.edu/catalog/9941642973506421.

christinach:
@kevinreiss which message we want to use?
```
 annex:
      email: "forranx@princeton.edu"
      email_subject: "Annex Request"
      brief_msg: "Item offsite at Forrestal Annex. Requests for pick-up typically take 1-2 business days to process."
      # temporary changes issue 438
      #email_conf_msg: "You will be notified via email when your item is ready for pick-up. Most items are ready in 1 to 2 business days."
      email_conf_msg: "Your request for this item has been received. Once we complete our processing, you'll receive an email stating that the book is available for pick-up."
      in_library_use_msg_html: After notification of your item's arrival at %{location_label}, you should visit the Library to view it
    annex_in_library:
      email: "forranx@princeton.edu"
      email_subject: "Annex In Library Use Request"
      brief_msg: "Item offsite at Forrestal Annex. Requests for viewing in a library typically take 1-2 business days to process."
      email_conf_msg: "Your request for this item has been received. Once we complete our processing, you'll receive an email stating that the book is available for pick-up."
```

annex or annex_in_library? 
Or do we want to add a new message for this case and what would that be?

kevinreiss:
@christinach this example is for Firestone https://catalog.princeton.edu/requests/993083506421?aeon=false&mfhd=22740191170006421&open_holdings=Firestone+Library+-+Stacks. I'm confused on that one. This ticket is for annex items. Right now messages in this state go only to recapproblems, they wanted to add the annex email to that notification.

christinach:
[...]

PR #5339, opened 2025-11-05
Annex requests with no items should email forranx@princeton.edu

christinach:
Add new confirmation and email templates when
there are no annex items - add new RequestMailer methods to support the templates
email this case to forranx@princeton.edu

Update request_mailer_preview.rb
 * remove stale and unused methods
 * parsing of location label Update mock data in request_mailer_preview.rb spec

related to #4183


kevinreiss (approved):
Thanks @christinach. I tested this with both physical delivery and electronic delivery requests in all three Annex scenarios. (1) a record that has no items at all, (2) a record that has some items but also displays the fill-in form, and (2) an standard Annex record with an item record without the fill-in form.

PR #5344, opened 2025-11-07
Revise patron group service descriptions

kevinreiss:
Updated descriptions for various patron services and clarified service types available to all users.


regineheberlein (approved):
thanks @kevinreiss !

PR #5357, opened 2025-11-12
Remove unused conditional branch

sandbergja:
In Javascript, `[] === []` does _not_ evaluate to true. This is because arrays are a type of Object in JS, and [Objects are considered equal only if they reference the same symbol](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Equality_comparisons_and_sameness#loose_equality_using).

Therefore, this branch never was reached.


christinach (approved):
Thanks @sandbergja 🍄

Issue #5314, opened 2025-11-03
Replace jQuery with vanilla Js es6 in orangelight_ui_loader


christinach:
## Current state
[orangelight ui loader](https://github.com/pulibrary/orangelight/blob/main/app/javascript/orangelight/orangelight_ui_loader.es6) uses JQuery. We want to move away from this dependency.

## Future state
Replace all JQuery usage with JS es6 features.

## Acceptance criteria

- [ ] New code is sufficiently tested
- [ ] Old code is replaced 
- [ ] Check bundle size of js code to confirm it's size and avoid regressions.

related to #4584

PR #5365, opened 2025-11-14
Remove non-working focus management for modals

sandbergja:
This code attempted to automatically focus the first input in a modal when a modal was opened.  However, the event is never triggered in practice, so this focus does not actually happen in any of the modal forms I tried (SMS, Email, Ask a question, or Suggest a correction).

Let's remove this non-working code.

Helps with #5314 by removing jQuery code


christinach (approved):
thanks @sandbergja !

Issue #5310, opened 2025-11-03
Replace jquery in book_covers.es6


christinach:
## Current state
[book cover es6](https://github.com/pulibrary/orangelight/blob/main/app/javascript/orangelight/book_covers.es6) uses JQuery. We want to move away from this dependency.

## Future state
Replace all JQuery usage with JS es6 features.

## Acceptance criteria

- [ ] New code is sufficiently tested
- [ ] Old code is replaced 
- [ ] Check bundle size of js code to confirm it's size and avoid regressions.

related to #4584

PR #5383, opened 2025-11-18
Implement our own JSON-P requests

sandbergja:
jQuery's `$.getJSON` function can request JSON using the [JSON-P strategy](https://en.wikipedia.org/wiki/JSONP), which is needed for certain legacy APIs like Google Books.

Since we'd like to move away from jQuery, and since Google Books does not offer the same functionality (requesting multiple ISBNs and OCLC numbers at once) in their more modern APIs, we implement our own JSON-P

Helps with #5310


kevinreiss (approved):
Thanks @sandbergja.

Issue #5400, opened 2025-11-21
Record page with host_id should return successfully the availability for all the holdings


christinach:
### Expected behavior
Record page of constituent record that belongs to a host, has an html host id and renders the availability for both the constituent and the host successfully.

### Actual behavior
Record page of constituent record that belongs to a host, has an html host id attribute and does not render the availability for any of the holdings.

### Steps to replicate
See the following [catalog record 993594213506421](https://catalog.princeton.edu/catalog/993594213506421). The availability status does not display.

### Impact of this bug
Patrons cannot view/read the availability.

### Notes
The record page has a host_id. the ruby method that returns this host_id returns an array. The HTML data attribute translates this to a string.
When the host_id exists on the record page we add it in the bibdata availability url to get the availability for the host and the constituent. In this case the bibdata url is: https://bibdata.princeton.edu/bibliographic/availability.json?deep=true&bib_ids=993594213506421,[%2299125489791206421%22]

PR #5401, opened 2025-11-21
[Record page]-Constituent record with host_id should render the availability 

christinach:
closes https://github.com/pulibrary/orangelight/issues/5400

In the show page we have in main content a
data-host-id attribute that returns a JSON formatted array string

This is building an erroneous bibdata availability url with constituent id and host id represented as an array. As a result this url will fail the fetch

With this commit we use JSON.parse and convert it to an array so that we can join the ids to a string

### DIFF
--- a/app/javascript/orangelight/availability.es6
+++ b/app/javascript/orangelight/availability.es6
@@ -69,9 +69,17 @@ export default class AvailabilityUpdater {
   request_show_page_availability(allowRetry) {
     this.id = window.location.pathname.split('/').pop();
     const mainContent = document.getElementById('main-content');
-    this.host_id = mainContent
+    const hostIdAttr = mainContent
       ? mainContent.getAttribute('data-host-id') || ''
       : '';
+
+    if (hostIdAttr && hostIdAttr.startsWith('[') && hostIdAttr.endsWith(']')) {
+      const parsed = JSON.parse(hostIdAttr);
+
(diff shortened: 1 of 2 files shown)


app/javascript/orangelight/availability.es6
```diff
@@ -69,10 +69,18 @@ export default class AvailabilityUpdater {
   request_show_page_availability(allowRetry) {
     this.id = window.location.pathname.split('/').pop();
     const mainContent = document.getElementById('main-content');
-    this.host_id = mainContent
+    const hostIdAttr = mainContent
       ? mainContent.getAttribute('data-host-id') || ''
       : '';
 
+    if (hostIdAttr && hostIdAttr.startsWith('[') && hostIdAttr.endsWith(']')) {
```
christinach: If it looks like a JSON array then parse it

app/javascript/orangelight/availability.es6
```diff
@@ -269,7 +281,11 @@ export default class AvailabilityUpdater {
     // problematic availability response behavior for constituent record page with host records.
     // It treats host records as holdings of the constituent record. see: https://github.com/pulibrary/bibdata/issues/1739
     if (this.host_id.length > 0) {
-      this.host_id.forEach((mms_id) => {
+      // Handle both string and array cases for host_id
```
christinach: Handle both string and array cases for host_id

kevinreiss (approved):

PR #5479, opened 2026-01-05
Update Gemfile to use rails version 8.1 instead of 8.1.1

christinach:


sandbergja (approved):
Thanks, @christinach 🐬

PR #5497, opened 2026-01-05
Use View Component version 4

sandbergja:
Using the unreleased github main branch, since https://github.com/ViewComponent/view_component/pull/2426 has not yet been released and several tests fail without it.


christinach (approved):
thanks @sandbergja

Issue #5495, opened 2026-01-05
Items in Resource Sharing Temporary Location Should Display as "Unavailable" on the Search Results Page


kevinreiss:
## Description
Items that are in the temporary location "RES_SHARE$IN_RS_REQ" have been charged out to partner resource sharing are not readily available to PUL users. 

## Issue
When you find one of these items on the search results page it displays green as "available" rather than grey as "Request"

Example searches
https://catalog.princeton.edu/catalog?search_field=all_fields&q=99131390231506421
https://catalog.princeton.edu/catalog?search_field=all_fields&q=99131686597006421

## Notes
Two Example records in this location: 

https://catalog.princeton.edu/catalog/99131686597006421
https://catalog.princeton.edu/catalog/99131390231506421

Looking at the [response](https://catalog.princeton.edu/catalog?search_field=all_fields&q=99131390231506421) for one of these items for availability from bibdata it does appear we mark the item as "unavailable" there so there is likely something with the current javascript processing this response that isn't seeing that. 

I don't believe this an issue related to a recently changed status, both of these examples were charged out several months ago. There are approximately 51,000 items in this locations so this a fairly impactful discrepancy to resolve.

christinach:
resolved by https://github.com/pulibrary/orangelight/pull/5499

PR #5499, opened 2026-01-06
Items in temporary location RES_SHARE$IN_RS_REQ are Unavailable

christinach:
Items in temporary location RES_SHARE$IN_RS_REQ are Unavailable
The status display in the search results page displays `Request` (gray color)
while in the record page `Unavailable` (red color)

We check if it's a temp location using the temp_location attribute from the bibdata response. We keep a fallback to check the bibdata response holding_id.

related to [#5495]

### DIFF
--- a/app/javascript/orangelight/availability_search_results.js
+++ b/app/javascript/orangelight/availability_search_results.js
@@ -11,7 +11,6 @@
       'marquand$ref',
       'marquand$ph',
       'marquand$fesrf',
-      'RES_SHARE$IN_RS_REQ' 
 */
 import AvailabilityBase from './availability_base.js';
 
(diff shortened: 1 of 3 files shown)


app/javascript/orangelight/availability_search_results.js
```diff
@@ -72,10 +73,29 @@ export default class AvailabilitySearchResults extends AvailabilityBase {
...
     });
-
     return true;
   }
 
+  #updateTemporaryLocationAvailability(record_id, holding_id) {
+    // We assume that items in temp locations are always available
+    // except RES_SHARE$IN_RS_REQ which is unavailable.
+    const availability_display_element = document.querySelector(
+      `*[data-availability-record='true'][data-record-id='${record_id}'][data-holding-id='${holding_id}'] span.lux-text-style`
```
christinach: the holding_id is coming from the bibdata response. For temporary locations it is the string of the location code unless the temporary location is `RES_SHARE$IN_RS_REQ`. In the case of the RES_SHARE$IN_RS_REQ temporary location the holding id is the actual permanent holding.

Issue #1065, opened 2017-12-18
Support Staff View for SCSB Records


escowles:
Acceptance Criteria

- [ ] Staff view responds with marc data for the SCSB record

# Implementation Notes
See https://github.com/pulibrary/orangelight/pull/1055#discussion_r157582513. We will need undo the workaround implemented here when the blocking bibdata ticket for this issue is completed.

christinach:
We can bring this up to the AUX team as future work.
Do users use these features? 
Is this a feature for library staff? If so we can give them access to scsb and view it.

PR #5537, opened 2026-01-16
[Record page] SCSB records with new marcxml field display the Staff view link

christinach:
[Record page] When a SCSB record:
* connect to solr and get the marcxml field which is stored and not indexed
* decompress and use the marcxml field in the librarian view
* for scsb records render the staff view link only if marcxml field is present?

    related to #1065


kevinreiss (approved):
Thanks @christinach

PR #5551, opened 2026-01-20
Update specs for theses and dissertations

christinach:
based on the new raw we get


christinach:
I should open a new pr . the raw has changed again.

PR #5556, opened 2026-01-22
Add Openssl Gem

kevinreiss:
Get around ssl verification erros when working …with omniauth and localhost applications.


christinach (approved):
Thanks @kevinreiss 🚴🏻‍♂️

PR #5631, opened 2026-02-16
Display the lux alert when submitting one of the 'Need Help?' forms

### LINKED ISSUE #5579: [Access] Flash Message for Modal Forms on Record Page Behaves Oddly
### Expected behavior
When I submit a modal form from the "Need Help?" section on the record page I expect to get feedback that my submission was successful after submitting form. 

### Actual behavior
When I fill out the form and submit it the modal closes with no acknowledgement of success. The flash message noting that the form was successfully submitted appears when I navigate away from the record page that was used to submit the form. 

### Steps to replicate
1. Go to a catalog record page.
2. Select any of the three options under "Need Help?"
3. Fill out the form and submit it. 
4. Navigate to another page in the catalog or refresh the page you are on (record page).
5. Observe the flash message. 

### Impact of this bug
Confusing user experience.

christinach:
closes #5579

The issue:
* Rails ujs remote: true expects a format.js template.
* The lux-alert components needed to be remounted to get the lux styles. Otherwise after the 2 fixes we were getting a plain text.

Solution:
* Use flash.now[:success] for AJAX requests
* Create three different js.erb templates to update the flash message
with the rendered partial
* In the js.erb template call window.luxImport() to mount and style the lux-alert component
* Refactored luxImport function to mount components on all .lux elements
every time it's called not only on page load
* Rescue if an exception is raised - in this case display the flash message
* Display the flash message if
@form.valid? returns false or
@from.submit returns false and no exception is raised

related to [#5579]

### DIFF
--- a/app/controllers/contact_controller.rb
+++ b/app/controllers/contact_controller.rb
@@ -1,35 +1,62 @@
 # frozen_string_literal: true
 class ContactController < ApplicationController
+  # :reek:DuplicateMethodCall
+  # :reek:TooManyStatements
+  # :reek:UncommunicativeVariableName
   def question
     @form = AskAQuestionForm.new(question_params)
-    if @form.valid? && @form.submit
-      flash[:success] = 'Your question has been submitted'
-
-      render "question_success"
-    else
-      render partial: "catalog/ask_a_question_form", locals: { form: @form }, status: :unprocessable_content
+    begin
+      if @form.valid? && @form.submit
+        flash_now_success('Your question has been submitted')
+      else
+        flash_now_error('There was a problem submitting your question')
+      end
+    rescue StandardError => e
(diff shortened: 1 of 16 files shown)


christinach:
@rladdusaw 🐈‍⬛  @sandbergja 🍄 this is a draft PR to fix the flash alerts. I include some console logs that I used for debugging in case you want to run it locally and see the console.log in the dev tools.
I had to add three js.erb templates to make the flash message work after submitting the ajax form (the forms use rails ujs - remote: true). I'm happy to move it now to vanilla JS before merging it. I wanted to know your thoughts before I do that. FYI the lux alert never worked properly on the record page. In the past it was triggered but with no lux styling and would not go away after 5 secs. I had to remount the component in order to see the lux-alert with the correct styling.

Issue #5753, opened 2026-04-13
Review list of unavailable statuses - in process - on order


christinach:
### What maintenance needs to be done?
After the changes in https://github.com/pulibrary/bibdata/pull/3178 we have simplified the status label we sent with the bibdata api call. We want to review the hardcoded labels in orangelight related to Unavailable, in progress and on order.

### Why is this maintenance needed?
- After the changes in https://github.com/pulibrary/bibdata/pull/3178 we have simplified the label we sent with the bibdata api call. We want to review the hardcoded labels in orangelight related to Unavailable, In Progress and On Order.

PR #5754, opened 2026-04-13
The specific unavailable statuses come from bibdata as Unavailable.

christinach:
No need to override them in Orangelight.

related to #5753 
[
bibdata calculate status](https://github.com/pulibrary/bibdata/blob/bf008ee7a4f7afa8ec3bebf835ff7849cf0bcc43/app/adapters/alma_adapter/alma_item.rb#L145-L154)


spec/cassettes/form_features.yml
```diff
@@ -476,8 +476,7 @@ http_interactions:
       - Phusion Passenger(R) 6.0.9
     body:
       encoding: UTF-8
-      string: '[{"barcode":"32101032361436","id":"23503918390006421","holding_id":"22503918400006421","copy_number":"1","status":"Not
-        Available","status_label":"Lost","status_source":"process_type","process_type":"LOST_LOAN","on_reserve":"N","item_type":"Gen","pickup_location_id":"firestone","pickup_location_code":"firestone","location":"firestone$stacks","label":"Firestone
+      string: '[{"barcode":"32101032361436","id":"23503918390006421","holding_id":"22503918400006421","copy_number":"1","status":"Unavailable","status_label":"Unavailable","status_source":"process_type","process_type":"LOST_LOAN","on_reserve":"N","item_type":"Gen","pickup_location_id":"firestone","pickup_location_code":"firestone","location":"firestone$stacks","label":"Firestone
```
christinach, line 479: for 3 specs in form_spec I had to edit in the cassette the status_label which we know it is Unavailable for the ones we send to Illiad. . In a follow up PR I will update these three specs with stubs.
These specs have more problems for example if you run them one at a time they fail with a login error. This also will be addressed in a follow up maintenance PR.

kevinreiss (approved):
Thanks @christinach

PR #5765, opened 2026-04-14
Manage bad data with empty enum and description. Remove unused Preservation conservation check

### LINKED ISSUE #5755: Items with bad data (enum or description empty strings) should not prevent from sending to illiad if the rest of the checks satisfy the criteria
The specific issue appears for items in East Asian with bad data in Alma that fail the eligibility check for the Illiad service. 
The problem was only revealed now after the changes in https://github.com/pulibrary/bibdata/pull/3178. 

For the East Asian items the enum and description values are empty strings, so the specific  method returns that are present and not nil . The next Illiad eligibility check is the Preservation Conservation which does not exist anymore as a status label. So the Illiad service eligibility check fails.

### Level of urgency
- [x] High
- [ ] Moderate
- [ ] Low

- Patrons cannot request through Illiad an item tht is in Preservation Conservation.

### 
Example:
https://catalog.princeton.edu/requests/9942430233506421?aeon=false&mfhd=22600149340006421&open_holdings=East+Asian+Library

This issue is related to bad data in Alma and exists in the catalog production for years.

christinach:
closes #5755

Make sure description and enum is not an empty string

The specific issue appears for items in East Asian with bad data in Alma that fail the eligibility check for the Illiad service. The problem was only revealed now after the changes in https://github.com/pulibrary/bibdata/pull/3178.

For the East Asian items the enum and description values are empty strings, so the specific  method returns that are present and not nil . The next Illiad eligibility check is the Preservation Conservation which does not exist anymore as a status label. So the Illiad service eligibility check fails.

related to [#5755]


app/models/requests/item.rb
```diff
@@ -54,10 +54,6 @@ def on_reserve?
       self[:on_reserve] == 'Y'
     end
 
-    def preservation_conservation?
-      status_label == "Preservation and Conservation"
```
christinach, line 58: In bibdata we're setting status label Unavailable when the work order type is Preservation and Conservation.

app/models/requests/item.rb
```diff
@@ -128,11 +124,13 @@ def location
     private
 
       def short_description_from_alma_availability_call
-        self[:enum_display]
```
christinach, line 131: East Asian items appear to have empty strings for enum and description.

kevinreiss: Interesting, I wonder why they have a different cataloging practice for this data.

kevinreiss (approved):
Thanks @christinach

PR #5770, opened 2026-04-15
Remove EZproxy service

### LINKED ISSUE #5744: Sunset the EZproxyService
The EzproxyService class and the associated hosts.dat file and config initializer can be removed from the project. 

### Level of urgency
- [ ] High
- [x] Moderate
- [ ] Low

- Subscription data no longer resides in Marc 856 fields coming from Alma. This feature's purpose was to add the Ezproxy prefix to subscription links stored in 856 data. 

### Acceptance criteria
 - [ ] EzproxyService class is removed along with method calls to it. 
 - [ ] Hosts.dat file and initializer are removed. 
 - [ ] Specs associated with these features are removed. 

[Initializer](https://github.com/pulibrary/orangelight/blob/698dd413dae11d302d706bbaefbc50e0fbd4d9ea/config/initializers/ez_proxy_hosts.rb). Data was stored in the EZ_PROXY_HOST_LIST environment variable.

rladdusaw:
Resolves #5744


spec/services/online_holdings_markup_builder_spec.rb
```diff
@@ -138,7 +138,7 @@
       end
 
       it 'flattens the set of electronic access titles' do
-        expect(urlified_markup).to include Requests.config['proxy_base']
+        expect(urlified_markup).not_to include Requests.config['proxy_base']
```
sandbergja: Could we just remove this test and the `proxy_base` config value?

spec/components/holdings/online_holdings_component_spec.rb
```diff
@@ -15,10 +15,10 @@
     expect(rendered.css('.online-holdings-list')).not_to be_empty
   end
 
-  it 'Converts links to EZProxy links where appropriate' do
+  it 'No longer converts links to EZProxy links' do
```
sandbergja: Could we remove this test?  I think it doesn't really make sense without knowing the history of the feature.

sandbergja (commented):
Thanks, @rladdusaw !  🐈‍⬛   Just some ideas on more things we could remove.

spec/services/online_holdings_markup_builder_spec.rb
rladdusaw: Done.

spec/components/holdings/online_holdings_component_spec.rb
rladdusaw: Done.

sandbergja (approved):
Thanks, @rladdusaw ! 🐈‍⬛

PR #5782, opened 2026-04-21
Remove yard

sandbergja:
We should still add good documentation to our code, but we currently don't generate the documentation in another format.


christinach (approved):
🍄

PR #5789, opened 2026-04-24
Remove unused field-hierarchy CSS rule

sandbergja:
It was originally added by some jquery, see https://github.com/pulibrary/orangelight/blob/41c587415bd497de5d69be46315cb1d1b514d331/app/assets/javascripts/orangelight.js#L7

We don't have this jquery or anything else adding this class anymore.


christinach (approved):
Thanks @sandbergja 🍄 !

PR #5791, opened 2026-04-24
Fix CSS variable name

sandbergja:
The variable is named `font-weight-regular`, not `font-weight-normal` -- oops!


kevinreiss (approved):
🍄

PR #5796, opened 2026-04-24
Remove unusued CSS related to Linked Versions panel

sandbergja:
`.location--panel` no longer has a descendant `.panel`.  We also no longer use `.location--online` anywhere.


christinach (approved):
🍄 🍄

PR #5799, opened 2026-04-24
Clean up front page CSS

sandbergja:
* Move CSS out of the shame file along with other front page styles
* Remove unused welcome-paragraph class
* Reduce duplication
* Prefer `@use` over `@import`
* Use lux variables where it is reasonable to do so


christinach (approved):
Thank you @sandbergja 🍄

PR #5801, opened 2026-04-27
Add marquand and stokes to active libmap libraries.

kevinreiss:


sandbergja (approved):
Thanks, @kevinreiss ! 🚴🏻

PR #5813, opened 2026-04-28
Use consistent blue colors in secondary header (aka search bar)

sandbergja:
Previously, the 'Advanced Search' and 'Bookmarks' links were slightly different shades.


regineheberlein (approved):
Thanks @sandbergja !

PR #5507, opened 2026-01-08
Removed biased results flipper and associated code

rladdusaw:
Resolves #5461

PR #5815, opened 2026-04-28
Remove CSS for search results feedback banner

sandbergja:
The feature was removed in #5507, but we never removed the relevant CSS.


regineheberlein (approved):
Thanks @sandbergja !

PR #5819, opened 2026-04-28
Migrate to SASS @use from @import for search-results.scss

sandbergja:
* Moves two colors to use Lux css variables as well
* Make some selectors more specific so they aren't overridden by styles from bootstrap


christinach (approved):
🍄 thanks!

PR #5820, opened 2026-04-28
Remove unneeded SASS @import rules

sandbergja:


christinach (changes requested):
@sandbergja I remember that 'colors' file overrides the bootstrap. I just deployed and confirmed that the colors are the ones from bootstrap  and not the ones from OL. can you take a look?

sandbergja:
Yikes!  Thanks for catching that @christinach !  I re-added those color overrides and re-deployed to staging, could you please check again when you have a chance?

christinach (approved):
Thanks @sandbergja 🍄

PR #5821, opened 2026-04-28
Use more consistent border radius across the application

sandbergja:


christinach (approved):
Thanks @sandbergja  🍄

PR #5823, opened 2026-04-28
Remove tiny left border on announcement

sandbergja:
I had never noticed this tiny border before!  I propose that we remove it, I don't think it adds much.


christinach (approved):
thanks @sandbergja

PR #5824, opened 2026-04-29
Migrate to blacklight CSS from blacklight SASS

sandbergja:
blacklight and blacklight-marc gems now provide their styles as CSS rather than SASS.  This uses their CSS verions.

A few nice benefits:
* blacklight-marc now uses flex rather than float to position the tags and indicators
* This reduces our uncompressed CSS bundle from 376K to 364K
* This gives us access to some Blacklight-related CSS variables in case we need them


christinach (approved):
Thanks @sandbergja 🍄

PR #5826, opened 2026-04-29
Remove @imports from availability.scss

sandbergja:
* Replace harcoded and SASS variables with lux variables where appropriate
* Remove unused CSS


christinach (approved):
Thanks @sandbergja 🍄

PR #5827, opened 2026-04-29
Remove some more CSS that is no longer needed

sandbergja:


christinach (approved):
Thanks @sandbergja 🍄 🐬

PR #5835, opened 2026-04-30
Remove small unused or unwanted CSS rules

sandbergja:
All of these are unused except for the .login_actions padding, which in my opinion does not look great
(I think it looks better when vertically aligned with the other form elements).

Before:
the log in button is several pixels right of the other form fields

After:

the log in button is vertically aligned with the other form fields


christinach (approved):
Thanks @sandbergja !

PR #5840, opened 2026-05-01
Remove unused CSS

sandbergja:
we don't have a `.close-button` class


christinach (approved):
Thanks @sandbergja 🍄 !

Issue #3058, opened 2022-07-06
Intermittent errors requesting from SCSB


maxkadel:
From Honeybadger https://app.honeybadger.io/projects/54399/faults/85848704/01G6VE35PB9JGCGZ23990RAVNX . 

Investigate whether this is an intermittent connection error, or if there is an issue with the way we're building the scsb request and create any needed follow-up tickets.

maxkadel:
Another example of possibly the same or similar issue - https://app.honeybadger.io/projects/54399/faults/86950285#notice-comments 

Could we re-try once or twice when we hit this error?

christinach:
This is still happening https://app.honeybadger.io/projects/54399/faults/85848704

christinach:
we [have changed the submit to be through JS ,](https://github.com/pulibrary/orangelight/pull/5281) so we will see if we're still seeing a similar error.

christinach:
@kevinreiss can we close this ticket? We've implemented the 'submit' to happen only through javascript and not through rails/js. Now we separate the javascript action from rails which is serving the messages and errors if any to the Js RequestHandler. We don't have anymore the submit.js.erb template file that was triggering this honeybadger error.

kevinreiss:
This error still seems to be coming up from the rails codebase, https://app.honeybadger.io/projects/54399/faults/117707295.

kevinreiss:
Closing ticket since we now have the re-try solution in place.

PR #5841, opened 2026-05-01
Rescue Farady Time out when trying to connect to SCSB

christinach:
helps with #3058


sandbergja (approved):
Thanks, @christinach ! 🐬

PR #5842, opened 2026-05-03
Migrate more SASS @import to @use

sandbergja:
Note that the styles in record--subject.scss were always being overriden by a more specific selector, so we remove them here.


christinach (approved):
thanks @sandbergja !

PR #5861, opened 2026-05-08
Compile Blacklight js using vite rather than sprockets

sandbergja:
This has two benefits:

1. The vite bundle does not block rendering, so the user has to wait a bit less time before the page renders.
1. Vite is better at minifying JS size.  Sprockets used 19 kB (6 kB compressed) to provide the Blacklight JS, while Vite uses 7 kB (2 kB compressed)

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -16,14 +16,4 @@
 //
 //= require bootstrap
 //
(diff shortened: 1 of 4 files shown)


christinach (approved):
Thanks @sandbergja !

PR #5862, opened 2026-05-08
Reduce bootstrap js package size

sandbergja:
* Use vite rather than sprockets
* Use only the js we need

This reduces the total bundled JS size by 104 kB (uncompressed) or 18 kB (compressed).

Specifically:

||Before|After|
|---|---|---|
|sprockets package|609 kB|448 kB|
|vite package|368 kB|425 kB|
|sprockets compressed|160 kB|125 kB|
|vite compressed|115 kB|132 kB|

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -12,8 +12,5 @@
 //
 //= require jquery3
 //= require rails-ujs
-//= require popper
-//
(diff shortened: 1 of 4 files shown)


kevinreiss (approved):

kevinreiss:
Thanks @sandbergja

PR #5865, opened 2026-05-11
Only load CSS needed for the requests form when on the requests form

sandbergja:
Reduces the CSS bundle needed for the home page, show page, search results, etc. by 16 kB.

### DIFF
--- a/app/assets/stylesheets/requests.scss
+++ b/app/assets/stylesheets/requests.scss
@@ -1,0 +1,2 @@
+@use "datatables";
+@use 'components/request';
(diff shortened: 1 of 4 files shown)


app/assets/stylesheets/requests.scss
```diff
@@ -0,0 +1,2 @@
+@use "datatables";
```
christinach, line 1: 👍🏻

christinach (approved):
Thanks @sandbergja 🍄

christinach:
tested the requests form on catalog-staging. it looks the same.

PR #5867, opened 2026-05-11
Remove overriden CSS rule

sandbergja:
We set it here, only to override it in overrides.scss


christinach (approved):
Thanks @sandbergja 🍄

PR #5871, opened 2026-05-12
Use vite, rather than sprockets, for the main scss bundle

sandbergja:
This really shrinks the size of the uncompressed CSS bundle:

- Before: 339 kB (47 kB compressed)
- After: 287 kB (44 kB compressed)

Running a single feature or system test locally is reliably faster as well, since SASS compilation was a significant percentage of the time spent running one of those tests:

```
bundle exec rspec spec/features/numismatics_spec.rb

before: 7.11 seconds
after: 5.24 seconds
```


kevinreiss (approved):
Thanks @sandbergja

PR #5883, opened 2026-05-12
Remove incorrect border-spacing rule

sandbergja:
border-spacing only has an effect on tables, but the facets are presented as a list, not a table.


christinach (approved):
Thanks @sandbergja 🍄

PR #5884, opened 2026-05-12
Another attempt at using vite for our main SASS bundle

sandbergja:
#5871 had an issue where our custom fonts were still being managed by sprockets, so vite was unable to determine the correct path for them.  This pr includes a commit that allows vite to manage these fonts.


christinach (approved):
@sandbergja I did a manual test. Everything looks good to me!
🍄

PR #5885, opened 2026-05-12
Skip links: fix a regression where they are not visible when you focus them

sandbergja:
I'm sure I introduced this at some point while mucking around with SCSS, but I'm not sure when...

Before:

A grey area shows at the top of the screen but there is no link in it

After:

A Skip to Search link appears in the grey area at the top of the screen

### DIFF
--- a/spec/system/skip_links_spec.rb
+++ b/spec/system/skip_links_spec.rb
@@ -2,6 +2,13 @@
 
 require 'rails_helper'
 
+RSpec::Matchers.define :be_off_screen do
+  match do |selector|
+    # Check to see if the element is clipped as in bootstrap visually hidden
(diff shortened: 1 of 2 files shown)


spec/system/skip_links_spec.rb
```diff
@@ -2,6 +2,13 @@
 
 require 'rails_helper'
 
+RSpec::Matchers.define :be_off_screen do
+  match do |selector|
+    # Check to see if the element is clipped as in bootstrap visually hidden
+    page.execute_script("return window.getComputedStyle(document.querySelector('#{selector}')).clip?.includes('rect(0')")
```
christinach, line 8: 👍🏻

christinach (approved):
Thanks @sandbergja !

Issue #2906, opened 2016-08-25
Non-title fields should also have language tags


escowles:

christinach:
See: [Pull request to add lan](https://github.com/pulibrary/orangelight/pull/2905/files)
See [key sites](https://docs.google.com/spreadsheets/d/14glZXKkH0hb3RBXrZhgXtrUI1DpkulH4eoorhzrUY6s/edit#gid=227276045)

PR #5886, opened 2026-05-12
Add language tags to more fields on the show page that are likely to be in the language of the resource

sandbergja:
Helps with #2906


christinach (approved):
Thanks @sandbergja !

PR #5888, opened 2026-05-12
Add language tags to Related works on the show page

sandbergja:
Prior to this commit, this field was rendered by a helper function named name_title_hierarchy.  The rendering pipeline was not consulted at all, so adding `language_tag: true` in the catalog controller had no effect.

This commit adds a new processor to the rendering pipeline that creates the same output as name_title_hierarchy, and migrates the field to use it (as well as adding language_tag: true)

Helps with #2906


christinach (approved):
Thanks @sandbergja !

PR #5893, opened 2026-05-13
Remove methods that look for location_code elf

christinach:
Bibdata does not have any elf in holding_locations.json elfs used to be in voyager


sandbergja (approved):
Thanks, @christinach ! 🐬

kevinreiss (approved):
Thanks @christinach

PR #5896, opened 2026-05-14
Add language tags to related names fields

sandbergja:
This requires us to refactor the _hash_show view, which previously did not use the Blacklight Rendering Pipeline.

This PR refactors fields that use `hash: true` to use a view component that uses the Blacklight Rendering Pipeline, so that it observes all the field configurations.

Thanks to @christinach for talking through the approach to this problem with me!

Helps with #2906


spec/system/catalog_show_spec.rb
```diff
@@ -134,29 +134,29 @@
 
   describe 'showing top fields' do
     let(:document_id) { '9946093213506421' }
-    let(:top_fields) { ['uniform_title_1display', 'format', 'pub_created_display', 'description_display'] }
+    let(:top_fields) { ['Uniform title', 'Format', 'Published', 'Description'] }
```
sandbergja, line 137: The text of these fields has not changed, but some of the DOM selectors have changed.  I switched this test to check for the field labels, rather than the DOM selectors, since that is what users will actually see.

app/components/hash_metadata_field_component.html.erb
```diff
@@ -0,0 +1,6 @@
+<% fields.each do |label, raw_values| %>
+  <%= render Blacklight::MetadataFieldLayoutComponent.new(field: field_config) do |component|
+    component.with_label { label }
+    render_values(raw_values).each_with_index {|value, index| component.with_value(value:, index:)}
```
christinach, line 4: 👍🏻

app/views/catalog/_show_default.html.erb
```diff
@@ -6,7 +6,7 @@
       <% next if field&.coin_description %>
       <%# marc relator code for authors -%>
       <% if json_field?(field) %>
-        <%= render 'hash_show', document: document, field: field_name %>
+        <%= render HashMetadataFieldComponent.new(document:, solr_field_name: field_name, field_config: field) %>
```
christinach, line 9: 👍🏻

spec/components/hash_metadata_field_component_spec.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+require 'rails_helper'
+
+RSpec.describe HashMetadataFieldComponent, type: :component do
+  it 'uses the rendering pipeline' do
+    field_config = Blacklight::Configuration::Field.new(key: 'field')
+    document = SolrDocument.new({ field: { 'Woof woof!': ['Dog'], 'Meow!': ['Cat'] }.to_json })
+    processor = Class.new(Blacklight::Rendering::AbstractStep) do
+      def render = values.map { it == 'Dog' ? '🐕' : it }
```
christinach, line 9: Aspen 👍🏻

spec/components/hash_metadata_field_component_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+    end
+    operations = [processor]
+
+    component = described_class.new(field_config:, solr_field_name: 'field', document:, operations:)
+    rendered = render_inline(component)
+
+    expect(rendered.css('dt').map(&:text)).to eq(['Woof woof!', 'Meow!'])
+    expect(rendered.css('dd').map(&:text)).to eq(['🐕', 'Cat'])
```
christinach, line 17: 🐈

christinach (approved):
Thanks @sandbergja 🍄 !

PR #5899, opened 2026-05-15
Add language tags to more fields on the show page

### LINKED ISSUE #2906: Non-title fields should also have language tags

sandbergja:
These are the remaining fields on the show page that we can assume are in the language of the resource, rather than English

Closes #2906


rladdusaw (approved):
Thanks @sandbergja!

PR #5933, opened 2026-06-04
Delete temporary browse list files after they have been loaded into Postgres

sandbergja:
This should help with a flapping disk usage alert in CheckMK


kevinreiss (approved):
Thanks @sandbergja

PR #5943, opened 2026-06-08
Don't require Francis' keys to deploy

sandbergja:


kayiwa (approved):
🤦🏿  thanks @sandbergja  🍄

PR #5947, opened 2026-06-10
bundle install on main

christinach:
adds gem and ruby platform


sandbergja (approved):
Thanks, @christinach ! 🐬

PR #5972, opened 2026-06-24
Add arm64-darwin platform to Gemfile.lock

sandbergja:
Without this, I get a nokogiri compilation error 😢


christinach (approved):
Thanks @sandbergja !

PR #5980, opened 2026-06-26
Replace several holding location requests from VCR cassette

sandbergja:
Some of these weren't actually used, and others could be easily replaced with our `stub_single_holding_location` helper


christinach (approved):
Thanks @sandbergja !

PR #5994, opened 2026-06-30
Address violations of new semgrep rules

sandbergja:
* Pin a github action dependency to a specific commit
* Increase dependabot cooldown period to 7
* Remove bearer (it is installed using a curl, which semgrep is concerned about -- but also we have not found it very useful).


kevinreiss (approved):

christinach (approved):
Thanks @sandbergja 🍄

PR #5996, opened 2026-06-30
Move a few more mocked requests from VCR to WebMock

sandbergja:


christinach (approved):
thanks @sandbergja !

christinach (approved):
thanks @sandbergja !

PR #5998, opened 2026-07-02
Remove checksum yarn.lock

christinach:
using package manager yarn from the node orb takes care of the lockfile and the yarn cache path


rladdusaw (approved):
Thanks @christinach!

PR #5999, opened 2026-07-02
Migrate requests form_spec from VCR to WebMock stubs

regineheberlein:
Remove the form_features cassette and stub every HTTP request explicitly with the existing helpers (stub_single_holding_location, stub_availability_by_holding_id, stub_catalog_raw), so unstubbed requests now fail loudly instead of being silently served by VCR.

- Extract 76 fixtures from the cassette's recorded response bodies
- Update 13 existing fixtures that had drifted from the recorded data (delivery locations and item barcodes no longer matched, which the cassette's playback precedence had been masking)
- Stub two bibdata availability endpoints the cassette served as errors (a 400 for 99125452799106421 and a 504 for the bad-mfhd case) inline, since the helpers only produce 200s
- Replace an Illiad transaction stub whose body matcher was copied from a different bib and never matched; the cassette's 401 response had been masking it, so stub the 401 the tests actually exercised
- Delete spec/cassettes/form_features.yml (only form_spec referenced it)


sandbergja (approved):
Thanks, @regineheberlein !  It was very interesting to see how the 🤖 did!

regineheberlein:
Thanks for experimenting together @sandbergja !

Final summary of steps from Claude:

Summary
Migrates spec/features/requests/form_spec.rb off VCR entirely. Every HTTP request is now explicitly stubbed with the existing WebMock helpers (stub_single_holding_location, stub_availability_by_holding_id, stub_catalog_raw), so unstubbed requests fail loudly instead of being silently served by the cassette. The 1.4 MB form_features.yml cassette is deleted (nothing else referenced it); VCR config is untouched for other specs.

How the stubs were derived
Rather than guessing from failure output, a per-example HTTP request inventory was logged while the suite was still green under VCR (via a temporary WebMock.after_request hook). Each request VCR was still serving was mapped to a helper call in the example that makes it, and missing fixtures were extracted verbatim from the cassette's recorded response bodies.

Changes
Remove vcr: { cassette_name: 'form_features', record: :none } from the describe block; add ~155 stub helper calls across 66 examples
Extract 76 new fixtures from the cassette (availability, holding-location, and catalog /raw bodies)
Update 13 existing fixtures that had drifted from the recorded data — delivery locations and item barcodes no longer matched, which VCR's playback precedence over local stubs had been masking
Stub two bibdata availability endpoints the cassette served as errors (a 400 and a 504) inline, since the helpers only produce 200s
Replace an Illiad transaction stub whose .with body matcher was copied from a different bib and never matched; the cassette's 401 had been masking it, so the stub now returns the 401 the tests actually exercised
Delete spec/cassettes/form_features.yml
Testing
Baseline before changes: 72 examples, 0 failures (with VCR)
Final state (no VCR metadata, cassette deleted): 72 examples, 0 failures
Sibling specs sharing the updated fixtures (marquand feature, scsb model, requests helper, request mailer, request options): all pass except marquand_spec.rb:69, which fails identically on clean main — it's a pre-existing isolation flake where the first CAS-authenticated example in a fresh rspec process loses its login_as session and renders the sign-in page. Unrelated to this change.
🤖 Generated with [Claude Code](https://claude.com/claude-code)

regineheberlein:
And for completion's sake, our starting prompt was:
"Hi Claude, we’re working in /Documents/orangelight. The file spec/features/requests/form_spec.rb currently uses VCR.  We would like to use webmock to mock HTTP requests instead.  Could you please migrate this file to use Webmock, using the existing helpers (e.g.  stub_single_holding_location  and  stub_availability_by_holding_id )?"

Issue #5981, opened 2026-06-26
[Record Page] All Records Linked to Figgy Objects should display Available Online Access Options


kevinreiss:
### User story
A record that is linked to a figgy item and is viewed on the individual record page, has a link "Digital Content" that takes the user to the object's viewer and should be available in the "Available Online" section of the page. 

### Acceptance criteria
 
- [ ] All Figgy Linked records display "Available Online" section with the a link to the access option on the record page. 
- [ ] I can successfully use this feature with Voice Over in Safari.
- [ ] I have checked this feature using one of the following browser extensions and confirmed that it does not add any level A or level AA issues:
    * axe dev tools
    * Lighthouse
    * Wave

### Concrete example
See [9962641133506421](https://catalog.princeton.edu/catalog/9962641133506421).

Current display:

<img width="1391" height="515" alt="Image" src="https://github.com/user-attachments/assets/3deebc59-af10-4a53-b742-559a2b6f45d6" />

Example of desired display for the above record - [example record](https://catalog.princeton.edu/catalog/9952961993506421) with a View Online section:

<img width="1379" height="510" alt="Image" src="https://github.com/user-attachments/assets/90369234-312a-41f4-aa0b-7aa12f3ec455" />

### Implementation notes, if any
Records attached to figgy items that are currently missing this content on the record page have only the "figgy_1display" key in their indexed data. See the [example record](https://catalog.princeton.edu/catalog/9962641133506421/raw) for details. Records that have the electronic_access_1display do display the "Available Online" section on the page. See this [example](https://catalog.princeton.edu/catalog/9952961993506421/raw) with both data points as an example.

christinach:
For now we decided to build the anchor for the Digital content in orangelight so that we dont do the change in bibdata.
Future work would be to refactor bibdata electronic_links method and do the appropriate princeton ark checks in the codebase.

christinach:
Resolved by https://github.com/pulibrary/orangelight/pull/6015/

PR #6015, opened 2026-07-13
Update doc_electronic_access to build the anchor for the digital content using the figgy_1display

christinach:
Update doc_electronic_access to build the anchor for the digital content if the figgy_1display field exists

related to #5981


rladdusaw (approved):
Thanks @christinach!

PR #6059, opened 2026-07-29
PF is not a valid delivery location for annex items

### LINKED ISSUE #6056: Review and update Annex valid pickup locations PA PF and PB
## Actual behavior
Annex locations display 'Firestone Library' twice in the request form. This is specific to the method [valid_annex_pick_up_locations](https://github.com/pulibrary/orangelight/blob/b6edcd9b66ec2b6b7ba963e27fe9b6e868c7de42/app/models/requests/pick_up_locations/annex_pick_up_locations.rb#L34) and to the delivery locations for annex holding locations that are configured in [bibdata holding_locations.json](https://github.com/pulibrary/bibdata/blob/68ca7c8ee092567b426a78d1d4fd88fb9d1b3ca8/config/locations/holding_locations.json#L283) 

## Expected behavior
Display one Firestone delivery location

## Notes
in orangelight using [annex_pick_up_locations](https://github.com/pulibrary/orangelight/blob/b6edcd9b66ec2b6b7ba963e27fe9b6e868c7de42/app/models/requests/pick_up_locations/annex_pick_up_locations.rb#L34) we're checking for valid locations and have both PA and PF for `annex$stacks`. Both have the label 'Firestone Library' . So they will both display in the request form drop down menu.

For delivery location PF there is logic to keep this delivery location if the holding location is `firestone$pf` (microforms).

Currently any `annex$...` holding location that has delivery locations PA and PF configured in  bibdata/holding_locations.json it will keep both delivery locations.

## Questions: 
1. Which firestone delivery location do we keep? 
2. Is there a specific annex location that has only microforms so that we add logic similar to the firestone$pf .

christinach:
closes #6056

Remove PF from the method that checks if a pickup location is valid for annex items

### DIFF
--- a/app/models/requests/pick_up_locations/annex_pick_up_locations.rb
+++ b/app/models/requests/pick_up_locations/annex_pick_up_locations.rb
@@ -32,9 +32,7 @@ module Requests
 
       # :reek:UtilityFunction
       def valid_annex_pickup?(location_hash)
-        # This excludes PQ (Plasma) and PH (Mudd), which are not valid but are currently listed as pickup locations
-        # in bibdata holding_locations.json
-        ['PA', 'PB', 'PF', 'PK', 'PL', 'PM', 'PT', 'PW', 'QA', 'QC', 'QL', 'QP', 'QT', 'QX'].include?(location_hash[:gfa_pickup])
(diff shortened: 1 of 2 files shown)


rladdusaw (approved):
Thanks @christinach!

PR #6061, opened 2026-07-29
Filter out staff only locations when the patron_group:'lib'

christinach:
In the request form, filter out staff only locations when the patron_group:'lib'

Library staff are the users in patron_group: 'lib'. Users in this group can request items to be delivered to  staff locations, these are the delivery - pickup locations that have the attribute `staff_only: true`.
In the request form list the staff locations for offsite items only if the user is in the 'lib' group.


mzelesky:
The user group is `LIB` [all uppercase], not `lib`.

christinach:
> The user group is `LIB` [all uppercase], not `lib`.
Thanks @mzelesky . There is no problem because we downcase when we do a check.

christinach:
Make sure to test annex, recap and eastasian$hygf (offsite) locations.
Test as alma and cas user.

spec/features/requests/form_spec.rb
```diff
@@ -86,7 +87,13 @@
...
       clear_enqueued_jobs
     end
-
+    # around do |example|
+    #   cached_admin_netids = ENV['ORANGELIGHT_ADMIN_NETIDS']
+    #   ENV['ORANGELIGHT_ADMIN_NETIDS'] = ''
+    #   example.run
+    # ensure
+    #   ENV['ORANGELIGHT_ADMIN_NETIDS'] = cached_admin_netids
+    # end
```
sandbergja: Can we remove this commented-out code?

spec/features/requests/form_spec.rb
```diff
@@ -76,6 +76,7 @@
     end
 
     before do
+      allow(Rack::MiniProfiler).to receive(:authorize_request)
```
sandbergja, line 79: Are you sure that this is needed?  It seems unrelated, but if the tests really need it, it's fine. :-D

spec/features/requests/form_spec.rb
```diff
@@ -187,38 +194,52 @@
...
+          stub_scsb_availability(bib_id: "99114026863506421", institution_id: "PUL", barcode: nil, item_availability_status: nil, error_message: "Bib Id doesn't exist in SCSB database.")
+          visit "/requests/#{recap_in_process_id}"
+          expect(page).to have_content 'In Process'
+          expect(page.find(:css, ".request--availability").text).to eq("Unavailable")
+          select('Firestone Library', from: 'requestable__pick_up_23753408600006421')
+          expect(page).not_to have_content 'Technical Services 693 (Staff Only)'
+          expect(page).not_to have_content 'Technical Services HMT (Staff Only)'
+
+          # select('Technical Services 693 (Staff Only)', from: 'requestable__pick_up_23753408600006421')
+          # select('Technical Services HMT (Staff Only)', from: 'requestable__pick_up_23753408600006421')
```
sandbergja: Do we need these commented lines?

sandbergja (commented):
THanks, @christinach ! 🐬   Some non-blocker cleanup ideas

spec/features/requests/form_spec.rb
christinach: I will run it again and see.

christinach: it passes without it, so I removed it. Thanks for reviewing  @sandbergja !

spec/features/requests/form_spec.rb
christinach: 🧹

spec/features/requests/form_spec.rb
christinach: 🧹

christinach:
Please make sure to review locations at annex, recap and eastasian$hygf
Also test as a cas user and as an alma user. For the alma user use the credentials from last pass.

kevinreiss (approved):
The functionality added here is working well. There are a few open questions for follow-up with the circulation group for behavior for the "LIB" group in a few edge case locations that do not need to block merging/deploying this. 

1. EAL Hygf
2. In library use locations at ReCAP. I.e. Lewis$pn and $ps.

PR #6078, opened 2026-08-05
Add dedupe cap deploy env

rladdusaw:


config/deploy/dedupe.rb
```diff
@@ -0,0 +1,49 @@
...
+set :rvm_ruby_string, :local # use the same ruby as used locally for deployment
+set :rails_env, 'dedupe'
+set :branch, ENV['BRANCH'] || 'main'
+
+server 'catalog-dedupe1.princeton.edu', user: 'deploy', roles: %i[web app db worker mailcatcher cron_db]
+server 'catalog-dedupe2.princeton.edu', user: 'deploy', roles: %i[web app worker mailcatcher]
+server 'catalog-indexer-dedupe1.princeton.edu', user: 'deploy', roles: %i[cron_db worker indexer]
+server 'catalog-indexer-dedupe2.princeton.edu', user: 'deploy', roles: %i[cron_db worker indexer]
+
+namespace :env do
```
sandbergja, line 12: Why can't we set these environment variables in ansible with the other envvars?

sandbergja (commented):
Thanks, @rladdusaw ! 🐈‍⬛   Looks good overall, I had a question about the deploy config

config/deploy/dedupe.rb
rladdusaw: Are you asking about the block starting with line 12?

I mostly just copied that from the staging file, but my understanding is that is there so we can update an environment variable with just the cap deploy instead of having to run the playbook.

sandbergja: Ah, I see.  The staging changes look super old (apart from lint fixes, those lines are from 7 years ago).  It would be my preference to only have the envvars in the playbook, rather than needing to try to troubleshoot how the two sets of envvars interact with each other (e.g. having to think through questions of whether `cap` or `ansible-playbook` was run most recently, looking at two repos for envvar values).  My preference would not be to continue this pattern in a new file, but I don't think it's a blocker.

sandbergja (approved):
Thanks, @rladdusaw ! 🐈‍⬛ It would be nice to not continue the envvar pattern in the new file, but I don't think it is a blocker (but of course I would not complain it if you are up for adjusting it!  Feel free to merge with or without that change.)

config/deploy/dedupe.rb
rladdusaw: Is your preference to just remove it? I don't think I even knew it was there until now.

sandbergja: Removing it would be great, as long as it doesn't mess anything up with the environment!

PR #6079, opened 2026-08-05
Migrate to use the title_sort_key

sandbergja:
See https://github.com/pulibrary/pul_solr/pull/564

Once this is merged, we can stop indexing into the `title_sort` field (which took up way too much memory) and index directly into `title_sort_key`.


kevinreiss (approved):
Thanks @sandbergja.

regineheberlein (approved):
Wow, thanks @sandbergja !

PR #6081, opened 2026-08-05
Include new changes in schema.xml from pul_solr

christinach:


sandbergja (approved):
Thanks, @christinach ! 🐬

Issue #6022, opened 2026-07-14
[Results and Bookmarks] Physical holdings card - align call number with library location and availability status


caroldh:
### Current display

<img width="740" height="285" alt="current display" src="https://github.com/user-attachments/assets/c4e31cb0-1797-49fe-a75c-8ec59c9dfd49" />

### Desired display

<img width="735" height="362" alt="desired display" src="https://github.com/user-attachments/assets/07e6f172-ef98-46d1-b581-1ae8ebd07732" />

### Acceptance criteria

- [x] Call Numbers align with the text for the Library and availability status on result page cards
- [ ] I can successfully use this feature with Voice Over in Safari.
- [x] I have checked this feature using one of the following browser extensions and confirmed that it does not add any level A or level AA issues:
    * axe dev tools
    * Lighthouse
    * Wave
- [ ] ...

### Notes
Font should remain black, but moved over to align with the other text in the card.

caroldh:
PO reviewed and approved

christinach:
resolved by https://github.com/pulibrary/orangelight/pull/6082

PR #6082, opened 2026-08-06
Align call_number with library location and availability status

christinach:
Use rem unit rather than px to scale dynamically
based on the root font size

related to #6022


sandbergja (approved):
Thanks, @christinach ! 🐬

Issue #6088, opened 2026-08-07
[Sandbox Data Set] Construct and Add Example "Clustered Records"


kevinreiss:
### Description

Our current thinking is that solr document will provide the [following information](https://docs.google.com/document/d/195RSXVzqWfTprBSEjEirQM7ilD2rZfXiYqk7E9PaFTI/edit?tab=t.0#heading=h.pka1s2qg6m8n) for merged records allowing for this content to be displayed. Construct some example documents that can be used to represent this data structure for the purposes of displaying content in the initial testing/review of the Catalog DeDupe environment. 

### Acceptance Criteria

- [x] Create a sample record represent a Print PUL Book and a Print SCSB Partner Book at three partner institutions
- [x] Create a sample record represent a Print PUL Book Series and Print SCSB Book Series that are a cluster and collaboratively collected - EAL examples https://catalog.princeton.edu/catalog/99110313313506421 and https://catalog.princeton.edu/catalog/SCSB-9261616. 
- [x] Create a sample record representing a Print PUL Book and a Princeton Ebook Record
- [x] Create a sample record representing a Print PUL Book, a Princeton Ebook, and a Princeton Audio resource
- [x] Create a sample record representing a Print PUL Book, a Princeton Ebook and a SCSB Print Book
- [x] Create a sample record representing a Print PUL Manuscript/Book and a Print SCSB Book
- [x] Create a sample "control" record representing an ebook held in a portfolio that is not part of a cluster."
- [x] Create a sample "control" record representing a print book that is not part of a cluster".
- [x] Create a sample "control" record representing a print book that is linked to a figgy object. 
- [x] ~~Create a sample "control" record representing an ebook held in a 856 link that is not part of a cluster."~~
- [x] Create a sample "control" record representing an ebook that has a lengthy description field.

kevinreiss:
the electronic_access_1display e-holding isn't conducive to adding additional attributes, we'll need to break that out into a re-factoring ticket before an example can be created for that class of holding.

kevinreiss:
Created ticket https://github.com/pulibrary/orangelight/issues/6111 to investigate issues with key electron_access_1display.

PR #6096, opened 2026-08-10
Add Fixtures to Index as Clustered records 

kevinreiss:
Advances issue https://github.com/pulibrary/orangelight/issues/6088.


christinach (approved):
Thanks @kevinreiss .

PR #6099, opened 2026-08-12
Refactor OnlineHoldingsMarkupBuilder to a View Component

sandbergja:


app/components/holdings/online_holdings_component.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 # This component is responsible for displaying a brief description of
-# a document's online holdings for consice contexts like the search
+# a document's online holdings for concise contexts like the search
```
christinach, line 4: In a future PR we should rename this file to indicate that is used in the index view (search results and bookmarks)

christinach (approved):
Thanks @sandbergja 🍄

PR #6101, opened 2026-08-12
Rename component to clarify that it is to be used in index views

sandbergja:
Per https://github.com/pulibrary/orangelight/pull/6099#discussion_r3769481084


christinach (approved):
Thanks @sandbergja 🍄 !

PR #6103, opened 2026-08-12
Revert changes that I accidentally introduced in an existing

christinach:
cluster fixture 12be71a4-8e76-47ea-98f0-8da9ae511406.json


sandbergja (approved):
Thanks, @christinach ! 🐬

kevinreiss (approved):

PR #6104, opened 2026-08-12
[Duplicate records with clusters] Display badge in physical holdings on show page

sandbergja:
Approved by POs


christinach (approved):
Thanks @sandbergja !

PR #6110, opened 2026-08-13
Lux 7.9.0

sandbergja:
This includes an improvement to the badge style, to make it more consistent with other colors in the catalog and also in Tigerdata


christinach (approved):
Thanks @sandbergja !

Issue #6033, opened 2026-07-16
[Sandbox data set] Include Microform and Manuscripts Records that also classified as Format Book


kevinreiss:
## User Story
I would like to preview how microform and manuscript records that are also classified as books appear in the sandbox. 

## Acceptance Criteria

- [x] A sample clustered record that has both the format type [manuscript and book](https://catalog.princeton.edu/catalog?f%5Bformat%5D%5B%5D=Book&f%5Bformat%5D%5B%5D=Manuscript) are added to the sandbox data set
- [x] A sample record that has both format type [microform and book](https://catalog.princeton.edu/catalog?f%5Bformat%5D%5B%5D=Book&f%5Bformat%5D%5B%5D=Microform) are added to the sandbox data set

kevinreiss:
Added the manuscript example in PR https://github.com/pulibrary/orangelight/pull/6106.

kevinreiss:
Finished by https://github.com/pulibrary/orangelight/pull/6112.

PR #6112, opened 2026-08-14
Adds example with microform and book formats for #6033.

kevinreiss:


christinach (approved):
Thanks @kevinreiss 🚴🏻‍♂️

PR #6122, opened 2026-08-17
Update cluster fixture to have id a uuid

christinach:


sandbergja (approved):
Thanks, @christinach ! 🐬

PR #6123, opened 2026-08-17
Refactor: memoize fetching host IDs from the DOM, so that it can be called at any time

sandbergja:
Previously, the host_ids is initially empty, then we set it later, so you can't make arbitrary changes to the order of operations.  This will allow us to change the order of operations more freely without worrying that we might get an empty array on a page where there are, in fact, host_ids.


christinach (approved):
Thanks @sandbergja 🍄

Issue #6107, opened 2026-08-13
Support cluster IDs in availability javascript


sandbergja:
### User story

As a user of the clustered catalog, I would like to see whether holdings are available

### Acceptance criteria

- [x] I can successfully use this feature with Voice Over in Safari.
- [x] I have checked this feature using one of the following browser extensions and confirmed that it does not add any level A or level AA issues:
    * axe dev tools
    * Lighthouse
    * Wave
- [x] The availability js works for clustered records on the Record Show  page when the Deduplication feature is on
- [x] The availability js works for clustered records on the Search Results  page when the Deduplication feature is on
- [x] The availability js works for clustered records on the Call Number Browse
- [x] There are no regressions to the availability js for non-clustered records in production-like environments

### Concrete example

https://catalog-dedupe.princeton.edu/catalog/12be71a4-8e76-47ea-98f0-8da9ae511406 -- currently it does not show availability, but it should.

### Implementation notes, if any

Currently, the HTTP requests done by this JS include the cluster ID, rather than the specific bibliographic ID.  We may wish to use the new `source_id` field in the holdings_1display instead

sandbergja:
This is done -- closing

PR #6124, opened 2026-08-17
Use holding source_id (if available) instead of the solr document id for Show page data-record-id attribute

sandbergja:
We use this attribute to connect bibdata availability response data with the appropriate holding on the show page.

Helps with #6107


kevinreiss (approved):

kevinreiss:
thanks @sandbergja  🍄

rladdusaw (commented):
Thanks @sandbergja!

Issue #6025, opened 2026-07-15
[Record Page] Staff View that shows metadata for all records attached to a cluster


kevinreiss:
### User story
As a user, I want to be able to see the MARC view for each individual record in a clustered record. 

### Acceptance criteria

- [ ] Is behind the sandbox feature flipper
- [ ] There is one 'Staff View' link on the record with the label "Other Views".
- [ ] Clicking the link renders to a new screen where the user can see all the staff views of the clustered or non clustered record. The screen has a tabbed interface. Each tab displays a different staff view. 
- [ ] I can successfully use this feature with Voice Over in Safari.
- [ ] I have checked this feature using one of the following browser extensions and confirmed that it does not add any level A or level AA issues:
    * axe dev tools
    * Lighthouse
    * Wave

### Notes
- We will use a new binary field to save the various SCSB records
- We might also do the same for alma records so that we dont need to connect to Alma to get the record. (cons: The AVA tag will not be up to date since we re-index changed records every hour.)
- Example text of Tab list in the new Staff view. 

Staff view - Princeton (book)
Staff view - Princeton (ebook)
Staff view - Harvard (book)

~~Staff view for Princeton records can be loaded on demand from bibdata, currently we will need to preserve the marcxml field for partner records in the index available by SCSB ID if we no longer index all individual records. ~~

~~Also noting that for all the links to SCSB items to work will need to implement a backend solution that provides the Marc data for SCSB items contained in a cluster of records.. This work is out of scope for this ticket. [link to ticket for that work when available].~~

christinach:
@sandbergja  @rladdusaw  @kevinreiss , regarding multiple compressed marcxml fields (partner record). I was thinking of having a new dynamic field marcxml_s as: 
```
"marcxml_s":[
{"SCSB-9756116": "H4sIAAAAAAAA/9VZbW/bNhD+K4QGDDVgW6TelRcBmZO0HfK2uAXWj7RE29wkUSNlO86v31F20iR2GQYNMMwIYjs63j333PHuyBxJlgtZoLuqrNWxM2/b5sB1V6vVsBT5cCaW7uXJ7cgjrip55WzEDu4Ufya68odCzlwPY+L+eX"},
[...]

PR #6126, opened 2026-08-17
Add cluster with marcxml_bi field from 4 records

christinach:
related to https://github.com/pulibrary/orangelight/issues/6025

Will update the holdings of this fixture in a separate PR


sandbergja (approved):
Thanks, @christinach ! 🐬

PR #6132, opened 2026-08-18
Fix rendering of serials Location Has on show page

### LINKED ISSUE #6130: "location_has" values no longer display on Record Page
The values stored in "location_has" are no longer displaying on the record page. [Reported here](https://pulibrary.slack.com/archives/CA0UY0Q07/p1787053735773319) 

### Notes
The lux see more component attached to holdings notes is defaulting to "display: none"

```
<div class="lux-show-more-content" tabindex="-1" style="display: none;">CURRENT ISSUES IN: Periodicals Collection (PR). Firestone</div><div class="lux-show-more-content"></div>
```

See https://github.com/pulibrary/orangelight/blob/main/app/components/holdings/holding_notes_component.html.erb#L13 for where this is rendered.

sandbergja:
PR #6104 inadvertently added a class class="lux", which causes an issue with Vue mounting/rendering.

In this commit, we remove the inner class="lux"

Closes #6130

### DIFF
--- a/app/components/holdings/holding_notes_component.html.erb
+++ b/app/components/holdings/holding_notes_component.html.erb
@@ -1,4 +1,4 @@
-<div class="holding-details lux">
+<div class="holding-details">
   <%= if render_shelving_titles?
     render_list label: 'Shelving title', list_class: 'shelving-title', notes: holding['shelving_title']
   end %>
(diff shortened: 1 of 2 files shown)


christinach (approved):
Thanks @sandbergja 🍄 !

PR #6133, opened 2026-08-18
Remove unnecessary test cleanup

sandbergja:
We have the following already, so these are redundant:

```
afterEach(() => vi.clearAllMocks();)
```


christinach (approved):
Thanks @sandbergja !

PR #6134, opened 2026-08-18
Update fixture so all holdings have display_format and source_id values.

kevinreiss:


sandbergja (approved):
Thanks, @kevinreiss ! 🚴🏻

PR #6135, opened 2026-08-18
Be more selective about which OpenTelemetry gems we install, and don't instrument them in test or local dev

sandbergja:
This makes starting your local dev environment almost a minute faster

Before:
```
$ time bundle exec rake browse:all browse:load_all
bundle exec rake browse:all browse:load_all  3.52s user 2.16s system 9% cpu 59.022 total
```

After:
```
$ time bundle exec rake browse:all browse:load_all
bundle exec rake browse:all browse:load_all  3.19s user 1.83s system 62% cpu 8.044 total`
```


christinach (approved):
Thanks @sandbergja 🍄 !

PR #6137, opened 2026-08-18
Remove a return value that is never used

sandbergja:
We create and populate an array named `result`, but we never use it for anything.


christinach (approved):
Thanks 🍄 !

PR #6138, opened 2026-08-18
Remove flipflop features we don't need any more

christinach:
1.hide_marquand_non_rare_request_button
2.hide_marquand_special_collections_request_button


kevinreiss (approved):
Thanks @christinach

PR #6148, opened 2026-08-20
Refactor some logic into its own inner class

sandbergja:


christinach (approved):
Thanks @sandbergja 🍄 !

PR #6150, opened 2026-08-20
Refactor and correct display of 856 links on catalog show page

sandbergja:
Move the actual tag generation to the erb template where it is more readable.

Also, this code was previously generating a <ul> with no <li> children (it was something like `<ul><ul><li></li></ul></ul>`), now it generates a standard `<ul><li></li><li></li></ul>` format


kevinreiss (approved):
Thanks @sandbergja

PR #6155, opened 2026-08-21
Add badges for portfolio-based online holdings

sandbergja:
Part of #6024

This PR includes the commits from PR #6153

PR #6157, opened 2026-08-21
Fix layout regression for links on search results

sandbergja:
Introduced by #6155

Before this fix, the "opens external resource" icon is on a new line:
catalog search result in Korean -- the link takes up a lot of vertical space because the icon is a line below the link

After this fix:
the same record, the link is much more compact since the icon is next to the link text


christinach (approved):
Thanks @sandbergja ! 🍄

Issue #6024, opened 2026-07-14
[Record page]Add format pill to online holdings on the record page 


caroldh:
### Current display
<img width="824" height="373" alt="current display" src="https://github.com/user-attachments/assets/f3de2353-f164-44f2-ba34-cf1cb719fd6d" />

### Desired display

<img width="879" height="251" alt="desired display" src="https://github.com/user-attachments/assets/e2611437-5550-4a3a-8925-26d0ee1a8189" />

Figma board - https://www.figma.com/design/jWuYZ9oOrTW39A7CHeHRLh/Catalog-and-All-Search?node-id=4020-787

### Desired Mobile Display

<img width="400" height="277" alt="Image" src="https://github.com/user-attachments/assets/9f46ff3b-2554-4c2a-95b1-59ed6b1ee5b1" />

### Acceptance criteria

- [x] Is behind the sandbox feature flipper 
- [x] Pills appear for records that have data in [electronc_portfolio_s](https://github.com/pulibrary/orangelight/blob/main/spec/fixtures/raw/cluster/99125137485706421.json) 
- [ ] Pills appear for records that have data in figgy_1display resources
- [ ] Pills appear for records that have data in electronic_access_1display
- [ ] I can successfully use this feature with Voice Over in Safari.
- [ ] I have checked this feature using one of the following browser extensions and confirmed that it does not add any level A or level AA issues:
    * axe dev tools
    * Lighthouse
    * Wave

### Notes
Ebook or Audio format pill will be positioned to the left of and outside a clickable box for online item.

sandbergja:
There are a few outstanding items:

- The POs noticed that when you hover/focus an electronic access link, the focus indicator collides with the badge -- we should give them a bit more space
- Get some consistency with the styling for badges on #6018 -- POs would prefer text not to be bold, the figma has it as Semi-bold (600 font weight).  Make sure that the style is consistent for all badges in Lux
- This is only working for portfolios

PR #6159, opened 2026-08-21
Add more spacing between format badges and electronic access link

sandbergja:
This way, when the link has focus, the focus indicator does not collide with the format badge.

Helps with #6024

Before:
Overdrive link is focused with a large orange focus indicator -- it collides with the audio format badge

After:
same situation, but there are a few pixels between the badge and the link


christinach (approved):
Thanks @sandbergja 🍄 !

PR #6174, opened 2026-08-25
Add missing source_id to cluster fixture holdings_1display

sandbergja:
Without it, the availability.js in catalog-dedupe tried to use the solr document id to fetch the holdings, which was a UUID and caused bibdata to return a 400.

Thanks to @christinach for identifying the UUID issue!


christinach (approved):
Thanks @sandbergja ! 🍄

Issue #6170, opened 2026-08-24
[Cluster record] The request form is populated with the metadata of the cluster record.


christinach:
## Desired state
When I click the request button of a cluster member holding, the Request form should be populated with the metadata (title, author) of the cluster record and use the holding id and the items of the cluster member.

## Current state
When the deduplication feature is on I can pass the correct cluster member id but the request form does not have the metadata of the cluster record and it misses the items of the cluster member.

## Acceptance criteria

- [ ] The request form should use the title and author of the main cluster record
- [ ] The request form should use the items of the cluster member record
- [ ] The request form should display the location of the cluster member record. 
- [ ] The request form should display the availability of the cluster member record.
- [ ] The requests.yml should be updated so that the [pulsearch in dedupe environment- pulsearch_base: `https://catalog-staging.princeton.edu`](https://github.com/pulibrary/orangelight/blob/8bd79a3f34b711a05e253b15876f4cb5965d9d90/config/requests.yml#L38) is pointing to `https://catalog-dedupe.princeton.edu`

PR #6175, opened 2026-08-25
Update dedupe env in requests.yml to point to catalog-staging

christinach:


sandbergja (approved):
Thanks, @christinach ! 🐬

christinach:
we will merge this until https://github.com/pulibrary/orangelight/issues/6170 is resolved so that the request form does not look broken.

PR #5506, opened 2026-01-08
Remove unneeded flippers

rladdusaw:
Resolves #5462
Resolves #5463

PR #6178, opened 2026-08-26
Switch to the newer lc_pipe_facet field

sandbergja:
PR #5506 inadvertently kept the legacy field and stopped using the new field.  There is no change in behavior, but this allows us to remove the legacy field from our indexing and solr.


christinach (approved):
Thanks @sandbergja !

PR #6180, opened 2026-08-26
Remove old lc_1letter_facet and lc_rest_facet from the config

sandbergja:
We have not used them since #4505


christinach (approved):
Thanks @sandbergja 🍄 !

PR #6182, opened 2026-08-26
Availability.js: do not request Alma status for IDs that are not Alma Ids

sandbergja:
In the dedupe environment, we saw an issue in which a source_id was missing, so we fell back to trying to request availability for the document ID (in that case, a cluster UUID).  This caused bibdata to return an error, so no availability appeared on the search results page for any of the records.

We fixed the data in #6174, but this should protect us from this issue better in the future, by confirming that Alma IDs we send to bibdata are in the correct Alma MMS ID format.

This is a follow-up to the work in #6107

### DIFF
--- a/spec/javascript/orangelight/availability_search_results.spec.js
+++ b/spec/javascript/orangelight/availability_search_results.spec.js
@@ -7,6 +7,16 @@ global.console = {
 
 global.fetch = vi.fn();
 
+function mockSuccessfulAvailabilityFetch(jsonResponse) {
+  const mockResponse = {
+    ok: true,
+    status: 200,
+    json: vi.fn().mockResolvedValue(jsonResponse),
+  };
+
+  return vi.spyOn(global, 'fetch').mockResolvedValue(mockResponse);
+}
+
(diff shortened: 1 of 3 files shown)


spec/javascript/orangelight/availability_search_results.spec.js
```diff
@@ -219,9 +229,58 @@ describe('AvailabilitySearchResults', function () {
...
+    );
+    expect(fetch).not.toHaveBeenCalledWith(
+      expect.stringContaining('f7a2277a-9fe6-4528-b12e-2e60b1a8bb80')
+    );
+  });
+
   test('request_search_results_availability handles fetch errors', async () => {
     document.body.innerHTML = `
-      <div data-availability-record="true" data-record-id="123"></div>
+      <div data-availability-record="true" data-record-id="9932127373506421"></div>
```
sandbergja, line 283: Needed to update this test because `123` is not a valid Alma id, so we never did the fetch and hit the error under test.

christinach (approved):
thanks @sandbergja !🍄🍄

Issue #5691, opened 2026-03-17
Requests to SOLR with more than 1024 clauses return the Something went wrong screen


christinach:
### Current state
SOLR by default has <maxBooleanClauses>${solr.max.booleanClauses:1024}</maxBooleanClauses>
If we exceed this limit then we get the 'Look something went wrong screen'
[Example with many clauses](https://catalog.princeton.edu/catalog?search_field=all_fields&q=The+fyrst+boke+of+the+introduction+of+knowledge.+The+which+doth+teache+a+man+to+speake+part+of+all+maner+of+languages.+and+to+knowe+the+vsage+and+fashion+of+al+maner+of+co%C5%A9treys.+And+for+to+knowe+the+moste+parte+of+all+maner+of+coynes+of+money%2C+ye+which+is+curraunt+in+euery+region.+Made+by+Andrew+Borde%2C+of+phisicke+doctor) [that Angie reported in the catalog-circulation-holdings-request channel](https://pulibrary.slack.com/archives/C034AS0QPC7/p1773781286716539)

### Desired state
When the user searches for more than a specific number of characters truncate them and return results.

### Notes
Limit the number of words we send to solr.  

[honeybadger error](https://app.honeybadger.io/projects/54399/faults/127173812)
```
Blacklight::Exceptions::InvalidRequest: RSolr::Error::Http - 500 Internal Server Error
Error: Query contains too many nested clauses; maxClauseCount is set to 1024

URI: http://lib-solr9-prod.princeton.edu:8983/solr/catalog-production/advanced?wt=json
Request Headers: {"Content-Type" => "application/json"}
[...]

PR #6191, opened 2026-09-03
Do not send extremely long queries to solr

sandbergja:
Before this commit, we regularly got the error 'Query contains too many nested clauses' from Solr.

This commit truncates queries with more than 20 words, with the assumption that any query that long will be specific enough that the query will still likely provide the desired result.

It would be nice to also reduce the numbers of `qf`s and `pf`s in our default search, since they also contribute to this nested clause explosion.  Once we do that, we could probably increase this limit.

Helps with #5691


rladdusaw (approved):
Thanks @sandbergja!

Issue #6147, opened 2026-08-19
[Requests] Commons Library Items Can be Requested


kevinreiss:
### What maintenance needs to be done?
Currently we suppress the request button for the Commons library location. Items should be allowed to be requested. 

### Level of urgency
- [ ] High
- [x] Moderate
- [ ] Low

### Why is this maintenance needed?
- Commons Library items need to be requested for pick-up at other libraries on campus. 

### Acceptance criteria
 - [x] Holdings in the Commons Library display the request button.

christinach:
done

26 'Closed' items will move out of 'Closed'. we don't need any more changes.

christinach:
reopen to update the code so that the pickup locations are the ones [configured in bibdata](https://github.com/pulibrary/bibdata/blob/bd9960c7f9fd8b54fc1b7c1b76635b3066c88faa/config/locations/holding_locations.json#L590)

christinach:
Update commons to have the following pickup locations

<img width="1580" height="906" alt="screenshot of the request form - commons library item" src="https://github.com/user-attachments/assets/eb02fa29-409c-4ea6-a057-bbba76bbab2b" />

christinach:
resolve by https://github.com/pulibrary/orangelight/pull/6202

PR #6202, opened 2026-09-09
Include commons in the standard_circ_location? check

christinach:
related to #6147

see https://github.com/pulibrary/orangelight/issues/6147#issuecomment-5606233702


sandbergja (approved):
Nice, thanks @christinach 🐬

PR #6204, opened 2026-09-10
bundle update rubocop

christinach:


Gemfile.lock
```diff
@@ -349,7 +349,7 @@ GEM
       rails-dom-testing (>= 1, < 3)
       railties (>= 4.2.0)
       thor (>= 0.14, < 2.0)
-    json (2.21.2)
+    json (3.0.2)
```
sandbergja, line 352: Are we sure that our version of Rails is compatible with JSON 3?  I had [heard about this problem](https://calvin.my/posts/the-json-gem-v3-0-upgrade-broke-rails-app)

sandbergja (commented):
Thanks, @christinach !  Just one question 🐬

Gemfile.lock
christinach: Thanks for sharing @sandbergja . I'm running it locally to see if it has any errors

christinach: I 'm not seeing any issues locally

sandbergja: Deployed to staging and it looks good, plus all the tests pass.  So I think we are good!

sandbergja (approved):
Thanks, @christinach ! 🐬

PR #6205, opened 2026-09-10
bundle update vite-ruby

christinach:


sandbergja (approved):
Thanks, @christinach ! 🐬

sandbergja (approved):
Thanks, @christinach ! 🐬

PR #6210, opened 2026-09-11
Pin json to 2.21

christinach:
the upgrade to json v3.0 causes issues
we pin to v2.21 .
The json issues was fixed and backported to rails 8-1-stable branch.

Instead of using this branch we pin to 2.21
and will update when a new release is published with the fixed json version

see: [post on son-gem-v3-0-upgrade-broke-rails-app](https://calvin.my/posts/the-json-gem-v3-0-upgrade-broke-rails-app)
https://github.com/rails/rails/pull/58601


rladdusaw (approved):
Thanks @christinach!

Issue #6223, opened 2026-09-14
[Bookmarks] CSV Export is Failing


kevinreiss:
### Expected behavior
When using the bookmarks and I select the export to CSV option, I expect to get a CSV file of the bookmark records displayed on the current. 

### Actual behavior
Choosing the export to CSV option with any number of records produces the something went wrong screen. 

### Steps to replicate
1. Login into your account
2. Go to Bookmarks page
3. Select export to CSV
4. Observe error

### Impact of this bug
Bookmark export fails. No fallback option. 

### Honeybadger link and code snippet, if applicable

See https://app.honeybadger.io/projects/54399/faults/134236680

kevinreiss:
Replicated in both prod and local environments, the error is raised with 12 or more records in the export. The failure appears random, it's not a data issue with specific records.

christinach:
resolved by https://github.com/pulibrary/orangelight/pull/6228

PR #6228, opened 2026-09-15
Do not truncate long query that has only {!lucene}id:()

christinach:
We post this query when we export bookmarks.
The user should be able to export mode than 20 bookmarks

related to #6223


spec/models/search_builder_spec.rb
```diff
@@ -79,6 +79,26 @@
...
+      ids = (1..12).map { |id| "991223049235064#{id}" }
+      solr_parameters = { q: "{!lucene}id:(#{ids.join(' OR ')})" }
+
+      subject.truncate_long_queries(solr_parameters)
+
+      expect(solr_parameters[:q]).to eq("{!lucene}id:(#{ids.join(' OR ')})")
+    end
+
+    it 'still truncates long queries with multiple words' do
+      query = (1..25).map { |word| "Bonjour! #{word}" }.join(' ')
```
sandbergja, line 93: 😄

sandbergja (approved):
Good catch!  Thanks, @christinach ! 🐬

PR #6232, opened 2026-09-16
Add a Semantic Logger filter to remove complex Objects from the payload

sandbergja:
Before this commit, semantic logger was hitting this error before it could log most events:

```
Unable to create the configured appenders. The log level has been raised
to WARN and the output directed to STDERR until the problem is fixed. --
Exception: ArgumentError: Could not convert symbol: :json_filter_formatter
to a class in: SemanticLogger::Formatters. Looking for:
SemanticLogger::Formatters::JsonFilterFormatter
```

This commit takes an alternative approach to addressing payloads that contain infinitely recursive self-referential JSON -- we recursively remove anything from the payload that isn't a Hash, Array, String, or Number.

To exercise this patch, I deployed to staging and confirmed that:
* before deploying, a search in signoz-staging logs for `service.name in ['orangelight'] AND log_type in ['app-log'] host.name in ['catalog-staging1.princeton.edu'] controller` returned no results
* after deploying, the same query returned some results

Closes #6231

PR #6233, opened 2026-09-16
Format logs as json

sandbergja:
Use the json formatter from Semantic Logger.

This fixes a regression introduced in #6232, which inadvertently switched to the default semantic logger formatter.  We want the JSON one instead!


christinach (approved):
Thanks @sandbergja 🍄 🐬 !

